PR #2, opened 2025-03-19
Add basic readme

pepelsbey:
So that anyone can run it locally 🙂


README.md
```diff
@@ -0,0 +1,19 @@
...
+MDN's next fr(ont)e(n)d.
+
+## Getting started
+
+> [!NOTE]
+> You only need Node.js installed. If you don't have globally installed yarn, you can run commands as `npx yarn` instead of `yarn`.
+
+1. In the [yari](https://github.com/mdn/yari/) repository:
+  - Install dependencies `yarn`
+  - Start backend `yarn rari serve`
```
LeoMcA: We have rari available in fred now with `npm run rari`, so this bit can be rewritten

pepelsbey: I’m documenting what works here :) I couldn’t make `npm run rari serve` work.

```
npm run rari serve

> fer@0.0.1 rari
> rari serve

Updating @mdn/browser-compat-data to 5.7.4

thread 'main' panicked at crates/rari-types/src/settings.rs:83:54:
CONTENT_ROOT is not a valid path: Os { code: 2, kind: NotFound, message: "No such file or directory" }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

LeoMcA: Set `CONTENT_ROOT` to the path of your content, e.g:

`CONTENT_ROOT=../content/files npm run rari serve`

LeoMcA (approved):

PR #8, opened 2025-03-25
Refactor menu, fix issues

pepelsbey:
### Description

- Splits one long CSS file into multiple elements
- Fixes issue with button icons
- Simplifies theming for Learn, Tools, About sections
- Updates icons and sets consistent sizes


components/menu/elements/content.css
```diff
@@ -0,0 +1,107 @@
+/* Content */
+
+.menu__content {
+  display: grid;
+  grid-template-columns: repeat(3, 1fr);
+}
+
+/* Full */
+
+.menu__content-full {
```
caugner, line 10: Wouldn't this be more more BEMmy?

```suggestion
.menu__content--full {
```

pepelsbey: In this case, it’s not a modifier, it’s a multi-part name of an element:

```html
<div class="menu__content">
  <div class="menu__content-full">
```

I use modifiers in the panel to change colors:

```html
<div class="menu__panel menu__panel--learn">
```

LeoMcA (approved):
Had a brief scan through: it's made me realise I don't really like BEM, but let's table that discussion for our CSS style discussion we're having next week.

Haven't run locally, but nothing major stands out in the code: ship it!

PR #13, opened 2025-03-26
style(stylelint): add stylelint

caugner:
### Description

Adds `stylelint`, disable failing rules for now, and ignore one empty block.

### DIFF
--- a/components/colors/index.css
+++ b/components/colors/index.css
@@ -1,8 +1,8 @@
 /* Basic */
 
 :root {
-  --white: #fff;
-  --black: #000;
+  --white: #ffffff;
+  --black: #000000;
 }
 
 /* Gray */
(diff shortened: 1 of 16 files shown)


LeoMcA (approved):
Looks fine, though I'll reserve the right to complain about any of these in the future and recommend we switch them :P

components/colors/index.css
```diff
@@ -1,8 +1,8 @@
 /* Basic */
 
 :root {
-  --white: #ffffff;
-  --black: #000000;
+  --white: #fff;
+  --black: #000;
```
pepelsbey: I find it weird when `#369` and `#336698` colors look different in my code. I like it when all HEX colors are six-character. You can always make them shorter when optimizing the output.

Happy to complain about it later to let the PR be quickly merged.

caugner: Good point! Fixed via 68b1f7d, ensuring we consistently use the long notation.

PR #16, opened 2025-03-26
style: observatory styling

argl:
### Description

Observatory styling

### Motivation

frontend


.gitignore
```diff
@@ -1,4 +1,5 @@
 node_modules
 dist
 out
-.eslintcache
\ No newline at end of file
+.eslintcache
+.DS_Store
```
caugner:
```suggestion
.DS_Store

```

caugner:
@argl I took a quick look at the deployment, and noticed that the font sizes in the top menu changed as side-effect of these Observatory styling changes. Are you intentional changing the global font sizes?

argl:
@caugner I did [bring over some font/text related css](https://github.com/mdn/fred/blob/7e60f7ed1a1c1d29881f5c1e393aea1a3df5c97d/components/fonts/index.css#L37C2-L46C1) from the old frontend, like h1-h6 typography. I think that is what changed? Otherwise any changes down in observatory should never have any effect on the rest of the site.

I needed some baseline typography to build on, that is the reason why I brought it over.

components/icon/feedback.svg
```diff
@@ -1,3 +1,40 @@
...
+     inkscape:deskcolor="#d1d1d1"
+     inkscape:zoom="51.590909"
+     inkscape:cx="10.001762"
+     inkscape:cy="11"
+     inkscape:window-width="2048"
+     inkscape:window-height="1306"
+     inkscape:window-x="129"
+     inkscape:window-y="51"
+     inkscape:window-maximized="0"
+     inkscape:current-layer="svg1" />
```
caugner, line 34: Some inkscape attributes got inside here.

PR #45, opened 2025-04-07
feat(components): add Article Footer

caugner:
### Description

Adds the Article Footer as a server component.

### Additional details

See [file structure here](https://github.com/mdn/fred/tree/src-folder).


components/article-footer/feedback.js
```diff
@@ -0,0 +1,156 @@
...
+        >
+          ${this.l10n`Yes`}
+        </button>
+        <button
+          type="button"
+          class="button no"
+          @click=${{ handleEvent: () => this._handleVote(false) }}
+        >
+          ${this.l10n`No`}
+        </button>
```
caugner: @LeoMcA Do you have a tip why these click handlers don't attach?

See: https://fred-pr45.review.mdn.allizom.net/en-US/docs/Web#article_footer

caugner: Thanks, fixed via 44f9d67.

PR #51, opened 2025-04-08
add server and client button components

LeoMcA:
Turns the button component into one (or, really, two) usable both server side and client side, though I think we should start to use new terminology:
- server component: a functional component, taking a `Rari.Context` as its first argument, only used server side
- element (or custom/lit element if greater clarity is required): has a shadow dom, usually gets rendered server side and hydrated client side where it gains its interactivity
- pure component: a functional component, doesn't take a `Rari.Context`, used within both server components and custom elements

Using that new terminology, this adds a pure `Button` component, and an `<mdn-button>` element: the latter of which uses the former in rendering, and which share css.

The css piece I'm not entirely happy with: I need to think about it more, in the context of how we eventually want to optimise the css we're shipping on each page (only what's necessary, not one massive blob), but it works for now!

I also need to think about how we use the symmetric context (i.e. fluent) in a pure component.


components/button/static.js
```diff
@@ -0,0 +1,20 @@
...
+ * @param {string} [options.href]
+ */
+export function Button({ label, icon, disabled = false, href }) {
+  return href
+    ? html`<a class="button" href=${href}>
+        ${icon ? html`<span class="icon">${icon}</span>` : nothing} ${label}
+      </a>`
+    : html`<button ?disabled=${disabled} class="button">
+        ${icon ? html`<span class="icon">${icon}</span>` : nothing} ${label}
+      </button>`;
```
caugner: Deduplicate:

```suggestion
  const inner = html`${icon ? html`<span class="icon">${icon}</span>` : nothing} ${label}`
  return href
    ? html`<a class="button" href=${href}>${inner}</a>`
    : html`<button ?disabled=${disabled} class="button">${inner}</button>`;
```

components/content-feedback/index.js
```diff
@@ -72,18 +72,20 @@ export class ContentFeedback extends L10nMixin(LitElement) {
...
-        })}
-        ${Button({
-          extraClasses: "no",
-          handleClick: () => this._handleVote(false),
-          icon: thumbsDown,
-          label: this.l10n`No`,
-        })}
+        <mdn-button
+          data-vote="yes"
+          @click=${this._handleVote}
```
caugner, line 84:
```suggestion
          @click=${{ handleClick: () => this._handleVote(true) }}
```

PR #63, opened 2025-04-10
prototype how hooks might work using code-examples

LeoMcA:
two approaches here: a "custom elements light" approach, and a full embrace, in the first and second commits respectively


components/copy/index.js
caugner, line 1: Given that copy can mean something else, how about `copy-button`?

PR #66, opened 2025-04-10
fix: pass section titles through unescaped

argl:
### Description

fix: pass section titles through unescaped


caugner:
@argl Does this mean `title` is already escaped HTML?

argl:
`title` can contain verbatim html, see `/en-US/blog/color-palettes-css-color-mix/#getting_familiar_with_the_color-mix_function`

PR #77, opened 2025-04-14
Breadcrumb styling & greys

Rumyra:
### Description

First pass at styling the breadcrumb bar - notable things
- Reviewed the site for greys and suggested 6 greys which should be global are in the component
- Styled a button for search use - agreed to keep the input but move the button code to the search component commented out incase we change our mind - `mdn-search`
- Search should stay in top part of header and removed from breadcrumb bar - theme, save & language in breadcrumb bar
- We will show breadcrumbs on every page

### DIFF
--- a/components/breadcrumbs/index.css
+++ b/components/breadcrumbs/index.css
@@ -3,18 +3,30 @@
 .breadcrumbs {
   --breadcrumbs-side-padding: 1rem;
 
+  /* these colors should go globally - these are the greys */
+  --color-grey-light-super: hsl(212deg 7% 98%);
+  --color-grey-light: hsl(212deg 7% 93%);
+  --color-grey-mid-light: hsl(212deg 7% 80%);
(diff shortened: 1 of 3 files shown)


pepelsbey:
I find it a bit less intuitive that `mid-light` is lighter than just `light`.

What do you think about this naming scheme? Also, look how nicely it curves :D

```css
--color-grey-lightest: hsl(212deg 7% 98%);
--color-grey-lighter: hsl(212deg 7% 93%);
--color-grey-light: hsl(212deg 7% 80%);
--color-grey-dark: hsl(212deg 7% 26%);
--color-grey-darker: hsl(212deg 7% 16%);
--color-grey-darkest: hsl(212deg 7% 12%);
```

pepelsbey (approved):
Looks good! Thank you :) Any further naming improvements will come later.

PR #83, opened 2025-04-15
dynamically load custom elements

LeoMcA:


build/plugins/generate-element-map.js
```diff
@@ -0,0 +1,65 @@
...
+
+export class GenerateElementMapPlugin {
+  /**
+   * @param {Compiler} compiler
+   */
+  apply(compiler) {
+    compiler.hooks.beforeCompile.tapPromise(
+      "GenerateElementMapPlugin",
+      async () => {
+        const files = await fg("components/*/element.js", {
```
LeoMcA: Move to `fdir`

PR #93, opened 2025-04-16
fix(color-theme): load saved theme, simplify code

LeoMcA:
add support for inline js chunk to load theme as early as possible


components/color-theme/element.js
```diff
@@ -58,63 +54,67 @@ export class MDNColorTheme extends L10nMixin(LitElement) {
...
         data-side="right"
-        hidden
+        ?hidden=${!this._dropdown}
       >
         <ul class="color-theme__list">
           <li>
-            <button class="color-theme__option" @click=${setDefault}>
+            <button
+              class="color-theme__option"
+              data-mode="osDefault"
```
pepelsbey:
```suggestion
              data-mode="light dark"
```

Since we already use `light` and `dark` values when forcing the theme, we could also skip the variable and use it here. It would simplify the logic in the `utils.js` a bit.

PR #97, opened 2025-04-17
fix(nav): make hamburger interactive

LeoMcA:
put hook within component, as they're tightly coupled: I think this should be an approved paradigm


pepelsbey (approved):
Looks good! Thank you :)

PR #117, opened 2025-04-25
add ServerComponent class to support split css, migrate most components

LeoMcA:
- adds a ServerComponent class, and migrates most server components to it, to keep track of server component invocation on each page
- removes `import "foobar.css";` imports in js, and instead bundles css declaratively based on its name
- loads those css bundles on-demand based on the server components used in the page
- adds some eslint rules around naming of server components and custom elements
- moves almost everything under `./components/`: pages, blog, observatory, outer layout

still TODO before merging:
- [ ] a number of TODOs in code
- [ ] change/settle on naming of files: should probably be something like `server.js/css` for server components
- [ ] split commit into one commit renaming files, the other making changes, to make the changes here easier to see


LeoMcA:
After discussion this morning, will merge as is to avoid ongoing merge conflicts, then make fixes/enhancements in later commits/PRs

PR #145, opened 2025-05-06
feat(placement): initial support

fiji-flo:
### Description

Only sidebar and top for now.


types/placements.d.ts
```diff
@@ -0,0 +1,40 @@
+namespace Placements {
```
LeoMcA: These should be under `Fred.Placements` rather than just `Placements`, but if it's feasible it'd be best to have them traditionally imported/exported from a `components/placement/types.d.ts` file - namespaces are a bit flaky

types/user.d.ts
```diff
@@ -0,0 +1,55 @@
+namespace User {
```
LeoMcA: Same as above, and similarly I'd move to `components/user/types.d.ts` with traditional import/export if feasible

components/global/placements.js
LeoMcA, line 1: I'd move to something like `components/placements/context.js`

components/global/user.js
LeoMcA, line 1: Would also move to something like `components/user/context.js`

components/placement-utils/mixin.js
LeoMcA, line 1: I'd move everything under `components/placement-utils` to `components/placement` -> it'll be very neat if/when we're able to nest components like `components/placement/placement-top/element.js`

components/placement-utils/viewed.js
```diff
@@ -0,0 +1,15 @@
+/**
+ *
+ * @param {Placements.PlacementData} placementData
+ */
+export function viewed(placementData) {
```
LeoMcA: This function doesn't seem to be used anywhere, is there a missing file?

components/placement-utils/links.js
LeoMcA, line 1: I'd move these into the mixin as methods, then there's no need to import them in each placement element

components/placement-utils/mixin.js
```diff
@@ -0,0 +1,88 @@
...
+    _placementRef = createRef();
+
+    /**
+     * @type {string | undefined}
+     */
+    viewedUrl;
+    /**
+     * @type {number | undefined}
+     */
+    version;
```
LeoMcA: I'd suggest moving into the constructor, and prefixing with `_`

components/placement-utils/mixin.js
```diff
@@ -0,0 +1,88 @@
...
+    connectedCallback() {
+      super.connectedCallback();
+      this._dataTask.run();
+    }
+
+    /**
+     *
+     * @param {Placements.PlacementContextData} _placementContext
+     */
+    renderComplete(_placementContext) {}
```
LeoMcA: I'd suggest something like:
```suggestion
    /**
     * @abstract
     * @param {Placements.PlacementContextData} _placementContext
     */
    _renderComplete(_placementContext) {
      throw new Error("Must be implemented by subclass");
    }
```

components/placement-utils/mixin.js
```diff
@@ -0,0 +1,88 @@
...
+    renderPending() {
+      return nothing;
+    }
+    /**
+     *
+     * @returns {TemplateResult | symbol}
+     */
+    renderInitial() {
+      return nothing;
+    }
```
LeoMcA: I'd suggest merging these into one method, as there's a bug in lit SSR atm which requires them to be the same to avoid hydration errors: I also don't think it makes sense to have a different state for the two in this context. Same suggestions as above with `_renderComplete`.

components/placement-bottom/element.js
```diff
@@ -0,0 +1,70 @@
...
+    return html`<div
+      ${ref(this._placementRef)}
+      class="bottom-placement"
+      style=${styleMap(styles)}
+    >
+      <section class="placement-container">
+        <a
+          class="placement-link"
+          data-glean="pong: pong-&gt;click top-banner"
+          data-href=${clickLink(click, version)}
```
LeoMcA: Oops, is this left over from our testing of the `<p>` tag issue? There's a couple more in the other placements, though

PR #152, opened 2025-05-09
feat(a11y-menu): initial impl

caugner:
### Description

Adds a skip link menu for accessibility with two items:

- Skip to main content
- Skip to search

### Motivation

Parity with yari.


caugner:
@LeoMcA How would you recommend to implement these:

1. Add "Skip to search", and show the search modal.
2. If possible: Hide the "Skip to content" if there is no `#content` element.

PR #165, opened 2025-05-15
Update external link implementation

pepelsbey:
### Description

- Uses material icon
- Sticks to the current text color


components/external-link/global.css
```diff
@@ -1,19 +1,17 @@
...
+.external {
   display: inline-flex;
+  column-gap: 0.25em;
+  align-items: center;
 
-  width: 0.75em;
-  height: 0.75em;
+  &::after {
+    width: 1em;
+    height: 1em;
```
caugner, line 8: Why did we increase the icon size?

components/external-link/global.css
```diff
@@ -1,19 +1,17 @@
...
 
-  margin-left: 0.25em;
+    content: "";
 
-  content: "";
+    background-color: currentcolor;
```
caugner, line 12: As already mentioned, using the same color for the external indicator makes it less obvious to identify external links visually.

components/external-link/global.css
```diff
@@ -1,19 +1,17 @@
-.external::after {
+.external {
   display: inline-flex;
+  column-gap: 0.25em;
```
caugner, line 3: Making `.external` an `inline-flex` element affects all elements inside the link, not only the pseudo element, and this has changed the Specifications list to look like this:

image

Example:

```html
<a class="external" href="https://html.spec.whatwg.org/multipage/semantics.html#attr-link-integrity" rel="noopener" target="_blank">HTML<br><small># attr-link-integrity</small></a>
```

Can we revert to `margin-left` on the pseudo element?

caugner (commented):
FYI @pepelsbey

PR #200, opened 2025-05-28
Don’t render disabled links in pagination

pepelsbey:
### Description

Disabling interactive elements is not a good idea: they are still focusable and very confusing for screen readers. It’s better not to render them at all or render something inert instead. I’d suggest the simplest solution: render nothing.

### Note

Also, linter gives me the following error on line 148:

> This attribute binds the type 'DirectiveResult<{ prototype: any; directiveName: string; resultType: number; new(partInfo: ChildPartInfo | AttributePartInfo | ElementPartInfo) => UnsafeHTMLDirective; }> | undefined' which can end up binding the string 'undefined'. Use the 'ifDefined' directive?

### DIFF
--- a/components/pagination/server.css
+++ b/components/pagination/server.css
@@ -28,11 +28,6 @@
       text-decoration: none;
     }
 
-    .pagination__link--disabled {
-      cursor: not-allowed;
-      opacity: 0.5;
-    }
-
     .pagination__item--prev > .pagination__link,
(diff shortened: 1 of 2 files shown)


caugner:
This was meanwhile fixed by e2fc465181ed1c0616250aa7bc38cfcae8e55ae8.

PR #225, opened 2025-06-12
L10n Review PR

caugner:
### Description

(Ignore this PR.)


l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
+article-footer_last-modified = This page was last modified on <time data-l10n-name="date">{ $date }</timestamp> by <a data-l10n-name="contributors">MDN contributors</a>.
+article-footer_source_title = Folder: { $folder } (Opens in a new tab)
+
+baseline_asterisk = { $asterisk } Some parts of this feature may have varying levels of support.
```
peiying2: What is this variable for? Or is it for formatting?

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+baseline_high_extra = This feature is well established and works across many devices and browser versions. It’s been available across browsers since { $date }.
+baseline_low_extra = Since { $date }, this feature works across the latest devices and browser versions. This feature might not work in older devices or browsers.
+baseline_not_extra = This feature is not Baseline because it does not work in some of the most widely-used browsers.
+baseline_supported_in = Supported in { $browsers }
+baseline_unsupported_in = Not widely supported in { $browsers }
+baseline_supported_and_unsupported_in = Supported in { $supported }, but not widely supported in { $unsupported }
```
peiying2: What are shown in `{ $supported }` and `{ $unsupported }`? Add a comment will provide context will help.

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+
+homepage-hero_title = Resources for <u data-l10n-name="developers">Developers</u>,<br> by Developers
+homepage-hero_description = Documenting <a data-l10n-name="css">CSS</a>, <a data-l10n-name="html">HTML</a>, and <a data-l10n-name="js">JavaScript</a>, since 2005.
+
+reference_toc_header = In this article
+
+footer_mofo = Visit <a data-l10n-name="moco">Mozilla Corporation’s</a> not-for-profit parent, the <a data-l10n-name="mofo">Mozilla Foundation</a>.
+footer_copyright = Portions of this content are ©1998–2024 by individual mozilla.org contributors. Content available under <a data-l10n-name="cc">a Creative Commons license</a>.
+
+search_title = Search results for: <em>{ $query }</em>
```
peiying2: Need a comment on what the variable is for, like the kind of queries.

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+search_stats = Found { $results } matches in { $time } milliseconds.
```
peiying2: So this string is about how fast results can be found, not how many results found in X number of articles.

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+}
+blog_post_not_found = Blog post not found.
+
+blog_previous = Previous Post
+blog_next = Next Post
+
+Report = Report
+
+obs_title = HTTP Observatory
+obs_landing_intro = Launched in 2016, the HTTP Observatory enhances web security by analyzing compliance with best security practices. It has provided insights to over 6.9 million websites through 47 million scans.
```
peiying2: Is it better to change the websites and scan numbers to variables to minimize the needs for updates in the future?

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+obs_assessment = Developed by Mozilla, the HTTP Observatory performs an in-depth assessment of a site’s HTTP headers and other key security configurations.
+obs_scanning = Its automated scanning process provides developers and website administrators with detailed, actionable feedback, focusing on identifying and addressing potential security vulnerabilities.
+obs_security = The tool is instrumental in helping developers and website administrators strengthen their sites against common security threats in a constantly advancing digital environment.
+obs_mdn = The HTTP Observatory provides effective security insights, guided by Mozilla's expertise and commitment to a safer and more secure internet and based on well-established trends and guidelines.
+
```
peiying2: remove extra line break.

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+blog_time_to_read = { $minutes ->
+    [one]   { $minutes } minute read
+   *[other] { $minutes } minutes read
+}
+blog_post_not_found = Blog post not found.
+
+blog_previous = Previous Post
+blog_next = Next Post
+
+Report = Report
```
peiying2: Is this a verb or a noun?

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+compat_browser_version_date = { $browser } { $version } – Released { $date }
+
+compat_link_report_issue = Report problems with this compatibility data
+compat_link_report_issue_title = Report an issue with this compatibility data
+compat_link_report_missing_title = Report missing compatibility data
+compat_link_report_missing = Report this issue
+compat_link_source = View data on GitHub
+compat_link_source_title = File: { $filename }
+
+compat_deprecated = Deprecated
```
peiying2: Add a comment for this section, suggesting that these set of strings are related to Compatibility and where the strings appear. Is this from a report generated as the result of a search?

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+compat_experimental = Experimental
+compat_nonstandard = Non-standard
+compat_no = No
+
+compat_support_full = Full support
```
peiying2: Please add a comment where these strings show up? The context will help understand what these strings refer to, in the search result or the report?

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+compat_support_partial = Partial support
+compat_support_no = No support
+compat_support_unknown = Support unknown
+compat_support_preview = Preview browser support
+compat_support_prefix = Implemented with the vendor prefix: { $prefix }
```
peiying2: What does `vendor` mean here? Is this related to the name of vendor? Example what a "prefix" looks like would also help with context.

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+compat_support_altname = Alternate name: { $altname }
+compat_support_removed = Removed in { $version } and later
+compat_support_see_impl_url = See <a data-l10n-name="impl_url">{ $label }</a>
+
+compat_legend = Legend
```
peiying2: Are these strings shown in a graph , table generated from a search result? Please add a comment for the section to give some context.

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+compat_legend_preview = In development. Supported in a pre-release version.
+compat_legend_no = { compat_support_no }
+compat_legend_unknown = Compatibility unknown
+compat_legend_experimental = { compat_experimental }. Expect behavior to change in the future.
+compat_legend_nonstandard = { compat_nonstandard }. Check cross-browser support before using.
+compat_legend_deprecated = { compat_deprecated }. Not for use in new websites.
+compat_legend_footnote = See implementation notes.
+compat_legend_disabled = User must explicitly enable this feature.
+compat_legend_altname = Uses a non-standard name.
+compat_legend_prefix = Requires a vendor prefix or different name for use.
```
peiying2: Need examples for vendor and prefix look like in this context.

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
+article-footer_last-modified = This page was last modified on <time data-l10n-name="date">{ $date }</timestamp> by <a data-l10n-name="contributors">MDN contributors</a>.
+article-footer_source_title = Folder: { $folder } (Opens in a new tab)
+
+baseline_asterisk = { $asterisk } Some parts of this feature may have varying levels of support.
+baseline_high_extra = This feature is well established and works across many devices and browser versions. It’s been available across browsers since { $date }.
+baseline_low_extra = Since { $date }, this feature works across the latest devices and browser versions. This feature might not work in older devices or browsers.
```
peiying2: Nitpick, maybe changing the second `This feature` to `It`.

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+baseline_not_extra = This feature is not Baseline because it does not work in some of the most widely-used browsers.
+baseline_supported_in = Supported in { $browsers }
+baseline_unsupported_in = Not widely supported in { $browsers }
+baseline_supported_and_unsupported_in = Supported in { $supported }, but not widely supported in { $unsupported }
+
+homepage-hero_title = Resources for <u data-l10n-name="developers">Developers</u>,<br> by Developers
+homepage-hero_description = Documenting <a data-l10n-name="css">CSS</a>, <a data-l10n-name="html">HTML</a>, and <a data-l10n-name="js">JavaScript</a>, since 2005.
```
peiying2: Remove the comma before `since 2005`.

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+
+reference_toc_header = In this article
+
+footer_mofo = Visit <a data-l10n-name="moco">Mozilla Corporation’s</a> not-for-profit parent, the <a data-l10n-name="mofo">Mozilla Foundation</a>.
+footer_copyright = Portions of this content are ©1998–2024 by individual mozilla.org contributors. Content available under <a data-l10n-name="cc">a Creative Commons license</a>.
```
peiying2: Can we make `1998–2024` dynamically updated?

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+baseline_supported_in = Supported in { $browsers }
+baseline_unsupported_in = Not widely supported in { $browsers }
+baseline_supported_and_unsupported_in = Supported in { $supported }, but not widely supported in { $unsupported }
+
+homepage-hero_title = Resources for <u data-l10n-name="developers">Developers</u>,<br> by Developers
+homepage-hero_description = Documenting <a data-l10n-name="css">CSS</a>, <a data-l10n-name="html">HTML</a>, and <a data-l10n-name="js">JavaScript</a>, since 2005.
+
+reference_toc_header = In this article
+
+footer_mofo = Visit <a data-l10n-name="moco">Mozilla Corporation’s</a> not-for-profit parent, the <a data-l10n-name="mofo">Mozilla Foundation</a>.
```
peiying2: For links, we typically use this format `<a href="{ $moco }">Mozilla Corporation</a>` for example.

peiying2 (changes requested):
In addition to the comments to the strings, here are a few more general comments:
- Is there a staging URL to verify the translation? [An example](https://github.com/mozilla-l10n/www-l10n/blob/master/en/firefox/new/desktop.ftl) from mozilla.org project
- `en` vs `en-US` locale - In case en-US has US specific content that's not applicable to other languages/countries, it's a better practice to use `en` as the source for localizable content
- Speaking of locales, the Country code should be in capital letters, so en-US, pt-BR, zh-CN, zh-TW
- Mozilla brand and product names: when referencing names, [this approach](https://projectfluent.org/fluent/guide/references.html) ensures consistency. If there are more pages to add to the project, you can create a separate files for names only, and other pages can reference to it. 
- Links: I didn't call them all out but only commented in one string. They are typically coded like this `<a href="{ $moco }">{ -brand-name-mozilla-corporation}</a>.
- The page lacks comments for context. Please add them even if it seems obvious in some cases. Please note the [differences in comments](https://firefox-source-docs.mozilla.org/l10n/fluent/review.html#comments) for a section and comment for one string.

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
+article-footer_last-modified = This page was last modified on <time data-l10n-name="date">{ $date }</timestamp> by <a data-l10n-name="contributors">MDN contributors</a>.
```
bcolsson:
```suggestion
article-footer_last-modified = This page was last modified on <timestamp data-l10n-name="date">{ $date }</timestamp> by <a data-l10n-name="contributors">MDN contributors</a>.
```

Tags don't match

l10n/en-us.ftl
bcolsson: Looking into this it seems its just the character `*`. It's best to either use the character itself here, or if there area reasons for not wanting localizers to be able to adjust it, then perhaps its best to remove it from the string and style on the backend.

l10n/en-us.ftl
bcolsson: This is more stylistic - but can we consistently use dashes `-` instead of underscores for string IDs?
E.g. `article-footer-last-modified`

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+article-footer_source_title = Folder: { $folder } (Opens in a new tab)
```
bcolsson: Could you include a comment to describe variables? Something like the below:

```suggestion
# Variables:
#   $folder (String) - Link to github folder
article-footer_source_title = Folder: { $folder } (Opens in a new tab)
```

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+
+baseline_asterisk = { $asterisk } Some parts of this feature may have varying levels of support.
+baseline_high_extra = This feature is well established and works across many devices and browser versions. It’s been available across browsers since { $date }.
```
bcolsson: Could you include a comment for the variable here and similar date variables below?
```suggestion
# Variables:
#   $date (String) - a localized date string in format Month YYYY (e.g. December 2024)
baseline_high_extra = This feature is well established and works across many devices and browser versions. It’s been available across browsers since { $date }.
```

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+baseline_low_extra = Since { $date }, this feature works across the latest devices and browser versions. This feature might not work in older devices or browsers.
+baseline_not_extra = This feature is not Baseline because it does not work in some of the most widely-used browsers.
+baseline_supported_in = Supported in { $browsers }
```
bcolsson: Please also include a comment to provide context for these browser variables? (E.g. are these formatted lists? examples of what they would look like)

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+baseline_unsupported_in = Not widely supported in { $browsers }
+baseline_supported_and_unsupported_in = Supported in { $supported }, but not widely supported in { $unsupported }
+
+homepage-hero_title = Resources for <u data-l10n-name="developers">Developers</u>,<br> by Developers
+homepage-hero_description = Documenting <a data-l10n-name="css">CSS</a>, <a data-l10n-name="html">HTML</a>, and <a data-l10n-name="js">JavaScript</a>, since 2005.
+
+reference_toc_header = In this article
```
bcolsson:
```suggestion
# Header text followed by a table of contents
reference_toc_header = In this article
```

The `toc` in the string ID might not be easily recognized by localizers so we should include some sort of context.

l10n/en-us.ftl
bcolsson: I think this depends on the backend, it appears that's how [bedrock](https://github.com/mozilla/bedrock/blob/e55b1bc9b71833a5652578ec922e98e043758521/bedrock/base/templates/includes/protocol/footer/footer.html#L123) is doing things for their footer links, but might not be applicable in this case.

l10n/en-us.ftl
bcolsson: This string won't work as it is, assuming $results also returns a number of matches, you'll encounter grammatical errors even in English - `Found 1 results in 1 milliseconds.`
The likelihood is higher in a language like Russian which has 3 different plurals depending on the number returned.

Fluent plurals would fix this, but unfortunately it gets too complicated when you have two plural selectors in the same message. To fix this I'd split this into two independent sentences.
```
search-stats-results = { $results ->
    [one] Found { $results } match.
   *[other] Found { $results } matches.
}
search-stats-time = { $time ->
    [one] Search took { $time } millisecond.
    *[other] Search took { $time } milliseconds.
}
```

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+footer_copyright = Portions of this content are ©1998–2024 by individual mozilla.org contributors. Content available under <a data-l10n-name="cc">a Creative Commons license</a>.
+
+search_title = Search results for: <em>{ $query }</em>
+search_stats = Found { $results } matches in { $time } milliseconds.
+
+search-modal_site-search = Site search for <em>{ $query }</em>
+
+blog_time_to_read = { $minutes ->
+    [one]   { $minutes } minute read
+   *[other] { $minutes } minutes read
```
bcolsson: I'm guessing this is an existing string taken from somewhere? But reviewing this I think the natural English would be
`1-minute read` / `10-minute read`
(Please feel free to confirm with content team)

In that case this string might be better as:
```
blog-time-to-read = { $minutes ->
   *[other] { $minutes }-minute read
}
```
We'd still want to keep the plural selector in there to trigger plurals for other languages.

l10n/en-us.ftl
bcolsson: Another style request for string IDs, can we keep everything lowercase?

Also we should namespace with a more descriptive name, e.g. `report-page-title`

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+}
+blog_post_not_found = Blog post not found.
+
+blog_previous = Previous Post
+blog_next = Next Post
+
+Report = Report
+
+obs_title = HTTP Observatory
```
bcolsson: Should `HTTP Observatory` be localized or treated as a brand term / left unlocalized?

l10n/de.ftl
```diff
@@ -0,0 +1,33 @@
+content-feedback_question = War diese Übersetzung hilfreich?
```
bcolsson: Noting here that there seem to be string IDs here that don't match anything in en-US source.

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+compat_link_report_issue_title = Report an issue with this compatibility data
+compat_link_report_missing_title = Report missing compatibility data
+compat_link_report_missing = Report this issue
+compat_link_source = View data on GitHub
+compat_link_source_title = File: { $filename }
+
+compat_deprecated = Deprecated
+compat_experimental = Experimental
+compat_nonstandard = Non-standard
+compat_no = No
```
bcolsson: This likely needs a comment to say what "No" is refuting here (e.g. is this refusing some request, or is it short for `No compatibility` or something else?)

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+compat_legend_disabled = User must explicitly enable this feature.
+compat_legend_altname = Uses a non-standard name.
+compat_legend_prefix = Requires a vendor prefix or different name for use.
+compat_legend_more = Has more compatibility info.
+
+sidebar-filter-matches = { $matches ->
+  [0] No matches
+  [1] { $matches } match
+   *[other] { $matches } matches
+}
```
bcolsson:
```suggestion
sidebar-filter-matches = { $matches ->
  [0] No matches
  [one] { $matches } match
   *[other] { $matches } matches
}
```

l10n/en-us.ftl
```diff
@@ -0,0 +1,102 @@
...
+blog_next = Next Post
+
+Report = Report
+
+obs_title = HTTP Observatory
+obs_landing_intro = Launched in 2016, the HTTP Observatory enhances web security by analyzing compliance with best security practices. It has provided insights to over 6.9 million websites through 47 million scans.
+obs_assessment = Developed by Mozilla, the HTTP Observatory performs an in-depth assessment of a site’s HTTP headers and other key security configurations.
+obs_scanning = Its automated scanning process provides developers and website administrators with detailed, actionable feedback, focusing on identifying and addressing potential security vulnerabilities.
+obs_security = The tool is instrumental in helping developers and website administrators strengthen their sites against common security threats in a constantly advancing digital environment.
+obs_mdn = The HTTP Observatory provides effective security insights, guided by Mozilla's expertise and commitment to a safer and more secure internet and based on well-established trends and guidelines.
```
bcolsson:
```suggestion
obs_mdn = The HTTP Observatory provides effective security insights, guided by Mozilla’s expertise and commitment to a safer and more secure internet and based on well-established trends and guidelines.
```

bcolsson (commented):
In addition to the individual comments - we should consider if this needs separate brand terms for:
Mozilla
Mozilla Corporation
Mozilla Foundation
HTTP Observatory

This depends on how frequently new strings will be introduced once localization begings and how likely they'd include these terms.

caugner:
> * Is there a staging URL to verify the translation? [An example](https://github.com/mozilla-l10n/www-l10n/blob/master/en/firefox/new/desktop.ftl?rgh-link-date=2025-06-12T22%3A02%3A52Z) from mozilla.org project

Currently, all PRs by privileged users lead to a review deployment, and we'll extend this to non-privileged users eventually.

> * `en` vs `en-US` locale - In case en-US has US specific content that's not applicable to other languages/countries, it's a better practice to use `en` as the source for localizable content

MDN is only served in `en-US`, not `en`, i.e. `en` redirects to `en-US`.
 
> * Speaking of locales, the Country code should be in capital letters, so en-US, pt-BR, zh-CN, zh-TW

Fixed in ea86ac4e87a688d02a7c953415c5e511a3da829d (not on this branch).

> * Mozilla brand and product names: when referencing names, [this approach](https://projectfluent.org/fluent/guide/references.html) ensures consistency. If there are more pages to add to the project, you can create a separate files for names only, and other pages can reference to it.
> 
> * Links: I didn't call them all out but only commented in one string. They are typically coded like this `[{ -brand-name-mozilla-corporation}]({ $moco }).
> * The page lacks comments for context. Please add them even if it seems obvious in some cases. Please note the [differences in comments](https://firefox-source-docs.mozilla.org/l10n/fluent/review.html#comments) for a section and comment for one string.

Thanks, I added action items for us, to migrate to kebab-case, use comments, and consider using terms.

peiying2:
@caugner Any update on this?

LeoMcA:
Hey @peiying2 - sorry about the wait, the l10n work so far was just to ensure we had a functional API for development, rather than anything remotely usable by localisers. I've turned my attention to this now, but I need to explain a bit of our philosophy here:

I'm not a huge fan of writing strings out of context: I think it makes making mistakes easier, and is much harder for code contributors. Because of that, I've designed an API around inlining strings, like this (see the `this.l10n` template tags):

```js
html`<div class="language_switcher__remember">
  <mdn-switch
    @toggle=${this._togglePreferredLocale}
    ?checked=${this._isLocalePreferred}
    >${this.l10n`Remember language`}</mdn-switch
  >
  <mdn-button
    variant="plain"
    .icon=${infoIcon}
    icon-only
    href="https://github.com/orgs/mdn/discussions/739"
    target="_blank"
    title=${this
      .l10n`Enable this setting to always switch to the current language when available. (Click to learn more.)`}
    >${this.l10n`Learn more`}</mdn-button
  >
</div>`
```

I've got a proof of concept string extractor which produces a Fluent file like the following:

```ftl
remember-language-8ut5 = Remember language
enable-this-setting-to-always-sw-1ah2 = Enable this setting to always switch to the current language when available. (Click to learn more.)
learn-more-1o3l = Learn more
```

We do also support manually specifying the id, using syntax like `` this.l10n("default")`OS Default` ``. Otherwise for more complex strings - with arguments, etc. - we don't inline them (for now, at least).

I've got a bunch of questions about where to go from here:

1. How much do IDs matter: you'll see I've made a slug out of the string, truncated it, and appended a hash (so e.g. "hello" and "Hello!" don't clash). Could I do away with the slug entirely and just use hashes? Or would it be best to not use hashes at all, and error in the case of clashes, telling the developer to manually specify the ID.

2. Is there any issue in Pontoon with using a collection of many small Fluent files, rather than one big one with everything in? It could make life a lot easier for ID generation and overall performance if we have one Fluent file per component or even js file. How does Pontoon handle duplicated strings across files - if memory serves it has some kind of "this string is similar to this one already localised" functionality - so I hope it's not too bad for localisers.

3. I'm aware that the same string in different contexts in one locale can require multiple strings in another. Are there sensible assumptions that can be taken here (e.g. the same English string used twice in the same component is probably localised the same), or is it a hard requirement that each string be unique in the Fluent file? How tight is the feedback loop from Pontoon: will it be easy for localisers to ping us and tell us that we need two separate IDs in this context?

I think that's it for now: I've noted your points about using comments and brands - I've got some ideas for how to implement that with the inline string approach - thank you!

peiying2:
Hello @LeoMcA 

I will have my colleague @bcolsson to answer the more technical questions. I will address the questions below for the time being

>     2. Is there any issue in Pontoon with using a collection of many small Fluent files, rather than one big one with everything in? It could make life a lot easier for ID generation and overall performance if we have one Fluent file per component or even js file. How does Pontoon handle duplicated strings across files - if memory serves it has some kind of "this string is similar to this one already localised" functionality - so I hope it's not too bad for localisers.

Most of the Fluent based web projects have multiple files, such as Relay, Mozilla.org and Firefox.com. They tend to provide better context through the file names. It's easier to manage than having one big file.

Having a smaller file is a lot easier for ID generation. Both firefox.com and mozilla.org use the file name as part of the string ID and the first few words in the string to make up the string ID. Here is [an example](https://github.com/mozilla-l10n/www-firefox-l10n/blob/main/en/firefox/browsers/compare/edge.ftl).

Yes, Pontoon has a translation memory feature that shows localizers translations of identical or similar source content and rate them in % on the level of matches. Localizers can select the correct one based on context. Don't worry about duplicates.

>     3. I'm aware that the same string in different contexts in one locale can require multiple strings in another. Are there sensible assumptions that can be taken here (e.g. the same English string used twice in the same component is probably localised the same), or is it a hard requirement that each string be unique in the Fluent file? How tight is the feedback loop from Pontoon: will it be easy for localisers to ping us and tell us that we need two separate IDs in this context?

Preferably different ID for each of the strings even if they are the same. I remember a case where the same word is used for footer and navigation but in Turkish, they use different words for the same English words. 

Add context in the comments. If there are variables, provide examples how the variable would show up. If localizers have questions or need clarification, there is a comment field associated with each string. They usually ask questions and the l10n team will relay the question. 

You probably have seen [these documents](https://mozilla-l10n.github.io/documentation/localization/index.html) already.  It's worth checking it out or share with the team if you haven't.

peiying2:
@LeoMcA I want to share [a document](https://mozilla.github.io/bedrock/l10n/#fallback) the Bedrock team put together on writing code in Fluent with L10n in mind.

bcolsson:
>How much do IDs matter: you'll see I've made a slug out of the string, truncated it, and appended a hash (so e.g. "hello" and "Hello!" don't clash). Could I do away with the slug entirely and just use hashes? Or would it be best to not use hashes at all, and error in the case of clashes, telling the developer to manually specify the ID.

Speaking in best practices: IDs matter quite a bit to both those reviewing the source strings (us) and those translating the strings. They often provide important context to understand things like whether something is a button, header, etc. In other projects, we often use name spacing to both keep IDs unique, while also making things easier to find - especially when trying to find which instance of a common string (e.g. "Password" or "Sign in") is referring to.

Other challenges I see: 
- Handling updates to strings. If a string needs to change we need to update the string ID to ensure that translations are re-triggered. (See [here](https://mozilla-l10n.github.io/documentation/localization/dev_best_practices.html#string-identifiers-ids) for details.) Using hashes would conceivably ensure a new string is created, but how do you know what the old string was for removal? There's also times when you want to make a slight change to a string and not update translations (adding a comma/period or changing capitalization), and a hash would make that impossible. It also makes it hard for string reviewers to determine whether strings are new or updated (and again, if the string ID isn't labeled in the code then it's hard to find where the string is used).
- Touched on above: if a string is no longer used/removed in code, how would we know which strings are safe remove?
- Generating Fluent strings by extracting them from code would also mean that by default we aren't adding comments to them. 

My recommendation here would be to always manually specify IDs, name space them, but still have something check that there aren't duplicate IDs.

>Is there any issue in Pontoon with using a collection of many small Fluent files, rather than one big one with everything in? It could make life a lot easier for ID generation and overall performance if we have one Fluent file per component or even js file. How does Pontoon handle duplicated strings across files - if memory serves it has some kind of "this string is similar to this one already localised" functionality - so I hope it's not too bad for localisers.

In addition to what Peiying mentioned about translation memories and name spacing, I'll note that we need to have unique string IDs regardless if they are in different files. (So you can't have a `page-title` in File A and also a `page-title` in File B, this is where name spacing comes in handy.)

> I'm aware that the same string in different contexts in one locale can require multiple strings in another. Are there sensible assumptions that can be taken here (e.g. the same English string used twice in the same component is probably localised the same), or is it a hard requirement that each string be unique in the Fluent file? How tight is the feedback loop from Pontoon: will it be easy for localisers to ping us and tell us that we need two separate IDs in this context?

Best practices for localization strings is to prefer WET versus DRY code. These are hard errors to spot, especially if we aren't providing additional context via comments / string IDs.

peiying2:
@LeoMcA let us know if anything is unclear in our comments. We can also arrange a call to go over some of the issues if it's easier.

LeoMcA:
Apologies, I was wrapped up getting fred ready for launch.

All great advice: thank you. https://github.com/mdn/fred/pull/452 is a first step, making every string technically localisable, which we can build upon following your advice to improve the l10n experience.

PR #235, opened 2025-06-16
Use Lucide icons

pepelsbey:
### Description

- Replaces all kinds of icons with Lucide ones
- Uses icon names from Lucide (renames imports and variables)

Note:

- This is a step one, there are a few more icons left inlined into templates.
- Icons use strokes, not fills. Some inline styling might break, but masks are fine, though.


components/icon/global.css
```diff
@@ -36,6 +36,6 @@
 
   /* See https://github.com/mdn/rari/blob/927fac7cb89c7f7e49dd3ff265c3d74ad89d6490/crates/rari-doc/src/templ/templs/badges.rs#L43-L47 */
   &.icon-nonstandard {
-    mask-image: url("../notecard/icons/warning.svg");
+    mask-image: url("../notecard/icons/triangle-alert.svg");
```
caugner, line 39: @pepelsbey triangle-alert.svg is missing in this branch, explaining the following error:

```
SSR render error: Error: Cannot find module '../icon/triangle-alert.svg'
```

PR #251, opened 2025-06-18
fix(css): remove things from global, move into section component

LeoMcA:
- move definition-list into section, it was unscoped
- remove `.definition-list` styles, as the class doesn't exist: seems like a ux-components leftover
- remove `.formal-definition` styles, as the class doesn't exist: seems like a ux-components leftover
- remove generic-content styles, they were duplicated from the section component
- remove imports in generic-layout and reference-layout, and move them to the section component
- remove css-formal-syntax and notecard from the global styles, and move them into imports in the section component
- remove the sidebar filter global css, it does nothing

TBD:
are `badge/global.css` and `icon/global.css` truly global, or should they be scoped to particular components?


argl (approved):
LGTM

caugner:
> * remove the sidebar filter global css, it does nothing

That's because it uses the wrong color variable.

Edit: Fixed in 2bec754.

LeoMcA:
> That's because it uses the wrong color variable.

Also the wrong selector, as you fixed: but it still shouldn't be in a `global.css` file: I'll move it into `left-sidebar/server.css`

PR #254, opened 2025-06-19
fix: breadcrumbs on special pages

argl:
### Description

More breadcrumb support


components/breadcrumbs/server.js
```diff
@@ -2,32 +2,83 @@ import { html, nothing } from "lit";
...
+
+/**
+ * @typedef {object} DirectParentsContext
+ * @property {string} renderer The renderer to use
+ * @property {Parent[]} parents The parents to render
+ */
+
+/**
+ * @typedef {DocParentsContext | DirectParentsContext} BreadcrumbContext
+ */
```
caugner, line 25: I think we'd want to define these in a `types.d.ts` file.

components/breadcrumbs/server.js
```diff
@@ -2,32 +2,83 @@ import { html, nothing } from "lit";
...
+ * @returns {context is DirectParentsContext}
+ */
+function hasDirectParents(context) {
+  return [
+    "GenericDoc",
+    "SpaObservatoryLanding",
+    "SpaObservatoryAnalyze",
+    "GenericAbout",
+    "GenericCommunity",
+  ].includes(context.renderer);
```
caugner, line 55: Why not just:

```suggestion
  return "parents" in context;
```

components/breadcrumbs/server.js
```diff
@@ -2,32 +2,83 @@ import { html, nothing } from "lit";
...
+ */
+function hasDocParents(context) {
+  return [
+    "Doc",
+    "CurriculumModule",
+    "CurriculumLanding",
+    "CurriculumOverview",
+    "CurriculumAbout",
+    "CurriculumDefault",
+  ].includes(context.renderer);
```
caugner, line 40: Why not just:

```suggestion
  return "doc" in context && "parent" in context.doc;
```

components/heading-anchor/server.js
```diff
@@ -11,6 +11,8 @@ export class HeadingAnchor extends ServerComponent {
    * @param {string} title
    */
   render(level, id, title) {
+    // remove leading numbers, dots, underscores, and commas to form a valid element id
+    id = id ? id.replace(/^[0-9._,]+/, "") : id;
```
caugner, line 15: Shouldn't this happen in rari?

components/breadcrumbs/server.js
argl: ah nice, I'll have a look

PR #255, opened 2025-06-19
Fix icons fills

pepelsbey:
### Description

A few low-hanging ones


components/color-theme/element.css
```diff
@@ -84,10 +84,6 @@
   &:hover {
     background-color: var(--color-background-secondary);
   }
-
-  svg {
-    pointer-events: none;
-  }
```
caugner: Needs to be reverted, so that themes can be selected by clicking on the icon. Otherwise the SVG prevents the event from bubbling.

components/icon/README.md
caugner, line 1: This README seems incomplete, did you intend to check it in here?

pepelsbey: It’s a draft, work in progress

PR #276, opened 2025-06-20
refactor(compat-table): update icons

pepelsbey:
### Description

- Table now uses Lucide icons
- Browser icons now in the global `icon/browser` folder (also, optimized)
- Experimental icons are now contrast by default
- Timeline rendering is improved
- Timeline’s empty circle now uses icon as well

### Preview

image


components/compat-table/index-common.css
```diff
@@ -344,38 +346,21 @@ table {
...
-      z-index: -1;
-
-      width: 2px;
-      height: calc(100% - 6px);
-
-      content: "";
-
-      background: var(--color-background-primary);
-    }
-  }
```
caugner: Could this removal have caused the grey line to continue beyond the last item?

To reproduce, expand the Chrome or Opera cell for `alert` in the `Window` BCD table:
https://fred-pr276.review.mdn.allizom.net/en-US/docs/Web/API/Window#browser_compatibility

caugner (commented):
LGTM overall, but there seems to be a regression.

Could you extract the fixes and optimizations that go beyond replacing the icons, into a separate PR?

components/compat-table/index-common.css
pepelsbey: Should be fixed now

caugner (approved):
Thank you, LGTM, and the regression is fixed.

PR #280, opened 2025-06-23
fix(placement): add focus-visible to sidebar

fiji-flo:
### Description

Add focus visible to placement so focus does not vanish. Add padding to make it look nice.


caugner (approved):
LGTM, based on comparing this PR with main.

PR #287, opened 2025-06-24
fix(homepage-hero): avoid non-link underline

pepelsbey:
### Description

Only links should be underlined

### DIFF
--- a/l10n/en-US.ftl
+++ b/l10n/en-US.ftl
@@ -12,7 +12,7 @@ baseline-supported-in = Supported in { $browsers }
 baseline-unsupported-in = Not widely supported in { $browsers }
 baseline-supported-and-unsupported-in = Supported in { $supported }, but not widely supported in { $unsupported }
 
-homepage-hero-title = Resources for <u data-l10n-name="developers">Developers</u>,<br> by Developers
+homepage-hero-title = Resources for Developers,<br /> by Developers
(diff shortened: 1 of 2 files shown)


l10n/en-US.ftl
```diff
@@ -12,7 +12,7 @@ baseline-supported-in = Supported in { $browsers }
 baseline-unsupported-in = Not widely supported in { $browsers }
 baseline-supported-and-unsupported-in = Supported in { $supported }, but not widely supported in { $unsupported }
 
-homepage-hero-title = Resources for <u data-l10n-name="developers">Developers</u>,<br> by Developers
```
caugner, line 15: Let's make it `<em>` and style it with CSS.

l10n/en-US.ftl
```diff
@@ -12,7 +12,7 @@ baseline-supported-in = Supported in { $browsers }
...
+homepage-hero-title = Resources for Developers,<br /> by Developers
```
caugner: Avoid closing void tags:

```suggestion
homepage-hero-title = Resources for Developers,<br> by Developers
```

l10n/en-US.ftl
caugner: Never mind.

caugner (approved):

l10n/en-US.ftl
pepelsbey: @caugner, isn’t it what we do with Prettier by default for all MDN content/code?

See examples https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/br#try_it

PR #309, opened 2025-07-01
enhance(search-modal): update blur, mark, links

pepelsbey:
### Description

- Fixes blur: backdrop filter doesn’t work with solid background color
- Sets background color to dialog: otherwise, it had a default browser styles tint of black
- Uses link colors from the design system
- Adjusts the mark’s yellow color to be more visible
- Uses font size value from the design system
- Styles progress and renders it outside of the list
- Sets `aria-label`, the placeholder is not enough
- Uses pseudo-element for icon
- Contains scrolling inside the list, so the page won’t move

### DIFF
--- a/components/search-modal/element.js
+++ b/components/search-modal/element.js
@@ -199,12 +199,9 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
         ${this._queryIndex.render({
           initial: this._renderLoadingSearchIndex.bind(this),
           pending: this._renderLoadingSearchIndex.bind(this),
-          complete: () => nothing,
         })}
         <ul>
           ${this._queryIndex.render({
-            initial: () => nothing,
-            pending: () => nothing,
             complete: (results) =>
               results?.map(
                 ({ title, url }, i) => html`


components/search-modal/element.js
```diff
@@ -187,19 +187,24 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
...
             @input=${this._input}
             placeholder=${this.l10n`Search`}
+            aria-label=${this.l10n`Search`}
           />
         </form>
+        ${this._queryIndex.render({
+          initial: this._renderLoadingSearchIndex.bind(this),
+          pending: this._renderLoadingSearchIndex.bind(this),
+          complete: () => nothing,
+        })}
```
pepelsbey: There might be a better way of rendering progress outside the list 🤔

caugner: I think that's fine. 👍 

(Hadn't thought of reusing the task twice.)

components/search-modal/element.js
```diff
@@ -187,19 +187,24 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
...
         <ul>
           ${this._queryIndex.render({
-            initial: this._renderLoadingSearchIndex.bind(this),
-            pending: this._renderLoadingSearchIndex.bind(this),
+            initial: () => nothing,
+            pending: () => nothing,
```
caugner: (Nit) This should be equivalent to the following:

```suggestion
```

components/search-modal/element.js
```diff
@@ -187,19 +187,24 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
...
             autofocus
             @input=${this._input}
             placeholder=${this.l10n`Search`}
+            aria-label=${this.l10n`Search`}
           />
         </form>
+        ${this._queryIndex.render({
+          initial: this._renderLoadingSearchIndex.bind(this),
+          pending: this._renderLoadingSearchIndex.bind(this),
+          complete: () => nothing,
```
caugner: (Nit) Should be equivalent to the following:

```suggestion
```

PR #311, opened 2025-07-01
ci(workflows): add npm-publish-simulation workflow

caugner:
### Description

Adds `npm-publish-simulation` workflow.

### Motivation

Test fred in the context of the content repo.

### Additional details

See: https://github.com/mdn/yari/blob/main/.github/workflows/npm-published-simulation.yml


.github/workflows/npm-publish-simulation.yml
```diff
@@ -0,0 +1,85 @@
+name: NPM Publish simulation
+
+on:
+  pull_request:
+    branches:
+      - main
```
caugner: There isn't really a reason to only run this on PRs targeting `main`:

```suggestion
```

PR #324, opened 2025-07-04
chore(deps-dev): add lit-analyzer

caugner:
### Description

Adds `@jackolope/lit-analyzer`, and runs it with lefthook in the `pre-commit` and `pre-push` hooks, and as part of the `test` workflow.

### Motivation

Improve our Lit code quality.

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,3 @@
+{
+  "recommendations": ["jackolope.lit-analyzer-plugin"]
+}
--- a/components/not-found/element.js
+++ b/components/not-found/element.js
@@ -77,7 +77,7 @@ export class MDNNotFound extends L10nMixin(LitElement) {
               .reverse();
 
             return html`<div class="notecard note">
-              <p>${this.l10n("not-found-fallback-search")}
+              <p>${this.l10n("not-found-fallback-search")}</p>
               <ul>
                 ${normalizedLocationParts.map(
                   (part) =>
@@ -89,7 +89,7 @@ export class MDNNotFound extends L10nMixin(LitElement) {
                       </a>
                     </li>`,
                 )}
-              </ul></p>
+              </ul>
             </div>`;
           }
         },
--- a/components/not-found/element.js
+++ b/components/not-found/element.js
@@ -76,20 +76,21 @@ export class MDNNotFound extends L10nMixin(LitElement) {
               .map((part) => part.replaceAll("_", " "))
               .reverse();
 
+            const fallback = html`<ul>
+              ${normalizedLocationParts.map(
+                (part) =>
+                  html`<li>
+                    <a
+                      href=${`/${locale}/search?q=${encodeURIComponent(part)}`}
+                    >
+                      <code>${part}</code>
+                    </a>
+                  </li>`,
+              )}
+            </ul>`;
+
             return html`<div class="notecard note">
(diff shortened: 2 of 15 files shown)


caugner:
FYI @pepelsbey It looks like lit-analyzer has some bugs like https://github.com/runem/lit-analyzer/issues/239, which prevents us from adopting it.

caugner:
@pepelsbey It turns out there is a fork (`@jackolope/lit-analyzer`) that we can use.

It has the same issue with the sandbox attribute, not allowing multiple values, but we can just disable that rule for now.

.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
+  "recommendations": ["jackolope.lit-analyzer-plugin"]
```
LeoMcA, line 2: I'm not sure this does anything which `ts-lit-plugin` doesn't do

components/not-found/element.js
```diff
@@ -89,7 +89,7 @@ export class MDNNotFound extends L10nMixin(LitElement) {
                       </a>
                     </li>`,
                 )}
-              </ul></p>
+              </ul>
```
LeoMcA: Not sure I see the issue here: `<p><ul>...</ul></p>` is perfectly valid html, no? Was this just a manual change?

package.json
```diff
@@ -14,6 +14,7 @@
     "build": "rspack build",
     "preview": "NODE_ENV=production node ./server.js",
     "ssr": "node build/ssr.js",
+    "lit-analyzer": "npx @jackolope/lit-analyzer --rules.no-incompatible-type-binding off",
```
LeoMcA, line 17: You shouldn't need to use npx in the scripts

```suggestion
    "lit-analyzer": "@jackolope/lit-analyzer --rules.no-incompatible-type-binding off",
```

LeoMcA (approved):
lgtm, a few questions, but feel free to merge if you're satisfied:

.vscode/extensions.json
caugner: Just tested: The VSCode extension shows errors like `<a href=${new URL()"}>Example</a>`. Without, they don't show up.

caugner: Not sure why but it complains:

```
./components/not-found/element.js

    This tag isn't closed.
    80:  <p>${this.l10n("not-found-f
    no-unclosed-tag


  ✖ 1 problem in 1 file (1 error, 0 warnings)
```

components/not-found/element.js
caugner: Not sure why but it complains:

```
./components/not-found/element.js

    This tag isn't closed.
    80:  <p>${this.l10n("not-found-f
    no-unclosed-tag


  ✖ 1 problem in 1 file (1 error, 0 warnings)
```

caugner: Seems like a bug in the analyzer.

package.json
caugner:
```
% npm run lit-analyzer

> @mdn/fred@0.0.5 lit-analyzer
> @jackolope/lit-analyzer --rules.no-incompatible-type-binding off

sh: @jackolope/lit-analyzer: No such file or directory
```

components/not-found/element.js
caugner: As a workaround, I now extracted the list into a variable: bb359d3

PR #326, opened 2025-07-04
feat(button, blog): add iconPosition attribute, use icon arrows

pepelsbey:
### Description

- Add `iconPosition` attribute with `before` and `after` values
- Use an icon instead of the text arrow in the blog


components/button/pure.js
```diff
@@ -11,21 +11,29 @@ import { ifDefined } from "lit/directives/if-defined.js";
  * @param {string} [options.target]
  * @param {import("./types.js").ButtonVariants} [options.variant]
  * @param {import("./types.js").ButtonActions} [options.action]
+ * @param {import("./types.js").ButtonIconPositions} [options.iconPosition]
```
caugner: Nit: Usually one would use the same order in the JSDoc as below.

components/button/pure.js
```diff
@@ -11,21 +11,29 @@ import { ifDefined } from "lit/directives/if-defined.js";
...
-  ];
+  const iconElement = icon ? html`<span class="icon">${icon}</span>` : nothing;
+  const labelElement = html`<span id="label" class="label" ?hidden=${iconOnly}
+    >${label}</span
+  >`;
+
+  const inner =
+    icon && iconPosition === "after"
+      ? [labelElement, iconElement]
+      : [iconElement, labelElement];
```
caugner: Nit: If we check for `icon`, I would expect just `labelElement` if `icon` is not set. Otherwise we can just omit this check:

```suggestion
    iconPosition === "after"
      ? [labelElement, iconElement]
      : [iconElement, labelElement];
```

components/button/types.d.ts
```diff
@@ -1,2 +1,3 @@
 export type ButtonVariants = "primary" | "secondary" | "plain";
 export type ButtonActions = "positive" | "negative" | undefined;
+export type ButtonIconPositions = "before" | "after";
```
caugner, line 3: Nit: Usually one would use singular:

```suggestion
export type ButtonIconPosition = "before" | "after";
```

But we can refactor that later/separately together with the other two.

components/button/element.js
```diff
@@ -26,6 +27,8 @@ export class MDNButton extends LitElement {
     this.variant = "primary";
     /** @type {import("./types.js").ButtonActions} */
     this.action = undefined;
+    /** @type {import("./types.js").ButtonIconPositions} */
+    this.iconPosition = "before";
```
caugner: Nit: Would expect this to come directly after `this.iconOnly = false` above.

components/blog-index/server.js
```diff
@@ -50,8 +51,10 @@ function PostPreview(context, blogMeta) {
         ? html`<span class="sponsored">Sponsored</span>`
         : nothing}
       ${Button.render(context, {
-        label: "Read more →",
+        label: "Read more",
```
caugner, line 54: Nit: We have at least two more cases, so maybe we can fix these together separately?

pepelsbey: Good point! Fixed.

PR #333, opened 2025-07-07
chore(code-example): reduce font-size

Rumyra:
### Description

This reduces the font size in the code examples (and matches that in the interactive examples)

I have used a value and not a custom prop - this may need some discussion (ie do we want a `--font-size-code`?)

### DIFF
--- a/components/code-example/common.css
+++ b/components/code-example/common.css
@@ -33,6 +33,7 @@
 
     code {
       padding: 0;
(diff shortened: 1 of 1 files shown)


Rumyra:
I'll come back to this when we have a code font

Rumyra:
Closing in favour of replaced font and font-size-adjust which will be opened soon

PR #338, opened 2025-07-08
enhance(site-search): redesign

argl:
### Description

Adding styles and pagination to the site-search result

### Motivation

Paritiy with yari


components/search/server.css
caugner, line 1: Let's move `search/server.{css,js}` into `site-search`, which is probably where it belongs.

components/search/server.css
```diff
@@ -0,0 +1,14 @@
...
+
+  @media (--screen-layout-no-sidebar) {
+    grid-template-columns: var(--layout-no-sidebar);
+  }
+}
+
+mdn-site-search {
+  display: grid;
+  grid-column: content;
+}
```
caugner: This should probably go inside the `.site-search` block.

components/site-search/element.js
```diff
@@ -1,20 +1,75 @@
...
+  fr: "Français",
+  es: "Español",
+  it: "Italiano",
+  ja: "日本語",
+  ko: "한국어",
+  "pt-br": "Português (BR)",
+  ru: "Русский",
+  "zh-cn": "简体中文",
+  "zh-tw": "繁體中文",
+};
```
caugner, line 33: We should probably centralize this somewhere.

components/site-search/element.js
```diff
@@ -1,20 +1,75 @@
...
+ * @returns {string}
+ */
+function mdnUrl2Breadcrumb(url, locale) {
+  return url
+    .replaceAll("_", " ")
+    .split("/")
+    .slice(1)
+    .filter((p) => ![locale, "docs", "Web"].includes(p))
+    .join(" / ");
+}
```
caugner: We should put this somewhere central and reuse it in `components/search-modal/element.js`:

https://github.com/mdn/fred/blob/51fcf0ed4ff91afa5d1aba85f5eca842d55e45fd/components/search-modal/element.js#L214-L221

caugner (commented):
Partial review.

components/site-search/element.js
argl: Thought the same, but could not think of a proper place. The items in the language switcher are coming out of rari responses afaik, so this is the only place where this data is needed.

PR #353, opened 2025-07-10
enhance(content-section): use design colors

pepelsbey:
### Description

Use design colors, looks much better in dark scheme

### DIFF
--- a/components/content-section/server.css
+++ b/components/content-section/server.css
@@ -80,9 +80,7 @@
 
     margin: 2rem auto;
 
-    color: #1b1b1b;
-
-    background-color: #ffffff;
+    background-color: var(--color-background-secondary);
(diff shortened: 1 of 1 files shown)


Rob--W:
This change caused readability (contrast) issues on images that are partially transparent and not designed for use with dark backgrounds.

E.g. https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest
Screenshot showing poor contrast in diagram of webRequest API

Another similar example: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webNavigation

pepelsbey:
@Rob--W Could you please file an issue basically quoting your comment? Thank you!

PR #355, opened 2025-07-10
fix(external-link): icon alignment

pepelsbey:
### Description

Added `vertical-align: middle`, otherwise non-empty `content` pushes it up

### DIFF
--- a/components/external-link/global.css
+++ b/components/external-link/global.css
@@ -1,21 +1,21 @@
-.external {
-  &::after {
-    display: inline-block;
+.external::after {
+  display: inline-block;
 
-    width: 1em;
-    height: 1em;
+  width: 1em;
+  height: 1em;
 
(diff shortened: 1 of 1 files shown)


components/external-link/global.css
```diff
@@ -1,21 +1,21 @@
...
 
-    margin-left: 0.25em;
+  margin-left: 0.25em;
 
-    /*
-      Unicode zero width space as `<content replacement>` to expose the alt text in the a11y tree.
-      See: https://bugzilla.mozilla.org/show_bug.cgi?id=1976574
-    */
-    content: "\200b" / " (external)";
+  vertical-align: middle;
```
caugner: Looks better (even if it doesn't work in the About menu):
```suggestion
  vertical-align: text-top;
```

PR #358, opened 2025-07-10
enhance(notecard): add color to border + icon

caugner:
### Description

Adds color to the left border and the icon of notecards.

### Motivation

Parity with yari, as we have not decided to change this.

### Additional details

| Before | After |
|--------|--------|
| image | image |
| image | image |
| image | image |
| image | image | 
| image | image |

See:

- http://localhost:3000/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#notes_warnings_and_callouts

### DIFF
--- a/components/notecard/index.css
+++ b/components/notecard/index.css
@@ -10,23 +10,27 @@
   color: var(--color-text-primary);
 
   background-color: var(--color-background-secondary);
-  border-inline-start: 2px solid var(--color-text-secondary);
+  border-inline-start: 2px solid
+    var(--accent-color, var(--color-text-secondary));
   border-radius: 0.25em;
 
   &.note,
   &.secure,
   &.experimental {
     background-color: var(--color-background-blue);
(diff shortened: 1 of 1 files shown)


components/notecard/index.css
```diff
@@ -10,23 +10,27 @@
...
   &.experimental {
     background-color: var(--color-background-blue);
+    --accent-color: var(--color-text-blue);
   }
 
   &.warning,
   &.deprecated,
   &.nonstandard {
     background-color: var(--color-background-red);
+    --accent-color: var(--color-text-red);
```
caugner, line 28: @Rumyra @pepelsbey Should the warning accent color be yellow, like [on prod](https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#warnings)?

image

caugner: As discussed in today's check-in, we go with red border and icon for warnings.

PR #363, opened 2025-07-11
feat(site-search): implement site search suggestions

argl:
### Description

Implementing the suggestions when passed down from the search backend.

### Motivation

Parity with yari

### Additional details

Before:
image

After:
image

### DIFF
--- a/components/site-search/element.css
+++ b/components/site-search/element.css
@@ -3,14 +3,10 @@
 
   a {
     color: var(--color-link-normal);
-
-    /* text-decoration: none; */
   }
 
   a:visited {
     color: var(--color-link-visited);
-
-    /* text-decoration: none; */
   }
 
   a:hover,
--- a/components/site-search/element.css
+++ b/components/site-search/element.css
@@ -1,17 +1,21 @@
 :host {
+  line-height: var(--font-line-content);
+
   a {
     color: var(--color-link-normal);
-    text-decoration: none;
+
+    /* text-decoration: none; */
   }
 
   a:visited {
     color: var(--color-link-visited);
-    text-decoration: none;
+
+    /* text-decoration: none; */
   }
 
   a:hover,
   a:active {
-    text-decoration: underline;
+    text-decoration: none;
   }
   --color-highlight-background: light-dark(
     var(--color-yellow-80),
@@ -98,6 +102,43 @@
   font-weight: 400;
 }
 
+.site-search-suggestions__list {
+  padding: 0;
+  padding-inline: 0;
+}
+
+.site-search-suggestions__item {
+  &::before {
+    display: inline-block;
+
+    width: 1em;
+    height: 1em;
+
+    margin-right: 0.1em;
+
+    vertical-align: middle;
+
+    content: "";
+
+    background-color: var(--color-text-secondary);
+
+    mask-image: url("./../icon/arrow-right.svg");
+    mask-repeat: no-repeat;
+    mask-position: center;
+    mask-size: contain;
+  }
+  list-style-type: none;
+}
+
+.site-search-suggestions__text {
+  margin: 1rem 0;
+}
+
+.site-search-suggestions__matches {
+  font-size: var(--font-size-small);
+  color: var(--color-text-secondary);
+}
+
 .site-search-form {
   display: flex;
   flex-direction: column;
@@ -138,11 +179,11 @@
   article {
     display: grid;
     grid-template-areas:
-      "title"
       "path"
+      "title"
       "description";
   }
-  margin-bottom: 1rem;
+  margin-bottom: 1.5rem;
 }
 
 .site-search-results__title {
@@ -157,10 +198,12 @@
 
 .site-search-results__path {
   grid-area: path;
-  margin: 0 0 0.25rem;
+
+  margin: 0;
+
   font-size: var(--font-size-small);
 
-  /* color: var(--color-text-secondary); */
+  color: var(--color-text-secondary);
 }
 
 .site-search-results__description {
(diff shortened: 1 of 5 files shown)


components/site-search/element.js
```diff
@@ -288,99 +288,131 @@ export class MDNSiteSearch extends L10nMixin(LitElement) {
...
+                        )`Your search did not yield any results. Did you mean…`}
+                      </p>
+                      <ul class="site-search-suggestions__list">
+                        ${results.suggestions.map((suggestion) => {
+                          const url = new URL(location.href);
+                          url.searchParams.delete("page");
+                          url.searchParams.set("q", suggestion.text);
+                          return html`
+                            <li class="site-search-suggestions__item">
+                              <a class="site-search-suggestions__link" href=${url.href}> ${suggestion.text} </a>
```
caugner: Nit: Are those spaces intended?

```suggestion
                              <a class="site-search-suggestions__link" href=${url.href}>${suggestion.text}</a>
```

components/site-search/element.css
```diff
@@ -98,6 +98,43 @@
...
+
+    margin-right: 0.1em;
+
+    vertical-align: middle;
+
+    content: "";
+
+    background-color: var(--color-text-secondary);
+
+    mask-image: url("./../icon/arrow-right.svg");
```
caugner, line 121: Do we need this arrow? Can't we just use bullet points like on prod (and on the 404 page)?

components/site-search/element.js
```diff
@@ -288,99 +288,131 @@ export class MDNSiteSearch extends L10nMixin(LitElement) {
...
+                  args: {
+                    results: results.metadata.total.value,
+                  },
+                })}</p>
+                <ul class="site-search-results">
+                  ${results.documents.map(
+                    (result) =>
+                      html`<li class="site-search-results__item">
+                        <article>
                           <p class="site-search-results__path">
```
caugner, line 358: If we move the path above the `h2`, we should probably stop using an `h2` below?

components/site-search/element.js
```diff
@@ -288,99 +288,131 @@ export class MDNSiteSearch extends L10nMixin(LitElement) {
...
+                            <li class="site-search-suggestions__item">
+                              <a class="site-search-suggestions__link" href=${url.href}> ${suggestion.text} </a>
+                              <span class="site-search-suggestions__matches">
+                              ${this.l10n.raw({
+                                id: "suggestion-matches",
+                                args: {
+                                  matches: suggestion.total.value,
+                                  relation: suggestion.total.relation,
+                                },
+                              })}
```
caugner: Missing closing `</span>`?

components/site-search/element.js
```diff
@@ -288,99 +288,131 @@ export class MDNSiteSearch extends L10nMixin(LitElement) {
...
+              <h1 class="site-search__title">
+                ${this.l10n.raw({
+                  id: "search-title",
+                  args: {
+                    query: this._query,
+                  },
+                })}
+              </h1>
+
+              ${results.suggestions.length > 0 && results.documents.length === 0
```
caugner: This means we don't show suggestions if there are results:

- Fred: https://fred-pr363.review.mdn.allizom.net/en-US/search?q=svg1&page=1
- Prod: https://developer.mozilla.org/en-US/search?q=svg1&page=1

l10n/de.ftl
```diff
@@ -8,6 +8,22 @@ footer-tagline = Dein Bauplan für ein besseres Internet.
 footer-mofo = Besuche die gemeinnützige Muttergesellschaft der <a data-l10n-name="moco">Mozilla Corporation</a>, die <a data-l10n-name="mofo">Mozilla Foundation</a>.
 footer-copyright = Teile dieses Inhalts sind ©1998–2024 von einzelnen mozilla.org-Mitwirkenden. Inhalte sind verfügbar unter <a data-l10n-name="cc">einer Creative-Commons-Lizenz</a>.
 
+search-title = Suchergebnisse für: <em>{ $query }</em>
+search-stats = { $results } Dokumente gefunden.
+search-modal-site-search = Erweiterte Suche nach <em>{ $query }</em>
+suggestion-matches =  { $relation ->
+    [gt] mehr als{ $matches ->
```
caugner: Missing a space?

```suggestion
    [gt] mehr als { $matches ->
```

l10n/en-US.ftl
```diff
@@ -27,9 +27,19 @@ footer-mofo = Visit <a data-l10n-name="moco">Mozilla Corporation’s</a> not-for
...
+    [gt] more than { $matches ->
+        [one]   { $matches } match
+       *[other] { $matches } matches
+    }
+   *[eq] { $matches ->
+        [one]   { $matches } match
+       *[other] { $matches } matches
+    }
+}
+search-suggestions-text = Your search did not yield any results. Did you mean:
```
caugner: We didn't discuss this, but it would be nice prefixing the keys with the component name:

```suggestion
site-search-suggestion-matches =  { $relation ->
    [gt] more than { $matches ->
        [one]   { $matches } match
       *[other] { $matches } matches
    }
   *[eq] { $matches ->
        [one]   { $matches } match
       *[other] { $matches } matches
    }
}
site-search-search-suggestions-text = Your search did not yield any results. Did you mean:
```

components/site-search/element.css
argl: We rarely use naked bullet points, but I can tweak it a bit more. Since the site search is heavily redesigned anyway, I do not see us bound to the yari version in this instance.

components/site-search/element.js
argl: I was not aware that this is a valid reponse, thanks.

components/site-search/element.js
argl: I'll move tha path via css and move it below the h2 in the markup (like the original version).

PR #364, opened 2025-07-11
enhance(sidebar-filter): use mask icon, improve focus

pepelsbey:
### Description

- Use the mask icon
- Remove background from the hovered clear button
- Use native outline
- Improve the placeholder’s contrast
- Make the input and the button pill-shaped
- Align the icon better

### Before/after

before-after

### DIFF
--- a/components/sidebar-filter/element.css
+++ b/components/sidebar-filter/element.css
@@ -4,101 +4,102 @@
   position: relative;
   display: flex;
   flex-direction: column;
+}
 
-  .sidebar-filter {
-    display: flex;
-    place-items: center;
+.sidebar-filter {
+  display: flex;
+  place-items: center;
+}
+
+.sidebar-filter-label {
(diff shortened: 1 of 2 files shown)


caugner (commented):
Changes LGTM.

The outline is cut on the left (also on your screenshots), but maybe that's acceptable?
image

caugner:
> The outline is cut on the left (also on your screenshot), but maybe that's acceptable?

Adding `.left-sidebar { padding-left: 2px; }` resolves the issue, and would also allow us to use `box-shadow` on the highlighted sidebar item, like we do in the TOC.

pepelsbey:
> The outline is cut on the left (also on your screenshot), but maybe that's acceptable?

I would fix it in a separate PR. Maybe even change the overflow scope so it won’t cut off outlines and shadows.

PR #334, opened 2025-07-07
enhance(code-examples): relax requirements

caugner:
### Description

Stop requiring outer `div.code-example` and inner `code` wrapper, and add "plain" to language classes.

### Motivation

The plain code example on [this page](https://fred.review.mdn.allizom.net/en-US/docs/Learn_web_development/Getting_started/Your_first_website/Creating_the_content#so_what_is_html) wasn't rendered properly.

PR #365, opened 2025-07-11
fix(content-section): style plain pre tags

LeoMcA:
Alternative to https://github.com/mdn/fred/pull/334.

We shouldn't "upgrade" these plain pre tags to code blocks without changes in rari to ensure they have a header, so we don't get content shifts.

So for now, just style them like we do on prod:

## Current fred:

Screenshot 2025-07-11 at 14-08-16 HTML Creating the content

## Current prod:

Screenshot 2025-07-11 at 14-08-00 HTML Creating the content - Learn web development MDN

## New fred:

Screenshot 2025-07-11 at 14-16-36 HTML Creating the content

### DIFF
--- a/components/content-section/server.css
+++ b/components/content-section/server.css
@@ -66,13 +66,25 @@
     }
   }
 
+  pre {
+    padding: 1rem;
+    overflow: auto;
+
+    &:not(.code-example &) {
+      border: 1px solid var(--color-border-primary);
+      border-radius: 0.25em;
+    }
+  }
+
   code {
     padding: 0.125em 0.25em;
-
-    font-family: var(--font-family-code);
-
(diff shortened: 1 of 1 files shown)


components/content-section/server.css
```diff
@@ -66,13 +66,25 @@
     }
   }
 
+  pre {
+    padding: 1rem;
```
caugner, line 70: Nit: We seem to be using `em` below (in the `code {}` block):

```suggestion
    padding: 1em;
```

caugner (approved):
LGTM, just one nit.

components/content-section/server.css
caugner: Let's merge for now to include it in a fred release.

PR #368, opened 2025-07-11
fix(blog): responsiveness

argl:
### Description

Fixed 1-col / 2-col responsiveness on the blog index page

### Motivation

Yari parity.

Before:
image

After:
image

### DIFF
--- a/components/blog-index/server.css
+++ b/components/blog-index/server.css
@@ -59,6 +59,10 @@
   display: grid;
   grid-template-columns: 1fr 1fr;
   column-gap: 2rem;
+
+  @media (--screen-large-and-narrower) {
+    grid-template-columns: auto;
+  }
 }
 
 .blog-post-preview {
@@ -113,7 +117,7 @@
     display: flex;
 
     width: auto;
-    height: 200px;
+    max-height: 200px;
 
     margin: 0 0 0.125rem;
     margin-bottom: 0;
--- a/components/blog-index/server.js
+++ b/components/blog-index/server.js
@@ -12,13 +12,12 @@ import { ServerComponent } from "../server/index.js";
  * @param {object} params
  * @param {import("@rari").BlogImage} params.image
  * @param {string} params.slug
- * @param {number} params.width
  * @param {number} params.height
  * @param {boolean} params.lazyLoad
  */
 export function BlogIndexImageFigure(
   _context,
-  { image, slug, width, height, lazyLoad },
+  { image, slug, height, lazyLoad },
 ) {
   const src = `/en-US/blog/${slug}/${image.file}`;
   return html`<figure class="blog-post-preview__figure">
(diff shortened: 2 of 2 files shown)


components/blog-index/server.js
```diff
@@ -46,8 +44,7 @@ function PostPreview(context, blogMeta, lazyLoad = false) {
       ${BlogIndexImageFigure(context, {
         image: blogMeta.image,
         slug: blogMeta.slug,
-        width: 1200,
-        height: 630,
+        height: 200,
```
caugner: I _think_ @pepelsbey just updated these values.

components/blog-index/server.css
```diff
@@ -113,7 +117,7 @@
     display: flex;
 
     width: auto;
-    height: 200px;
+    max-height: 200px;
```
caugner: Would be good if @pepelsbey can take a look, as this might re-introduce layout shift?

caugner (commented):
Code changes LGTM overall, but would be good if @pepelsbey takes a look.

components/blog-index/server.css
```diff
@@ -59,6 +59,10 @@
   display: grid;
   grid-template-columns: 1fr 1fr;
   column-gap: 2rem;
+
+  @media (--screen-large-and-narrower) {
```
caugner: Maybe should be medium and narrower, or else we see this at 1200px width:

image

```suggestion
  @media (--screen-medium-and-narrower) {
```

pepelsbey (commented):
I didn’t have a chance to mention this PR at today’s sync. But @caugner is right, the layout shift is back with the mentioned changes. See https://github.com/mdn/fred/pull/357

components/blog-index/server.css
pepelsbey: Good catch, `height: 200px` is enough

components/blog-index/server.js
pepelsbey: The source images are 1200x630 and we should keep these values in the attributes to help the browser properly size a responsive image. Correct sizing is done in CSS (`width: auto`, `height: 200px`).

components/blog-index/server.css
caugner: I think `height: 200px` causes an overflow at small screens (<500px width).

argl:
> I didn’t have a chance to mention this PR at today’s sync. But @caugner is right, the layout shift is back with the mentioned changes. See #357

I noticed that the images were squashed horizontally, I'll check

components/blog-index/server.css
argl: see below

components/blog-index/server.js
argl: Right, I reverted this. I was seeing some horizontal squashing of images at some widths, but that is gone with the single column display on smaller sizes.

PR #369, opened 2025-07-11
enhance(sidebars): use learn area colours in learn area

LeoMcA:
Can be extended further in future with other areas.

There's a discrepancy between the "brand" link colour we use in the sidebar links on hover, and the one we use in the top nav.

Perhaps these should be called `--color-section` or `--color-area` instead of `--color-brand`.

### DIFF
--- a/components/color/area.css
+++ b/components/color/area.css
@@ -1,13 +1,11 @@
 :root {
   --color-area-background: var(--color-background-blue);
-  --color-area-text: var(--color-text-blue);
   --color-area-highlight-border: var(--color-blue-50);
   --color-area-link: var(--color-link-normal);
 }
 
-:root[data-current-section="learn"] {
+:root[data-current-area="learn"] {
   --color-area-background: var(--color-background-orange);
-  --color-area-text: var(--color-text-orange);
   --color-area-highlight-border: var(--color-orange-50);
   --color-area-link: light-dark(var(--color-orange-20), var(--color-orange-80));
 }
--- a/components/color/global.css
+++ b/components/color/global.css
@@ -5,7 +5,7 @@
 @import url("text.css");
 @import url("link.css");
 @import url("scrollbar.css");
-@import url("brand.css");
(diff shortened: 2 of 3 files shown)


components/color/brand.css
```diff
@@ -0,0 +1,26 @@
...
+:root[data-current-section="learn"] {
+  --color-brand-background: var(--color-brand-learn-background);
+  --color-brand-text: var(--color-brand-learn-text);
+  --color-brand-highlight-border: var(--color-brand-learn-highlight-border);
+
+  --color-brand-link: light-dark(
+    var(--color-orange-20),
+    var(--color-orange-80)
+  );
+}
```
pepelsbey:
```suggestion
:root {
  --color-brand-background: var(--color-background-blue);
  --color-brand-text: var(--color-text-blue);
  --color-brand-highlight-border: var(--color-blue-50);

  --color-brand-link: light-dark(
    var(--color-blue-20),
    var(--color-blue-80)
  );
}

:root[data-current-section="learn"] {
  --color-brand-background: var(--color-background-orange);
  --color-brand-text: var(--color-text-orange);
  --color-brand-highlight-border: var(--color-orange-50);

  --color-brand-link: light-dark(
    var(--color-orange-20),
    var(--color-orange-80)
  );
}
```

I would suggest simplifying it: set the default and then redefine it based on the current section. It would be much easier to read and maintain.

Also, I wouldn’t make the default brand link color dependent on the link variable and use the colors directly, just like you do for non-blue ones. Semantically, they are different, even if they match.

It requires some fixing where it’s used.

LeoMcA: Sure, simplified things further, removing the `--color-brand-text` variable because that was only used in the top nav. Kept the `--color-area-link: var(--color-link-normal);` because IMO they are the same semantically: our default area link colour is the global link colour, and in some areas it's different.

Rumyra: This all sounds good to me

PR #370, opened 2025-07-11
chore(styling): last styling of top menu

Rumyra:
### Description

This pr

Desktop
- Brings back borders where appropriate as per Anuja's design
- Modifies the Learn drop down so html, css & js are one level
- Modifies About - adds github and 'M' icon, reorders and adds space

Mobile
- Modifies the border underneath the first a/p in sections to be more subtle
- Adds a little space to About due to desktop changes

This pr *does not*
- Move search

This pr also includes (as it was a minor change)
- Blog area highlight is now blue


components/menu/mobile.css
```diff
@@ -50,6 +50,26 @@
...
+
+      display: block;
+
+      height: 1px;
+
+      content: "";
+
+      background-color: var(--color-border-primary);
+    }
+  }
```
caugner, line 65: Why does `border-bottom` not work here?

pepelsbey: image

It would be full-width

PR #372, opened 2025-07-11
chore(outer-layout): add robots meta

caugner:
### Description

Adds a `robots` meta tag to the outer layout.

### Motivation

Avoids that search engines index pages we don't want them to index (like the MDN Kitchensink).

### DIFF
--- a/components/outer-layout/server.js
+++ b/components/outer-layout/server.js
@@ -1,4 +1,5 @@
 import { html } from "@lit-labs/ssr";
+import { nothing } from "lit";
 import { ifDefined } from "lit/directives/if-defined.js";
 import { unsafeHTML } from "lit/directives/unsafe-html.js";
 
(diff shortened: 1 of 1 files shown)


components/outer-layout/server.js
```diff
@@ -110,6 +111,21 @@ export class OuterLayout extends ServerComponent {
    * @param {import("@fred").Context} context
    */
   _renderMeta(context) {
+    const ALWAYS_ALLOW_ROBOTS = false; // FIXME
```
caugner: @LeoMcA How should I pass this here from the environment?

See: https://github.com/search?q=repo%3Amdn%2Fyari+%2FALWAYS_ALLOW_ROBOTS%2F+path%3A%2F%5E%5C.github%5C%2Fworkflows%5C%2F%2F&type=code

LeoMcA: Make it something like `FRED_ALLOW_ROBOTS` (prefixed with `FRED_`) then add it to the env component: https://github.com/mdn/fred/blob/e2aceed83a39d2716fbe375d9258a2ab38468678/components/env/index.js

caugner: Gotcha, done in 918f315.

PR #373, opened 2025-07-11
fix(interactive-examples): dark theme now applies on load

LeoMcA:
https://github.com/mdn/fred/commit/dbf60d16afe109bba069c7f61d4ff2eff9513d39 fixed things for the playground but not interactive examples


caugner (approved):
Tested locally on http://localhost:3000/en-US/docs/WebAssembly/Reference/Control_flow/br#try_it, LGTM.

PR #374, opened 2025-07-11
enhance(font): remove code font layout features

pepelsbey:
### Description

- Remove layout features from the source file
- Rebuild the subset copies

⚠️ Still testing the rendering and might also supply a layout-features-stripping script


pepelsbey:
Fixed in https://github.com/mdn/fred/pull/397, closing

PR #380, opened 2025-07-14
chore(styling): prev/next, blue blog & spacing

Rumyra:
### Description

This pr modifies a few styling things:

1. Prev/next links are now styled (this requires a changlog @mirunacurtean )
2. Blog highlight colours are now blue to match top menu
3. Modifies _some_ spacing in content pages for flow

### DIFF
--- a/components/content-section/server.css
+++ b/components/content-section/server.css
@@ -12,6 +12,7 @@
   h4,
   h5,
   h6 {
+    margin: 1rem 0 0;
     font-weight: normal;
   }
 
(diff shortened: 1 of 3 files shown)


components/content-section/server.css
```diff
@@ -201,4 +202,9 @@
       border-left: 1px solid var(--color-border-primary);
     }
   }
+
+  /* Article footer */
+  &.article-footer {
```
pepelsbey, line 207:
```suggestion
  .article-footer {
```

Is it `.content-section.article-footer` or `.content-section .article-footer` (nested child)? I believe it’s the second case and the ampersand is not needed.

Rumyra: It's not it's the first case and this tripped me up for a bit :)

PR #385, opened 2025-07-14
fix(code-example): highlight good and bad examples

LeoMcA:
Examples of good and bad example blocks: http://localhost:3000/en-US/docs/MDN/Writing_guidelines/Howto/Markdown_in_MDN#links

Example of plain `pre` styling which I had to adjust a bit: http://localhost:3000/en-US/docs/Glossary/Base64#base64_characters

### DIFF
--- a/components/code-example/common.css
+++ b/components/code-example/common.css
@@ -23,22 +23,62 @@
   }
 
   pre {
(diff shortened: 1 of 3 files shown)


argl (approved):
LGTM

PR #387, opened 2025-07-14
fix(breadcrumbs-bar): show active value in dropdowns

LeoMcA:
image

### DIFF
--- a/components/color-theme/element.css
+++ b/components/color-theme/element.css
@@ -56,6 +56,12 @@
   &[data-mode="dark"]::before {
     mask-image: var(--color-theme-dark);
   }
+
+  &[data-mode="light dark"] ~ * [data-mode="light dark"],
+  &[data-mode="light"] ~ * [data-mode="light"],
+  &[data-mode="dark"] ~ * [data-mode="dark"] {
+    background: var(--color-background-secondary);
+  }
 }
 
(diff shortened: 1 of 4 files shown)


argl:
Nice, it should extrend to the language chooser as well for concistency.

components/color-theme/element.css
```diff
@@ -56,6 +56,12 @@
   &[data-mode="dark"]::before {
     mask-image: var(--color-theme-dark);
   }
+
+  &[data-mode="light dark"] ~ * [data-mode="light dark"],
+  &[data-mode="light"] ~ * [data-mode="light"],
+  &[data-mode="dark"] ~ * [data-mode="dark"] {
+    background: var(--color-background-secondary);
+  }
```
caugner: I think we should do this programmatically via a CSS class instead.

caugner (commented):
Nits:
- Move padding from container to items.
- Add gap between items (so that current item and hovered item don't merge).

LeoMcA:
> Nits:
> 
> - Move padding from container to items.
> - Add gap between items (so that current item and hovered item don't merge).

Agreed: but we should tackle this as a lower priority issue. Possibly also merging some styles from the two drop-downs for consistency, and easier implementation of future ones.

PR #388, opened 2025-07-14
fix(color): a number of text color variables weren't updated when they were renamed

LeoMcA:
A number of text colour changes (including links in the curriculum), all look good to me. Would be good to have a second set of eyes go through them, though.


components/baseline-indicator/server.css
argl, line 1: I'd remove these link-color overrides, these links should be shown with our global link colors.

components/blog-post/server.css
```diff
@@ -221,7 +221,7 @@
   h2:first-of-type {
     margin: 0;
 
-    /* color: var(--text-link); */
+    /* color: var(--color-link-normal); */
```
argl, line 224: remove this?

components/curriculum-landing/server.css
```diff
@@ -302,7 +302,7 @@
 
               font-weight: var(--font-body-strong-weight);
```
argl, line 303: related: `--font-body-strong-weight` also does not exist any more.

argl:
I'd close this one in favour of #396.

PR #392, opened 2025-07-15
fix(article-footer): remove link from header

pepelsbey:
### Description

- Removes the anchor link from the article footer’s heading
- Uses user-readable `#feedback` anchor

### Before/after

before-after


Rumyra (approved):

Rumyra:
lgtm!

PR #393, opened 2025-07-15
fix(css-formal-syntax): remove js and add consistent styling

LeoMcA:
- don't artificially add `code-example` class with js which was the
  source of this style breakage and adds a js dependency to proper
  rendering
- change font-size to same size as other pre tags
- add syntax highlighting, as exists on yari

## Before

Screenshot 2025-07-15 at 12-13-40 color-scheme

## After

Screenshot 2025-07-15 at 12-13-44 color-scheme

## yari

Screenshot 2025-07-15 at 12-13-50 color-scheme - CSS MDN


argl (approved):
LGTM

PR #397, opened 2025-07-15
fix(font): disable code ligatures

pepelsbey:
### Description

Disables code ligatures via CSS while I’m [figuring out](https://github.com/mdn/fred/pull/374) how to remove them from the font.

It would’ve been enough to do this in `@font-face` but Chrome doesn’t support `font-feature-settings` descriptor.

### DIFF
--- a/components/play-editor/element.css
+++ b/components/play-editor/element.css
@@ -32,6 +32,7 @@
       /* codemirror uses `font-family: monospace;` in multiple places
       which causes weird fonts to be used in various locales */
       font-family: var(--font-family-code) !important;
+      font-feature-settings: "calt" 0;
     }
   }
 }


LeoMcA (approved):
Looks like you missed adding `font-feature-settings` in the play-editor: https://github.com/mdn/fred/blob/8d239e78d8a8eb9d2ecdbb46bc65fc34057f53aa/components/play-editor/element.css#L34

Otherwise looks good, shame we have to add `font-feature-settings` everywhere.

An idea to avoid that: since we already set `font-family: var(--font-family-text);` globally on `html`, could we perhaps set `font-family: var(--font-family-code); font-feature-settings: "calt" 0;` globally on `pre` and `code` in the `global/global.css` file? Browser default styles are already to set them to a monospaced font, we might as well make it ours?

pepelsbey:
> could we perhaps set `font-family: var(--font-family-code); font-feature-settings: "calt" 0;` globally on `pre` and `code` in the `global/global.css` file?

I’m trying to avoid growing global styles as much as possible. Keeping component styles local to the component is much easier to understand and maintain. We already have a simple variable you can use `--font-family-code` when you need it.

PR #399, opened 2025-07-15
fix(font): link colors, font variables in curriculum, observatory

argl:
### Description

Fixes the link colors in curriculum

### Motivation

Yari parity and conforming to fred's typesetting rules

### DIFF
--- a/components/curriculum/shared.css
+++ b/components/curriculum/shared.css
@@ -61,37 +61,6 @@
       a:active {
         background-color: transparent;
       }
-
-      a[href^="#"]:not(.button) {
-        text-decoration: none;
-
-        &::before {
-          display: inline-block;
-          visibility: hidden;
-
-          width: 0.8em;
-
-          margin-left: -0.8em;
-
-          font-size: 0.7em;
-
-          line-height: 1;
-
-          color: var(--color-text-secondary);
-
-          text-decoration: none;
-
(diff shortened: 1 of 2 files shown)


components/curriculum/toc.css
```diff
@@ -45,11 +44,11 @@
         border-left: 2px solid var(--color-border-primary);
 
         &:hover {
-          color: var(--text-link);
+          /* color: var(--text-link); */
         }
```
LeoMcA: Remove instead of comment out?

components/curriculum-landing/server.css
```diff
@@ -299,10 +299,9 @@
 
             a {
               margin-left: 4rem;
+              font-weight: var(--font-weight-bold);
 
-              font-weight: var(--font-body-strong-weight);
-
-              color: var(--text-primary);
+              /* color: var(--text-primary); */
```
LeoMcA: Remove instead of comment?

LeoMcA (approved):
Looks good: there's a few more things changed here than just the curriculum link colours, so I wonder if the PR name (and commit message) should be something more like "fix(styling): introduce font-weight variables, and make link colors more consistent"

There's also some of the changes from https://github.com/mdn/fred/pull/388 missing, but I guess you're planning on merging that separately?

PR #400, opened 2025-07-15
fix(syntax-highlighting): ensure comment colour is accessible

LeoMcA:
Also not accessible on yari, so this is an improvement here.

Before:

Screenshot 2025-07-15 at 18-55-23 Control flow and error handling

After:

Screenshot 2025-07-15 at 18-51-30 Control flow and error handling

Contrast check for light and dark mode:

Screenshot 2025-07-15 185243
Screenshot 2025-07-15 185258

### DIFF
--- a/components/code-example/prism.css
+++ b/components/code-example/prism.css
@@ -4,11 +4,11 @@
   .token.prolog,
   .token.doctype,
   .token.cdata {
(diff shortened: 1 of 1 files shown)


pepelsbey (commented):
I don’t think we would ever use a code theme built around our color variables. So I would rather use a HEX values here instead of a variable, even if they come directly from the variable.

PR #402, opened 2025-07-15
fix: ensure content can't overflow the page

LeoMcA:
`overflow-wrap` inherits everywhere
make titles overflow nicely with use of `<wbr>`
some boutique approaches for interactive examples, compat table and footer

it'll be nice to apply the `<wbr>` algorithm to all code blocks, but that's a rari change for later

| before                                                                                        | after                                                                                         |
| --------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------- |
|  |  |
|  |  |
|  |  |


pepelsbey:
There’s a simpler solution that doesn’t require `<wbr>`: set [`overflow-wrap: anywhere`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#anywhere) on html, [somewhere here](https://github.com/mdn/fred/blob/main/components/html/global.css). It’s inherited down the tree and Baseline green, too.

Screen Shot 2025-07-16 at 14 32 01

LeoMcA:
@pepelsbey I've set `overflow-wrap: break-word;` which inherits yes, the point of the `<wbr>`s is to ensure these long strings break in *nice* places, not just anywhere.

pepelsbey:
> @pepelsbey I've set `overflow-wrap: break-word`

I’m specifically talking about the [`anywhere`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#anywhere) value for the `overflow-wrap` property. It behaves differently and is safer to be inherited:

> Soft wrap opportunities introduced by the word break **are considered** when calculating min-content intrinsic sizes.

I would suggest using `anywhere` value on the html root element **and** not using `break-word` values anywhere else. And maybe avoid `<wbr>` altogether…

Here’s how VoiceOver reads these h1s:

```html
<h1>CanvasRenderingContext2D.globalCompositeOperation</h1>

<h1>Canvas<wbr>Rendering<wbr>Context<wbr>2D<wbr>.global<wbr>Composite<wbr>Operation</h1>
```

wbr

If you add aria-label with the value, it becomes exactly like in the first case but with “7 items” at the end:

```html
<h1 aria-label="CanvasRenderingContext2D.globalCompositeOperation">
  Canvas<wbr>Rendering<wbr>Context<wbr>2D<wbr>.global<wbr>Composite<wbr>Operation
</h1>
```

LeoMcA:
> I’m specifically talking about the [anywhere](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#anywhere) value for the overflow-wrap property. It behaves differently and is safer to be inherited:

> > Soft wrap opportunities introduced by the word break are considered when calculating min-content intrinsic sizes.

@pepelsbey no, that makes `anywhere` *more* dangerous: if wrapping opportunities are considered when calculating `min-content`, then anywhere that uses `min-content` is gonna turn into a very long vertical column of letters: see [this example](https://developer.mozilla.org/en-US/play?id=wng0CmhRPZ8OMlWqoE1JXD6G5j2KOcFq7G%2FSEX%2FyRt%2BzJOiAwWVUEKvSyWe8Xd3or7weAw18N4z3O1d%2B):

```css
div {
  overflow-wrap: anywhere;
  background: red;
  width: min-content;
}

div:last-of-type {
  overflow-wrap: break-word;
}
```

image

> Here’s how VoiceOver reads these h1s:

Weird, [there's nothing clickable I can see about `<wbr>`s](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-wbr-element), and it's been around since Firefox 1. Does putting an `aria-hidden` on them fix things? If not, we can use the unicode equivalent character (though that has the unfortunate effect of appearing in your clipboard if you copy the string).

PR #408, opened 2025-07-16
enhance(reference-layout): improve h1 line-height

pepelsbey:
### Description

Before/after

h1


LeoMcA (approved):
lgtm

Rumyra:
Thanks @pepelsbey I really wanted to do this but not without a discussion and it wasn't critical - but as its done now let's merge it!

PR #409, opened 2025-07-16
fix(about): auto-scroll to top of tab panel when tab is switched

argl:
### Description

If scrolled down on a selected tab on the about page, selkecting a different tab on the sticky tab bar, the newly selected tab now automatically scrolls to the top.

### Motivation

Yari parity


LeoMcA (approved):
nice 👍

PR #414, opened 2025-07-17
chore(content-section): style blockquote

caugner:
### Description

Adds border, and sets color and spacing for `<blockquote>`s within content.

### Motivation

Parity with yari.

### DIFF
--- a/components/content-section/server.css
+++ b/components/content-section/server.css
@@ -70,7 +70,7 @@
   blockquote {
     margin-inline: 0;
     color: var(--color-text-secondary);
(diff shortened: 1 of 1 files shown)


components/content-section/server.css
```diff
@@ -67,6 +67,16 @@
     }
   }
 
+  blockquote {
+    margin-inline: 0;
+    color: var(--color-text-secondary);
+    border-inline-start: 4px solid var(--color-border-primary);
```
LeoMcA:
```suggestion
    border-inline-start: 0.25rem solid var(--color-border-primary);
```

PR #415, opened 2025-07-17
fix(content-section): add scroll-margin-top on :target

caugner:
### Description

Adds `scroll-margin-top` to `:target` elements within content.

### Motivation

Fixes an issue where clicking e.g. [#autofocus](https://fred.review.mdn.allizom.net/en-US/docs/Web/HTML/Reference/Elements/button#autofocus) scrolls the clicked item behind the page header.

### DIFF
--- a/components/content-section/server.css
+++ b/components/content-section/server.css
@@ -6,6 +6,10 @@
 .content-section {
   line-height: var(--font-line-content);
 
+  :target {
+    scroll-margin-top: var(--sticky-header-height);
+  }
+
   h1,
   h2,
   h3,


components/content-section/server.css
```diff
@@ -6,6 +6,10 @@
 .content-section {
   line-height: var(--font-line-content);
 
+  :target {
+    scroll-margin-top: var(--sticky-header-height);
```
LeoMcA, line 10: Is there any downside to applying `scroll-margin-top` on all children?

caugner: I was thinking of adding this rule to the global style, actually.

I don't think we should add it to elements where it has no effect, although I have already seen `[id] { … }`.

LeoMcA: Does it matter if it's on our universal selector, though - if there's no downside?

Seems like we can't use `:target` in shadow DOMs, so that would be an argument against doing `:target { scroll-margin-top...` in our `global/global.css`:

> Note: Due to [a possible bug in the CSS specification](https://discourse.wicg.io/t/target-css-does-not-work-because-shadowroot-does-not-set-a-target-element/2070/), :target doesn't work within a [web component](https://developer.mozilla.org/en-US/docs/Web/API/Web_components) because the [shadow root](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot) doesn't pass the target element down to the shadow tree.

https://developer.mozilla.org/en-US/docs/Web/CSS/:target#description

The code as written fixes the bug, I vote for merging then looping back round if we need to.

PR #416, opened 2025-07-17
enhance(search-mobile): restyle on mobile and add close button

LeoMcA:
introduce `@custom-media` queries for the full-width and hamburger menu
and use them across components rather than hard-coded values

### DIFF
--- a/components/search-modal/element.css
+++ b/components/search-modal/element.css
@@ -6,47 +6,74 @@
 
 dialog {
   padding: 0;
+  border: 0;
 
-  background-color: var(--color-background-primary);
-  border: 1px solid var(--color-border-primary);
-  border-radius: 0.5rem;
-
-  /* Desktop menu */
-  @media (width > 1006px) {
+  @media (--screen-menu-full) {
     width: calc(var(--layout-content-max) + 1rem);
     max-height: calc(100% - var(--sticky-header-height) - 2rem);
+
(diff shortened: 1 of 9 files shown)


components/search-modal/element.js
```diff
@@ -198,6 +204,14 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
             aria-label=${this.l10n`Search`}
           />
         </form>
+        <mdn-button
+          class="close"
+          variant="plain"
+          icon-only
+          .icon=${exitIcon}
+          @click=${this._close}
+          >Exit search</mdn-button
```
caugner:
```suggestion
          >${this.l10n`Exit search`}</mdn-button
```

components/search-modal/element.css
```diff
@@ -6,55 +6,91 @@
...
+    margin-inline: 0.3rem;
   }
 }
 
-progress {
-  width: calc(100% - 2rem);
-  margin: 0 auto 0.5rem;
+.close {
+  grid-area: close;
+  font-size: 1.37rem;
```
caugner, line 66: Nit: This font size seems to be bit odd.

caugner (approved):
LGTM, just one nit.

components/search-modal/element.css
LeoMcA: It's so the icon appears the same as the hamburger menu one: I'm not sure why that isn't using a `mdn-button`, but hey-ho - we can fix it later

PR #420, opened 2025-07-17
fix(curriculum): increase card height to accommodate for larger font size

argl:
### Description

Increase card size to avoid scrolling

Before
image

After
image

### DIFF
--- a/components/curriculum/module-list.css
+++ b/components/curriculum/module-list.css
@@ -177,6 +177,7 @@
           width: 100%;
           min-width: 15rem;
           max-width: 20rem;
+          height: 100%;
 
           padding: 0;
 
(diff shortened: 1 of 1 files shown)


components/curriculum/module-list.css
```diff
@@ -245,7 +245,7 @@
             align-items: center;
             justify-content: space-between;
 
-            height: 11rem;
+            height: 13rem;
```
caugner: Isn't there be a way in flexbox to have all items take the size of the highest element? 🤔

argl: Flexbox instances in a grid layout? This is basically the original styling, this is needed because the font sizes have changed slightly.

caugner: Oh, because these are nested, it would only work with subgrid maybe.

caugner (approved):

PR #424, opened 2025-07-17
fix(header): layout shift due to login link

LeoMcA:
use simple icon and set width/height outside of web component

| before | after |
| - | - |
| Screen Shot 2025-07-17 at 18 38 14 | Screen Shot 2025-07-17 at 18 47 43 |

before:

Screen Shot 2025-07-17 at 18 37 58

after:

Screen Shot 2025-07-17 at 18 38 05


pepelsbey (approved):
LGTM!

PR #427, opened 2025-07-18
fix(content-section): make anchors text color

pepelsbey:
### Description

Screenshot 2025-07-18 at 10 59 59

Anchors in dls:

- Text color
- No underline until hovered
- `#` next to it once hovered


LeoMcA (approved):
lgtm!

PR #436, opened 2025-07-18
fix(search): close when clicking backdrop cross-browser

LeoMcA:
the `closedby` attribute on `<dialog>` allows us to close these (including the search) by clicking on the backdrop, but this doesn't work in Firefox Release (only Firefox Nightly). https://github.com/tak-dcxi/dialog-closedby-polyfill fixes this, but it's a bit heavy - and really covers all edge cases, which we don't use - so I cut it down a lot, and included only what we need

Tested in Firefox Release/Nightly and Chrome: would be great to get a test in Safari too

### DIFF
--- a/components/modal/element.js
+++ b/components/modal/element.js
@@ -27,7 +27,7 @@ export class MDNModal extends LitElement {
 
   render() {
     return html`
-      <dialog>
+      <dialog closedby="any">
         <header>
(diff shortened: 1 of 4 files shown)


LeoMcA:
@pepelsbey latest commit should fix that (tested in Epiphany)

pepelsbey (approved):
Looks good!

PR #437, opened 2025-07-18
enhance(font): remove ligatures from JB Mono

pepelsbey:
### Description

Replaces JB Mono source files with the new ones compiled without ligatures from the source.

Saves some weight for WOFF files as well.


argl (approved):

pepelsbey:
@LeoMcA let me know if [the instructions](https://github.com/mdn/fred/pull/437/commits/d5073e02eb555e969031dbc4fdce8298e32514e1) are clear enough to reproduce the non-ligature font.

LeoMcA:
@pepelsbey would be great to specify either the commit that patch currently works against, or the "method" taken (is it just deleting everything out of the "calt" block), or both - will help in the future if we have to apply this again against and updated font

LeoMcA (approved):
Nice job: a little more clarity about the approach taken to remove ligatures and/or the commit hash that patch applies cleanly against would be great, as commented above

Also a couple more `font-feature-settings` which can go:
https://github.com/mdn/fred/blob/0aeb772092fd82282b6c72f94489e7f365624560/components/content-section/server.css#L85
https://github.com/mdn/fred/blob/0aeb772092fd82282b6c72f94489e7f365624560/components/code-example/common.css#L43

PR #452, opened 2025-07-24
feat(l10n): extract strings, add docs, load locales on demand

LeoMcA:
Gets things to a stage where all strings are - in theory - l10n-able. Next steps will be improving the l10n experience, and further optimising which strings we ship to the client.

The added README is a good source for some of what's going on here.


caugner:
Converted to draft, as it has merge conflicts.

l10n/template.ftl
```diff
@@ -0,0 +1,288 @@
...
+not-found-description = Sorry, the page <code data-l10n-name="url">{ $url }</code> could not be found.
+not-found-fallback-english = <strong data-l10n-name="strong">Good news:</strong> The page you requested exists in <em data-l10n-name="em">English</em>.
+not-found-fallback-search = The page you requested doesn't exist, but you could try a site search for:
+not-found-back = Go back to the home page
+reference-toc-header = In this article
+footer-mofo = Visit <a data-l10n-name="moco">Mozilla Corporation’s</a> not-for-profit parent, the <a data-l10n-name="mofo">Mozilla Foundation</a>.
+footer-copyright = Portions of this content are ©1998–{ $year } by individual mozilla.org contributors. Content available under <a data-l10n-name="cc">a Creative Commons license</a>.
+search-modal-site-search = Site search for <em>{ $query }</em>
+site-search-search-stats = Found { $results } documents.
+site-search-suggestion-matches =
```
bcolsson, line 24: While this is technically valid fluent it's best to avoid this pattern. This should be two strings - otherwise you can't guarantee that all patterns will have translations ([reference](https://github.com/projectfluent/fluent/wiki/Good-Practices-for-Developers#prefer-separate-messages-over-variants-for-ui-logic)). 

Typically you want to remove the logic from the strings themselves, so it's preferable to have a `-greaterthan` and a `-equals` string then have the logic happening outside of fluent to choose the correct string.

l10n/template.ftl
```diff
@@ -0,0 +1,288 @@
...
+       *[preference] preference
+        [runtime_flag] runtime flag
+    }{ NUMBER($has_value) ->
+        [one] { " " }to <code data-l10n-name="value">{ $flag_value }</code>
+       *[other] { "" }
+    }{ "." }
+    { $flag_type ->
+        [preference] To change preferences in { $browser_name }, visit { $browser_pref_url }.
+       *[other] { "" }
+    }
```
bcolsson, line 88: This is untranslatable. Even in English I barely understand what a final string would look like, so compounding translations with different word order, you'll have a very difficult time to get something comprehensible out of this.

This should be split into multiple strings. E.g.
`string-a = From version { version }`
`string-b = From version { version } until {version} users`
ETc.

l10n/template.ftl
```diff
@@ -0,0 +1,288 @@
...
+compat-link-report-issue-title = Report an issue with this compatibility data
+compat-link-report-missing-title = Report missing compatibility data
+compat-link-report-missing = Report this issue
+compat-link-source = View data on GitHub
+compat-link-source-title = File: { $filename }
+compat-deprecated = Deprecated
+compat-experimental = Experimental
+compat-nonstandard = Non-standard
+compat-no = No
+compat-support-full = Full support
```
bcolsson: I'm finding this and multiple other strings with the exact same ID. You'll need to find a way to avoid duplication.

l10n/template.ftl
```diff
@@ -0,0 +1,288 @@
...
+       *[other] { "" }
+    }
+compat-legend = Legend
+compat-legend-tip = Tip: you can click/tap on a cell for more information.
+compat-legend-yes = { compat-support-full }
+compat-legend-partial = { compat-support-partial }
+compat-legend-preview = In development. Supported in a pre-release version.
+compat-legend-no = { compat-support-no }
+compat-legend-unknown = Compatibility unknown
+compat-legend-experimental = { compat-experimental }. Expect behavior to change in the future.
```
bcolsson, line 94: I sort of understand why you want to use a reference here and below, but there is the potential for localization issues here. It'd be simpler for localizers to just have the word instead of a message reference.

bcolsson (commented):
I understand that more work is happening around localization at some point in the future, but I noticed some major issues that could cause headaches down the road, so adding comments. I'd suggest making changes before asking localizers to translate these. (If the expectation is to use Pontoon, then the complicated Fluent logic wouldn't be supported well either.)

LeoMcA:
@bcolsson thanks for your comments, once we've got the mechanics of our string scraping sorted out in this PR, we'll open others to resolve what you've noted, and make some of the auto-generated ids more meaningful.

@caugner I've updated this PR in line with what we discussed a little while ago in Berlin: removing the ability to add strings without defining an id. I temporarily used some of that logic in the migration script to update the current id-less instances to ones with ids.

Commits are mostly atomic - they may not all stand alone, because for some of them I simply split up my previous large commit more logically - but should be much nicer to review than before.

package.json
```diff
@@ -20,7 +20,9 @@
     "tsc": "tsc",
     "rari": "rari",
     "prepack": "npm run build",
-    "wdio": "wdio run ./wdio.conf.js"
+    "wdio": "wdio run ./wdio.conf.js",
+    "l10n:extract": "node l10n/parser/extract.js",
+    "l10n:lint": "node l10n/parser/extract.js --lint"
```
caugner: How about:

```suggestion
    "l10n": "node l10n/parser/extract.js",
    "l10n:lint": "node l10n/parser/extract.js --lint"
```

Rather than `extract.js` maybe it could be called `l10n/cli/index.js`?

server.js
```diff
@@ -224,6 +224,12 @@ export async function startServer() {
       },
       selfHandleResponse: true,
       on: {
+        proxyReq: async (req) => {
+          const locale = req.path.split("/")[1];
+          if (locale && /^q[a-t][a-z]$/.test(locale)) {
+            req.path = req.path.replace(locale, "en-US");
+          }
```
caugner, line 234: I don't understand what this does, can you explain in a code comment?

l10n/template.ftl
caugner, line 1: Is this intentionally not alphabetically ordered?

symmetric-context/both.js
```diff
@@ -1,7 +1,7 @@
 /**
  * Runs on either client or server,
  * and returns the client or server context respectively
- * @returns {import("./types.js").SymmetricContext}
+ * @returns {import("./types.js").SymmetricContext | undefined}
```
caugner: Should the JSDoc comment explain when this is `undefined`?

l10n/parser/transform.js
```diff
@@ -0,0 +1,78 @@
+import { readFile, writeFile } from "node:fs/promises";
+import { fileURLToPath } from "node:url";
+
+import { Transformer, parse, serialize } from "@fluent/syntax";
+
+/**
+ * @import { TextElement } from "@fluent/syntax";
+ */
+
+class AccentTransformer extends Transformer {
```
caugner, line 16: Can you add a JSDoc comment explaining what this transformer does?

l10n/fluent.js
```diff
@@ -210,7 +192,23 @@ function getLocale(locale) {
 }
 
 /**
- * @param {string} [locale]
+ * @param {string} locale
+ */
+export async function loadFluentFile(locale) {
+  if (locale !== "qa" && !ftlMap[locale]) {
```
caugner: [nit] Do we explain somewhere what the "qa" locale is?

l10n/mixin.js
```diff
@@ -1,11 +1,14 @@
...
 
-import getFluentContext from "./fluent.js";
+import getFluentContext, { loadFluentFile } from "./fluent.js";
 
 /**
  * @import { LitElement } from "lit";
  */
 
+const locale = getSymmetricContext()?.locale;
+if (locale) await loadFluentFile(locale);
```
caugner: Should we be calling this at top level?

l10n/mixin.js
```diff
@@ -17,7 +20,13 @@ export const L10nMixin = (Base) =>
      */
     constructor(...args) {
       super(...args);
-      const context = getSymmetricContext();
+      let context = getSymmetricContext();
+      if (!context) {
+        console.error("SymmetricContext is undefined, reverting to defaults");
+        context = {
+          locale: "en-US",
+        };
```
caugner: When would this happen? (Add comment?)

l10n/fluent.js
```diff
@@ -219,7 +219,11 @@ export default function getFluentContext(locale) {
   function l10n(id, _comment) {
     // called as a function, returning a template tag:
     // l10n("foobar")`Foobar`
-    const localizedString = getLocale(locale)?.get(id);
+    const localizedStringOrHtml = getLocale(locale)?.get(id);
+    const localizedString =
+      typeof localizedStringOrHtml === "string"
```
caugner, line 225: What other type could it have?

l10n/parser/extractor.js
```diff
@@ -0,0 +1,107 @@
...
+          }
+        }
+      }
+    }
+  }
+
+  fluentResource.body = [
+    new Comment(
+      `WARNING: do not edit this file, it's automatically generated by ExtractL10nPlugin.
+If you need to manually add strings, do so in ./locales/en-US.ftl. See ./README.md for more details.`,
```
caugner: Could we make this a constant at the top of the file?

l10n/parser/extractor.js
```diff
@@ -0,0 +1,107 @@
...
+          const [arg] = tagNode.getArguments();
+          if (Node.isStringLiteral(arg)) {
+            const key = arg.getLiteralValue();
+            const value = getLiteralValue(taggedTemplate);
+            map.set(key, value);
+          }
+        }
+      }
+    }
+  }
```
caugner: Could we split this up a bit, to make some parts of it unit-testable?

l10n/locales/.gitignore
caugner, line 1: Missing newline at EOF.

l10n/fluent.js
```diff
@@ -132,7 +132,7 @@ export class Fluent {
     const parentMessage = bundle ? bundle.getMessage(id) : undefined;
     let message;
 
-    if (this.locale === "qa") {
+    if (this.locale === "qai") {
```
caugner, line 135: What's qa vs qai?

l10n/README.md
```diff
@@ -0,0 +1,61 @@
...
+The other examples are automatically extracted, and combined with the manually specified strings, into `./template.ftl`.
+
+### Pseudolocalization
+
+We have a few pseudo locales for testing if strings have been added, or if components will adapt to localized strings:
+
+- `qaa`: "accented" locale: adds accents to all characters, duplicates some vowels to create longer strings, wraps string in square brackets to help detect truncation
+- `qai`: "id" locale: replaces strings with their identifiers, wrapped in square brackets
+
+The `qai` locale works all the time, the `qaa` locale must be manually generated with `node ./parser/transform.js`
```
caugner: Maybe make `node ./parser/transform.js` a script in `package.json`?

l10n/template.ftl
```diff
@@ -0,0 +1,171 @@
...
+baseline-supported-in = Supported in { $browsers }
+baseline-unsupported-in = Not widely supported in { $browsers }
+baseline-supported-and-unsupported-in = Supported in { $supported }, but not widely supported in { $unsupported }
+homepage-hero-title = Resources for Developers,<br> by Developers
+homepage-hero-description = Documenting <a data-l10n-name="css">CSS</a>, <a data-l10n-name="html">HTML</a>, and <a data-l10n-name="js">JavaScript</a>, since 2005.
+not-found-title = Page not found
+not-found-description = Sorry, the page <code data-l10n-name="url">{ $url }</code> could not be found.
+not-found-fallback-english = <strong data-l10n-name="strong">Good news:</strong> The page you requested exists in <em data-l10n-name="em">English</em>.
+not-found-fallback-search = The page you requested doesn't exist, but you could try a site search for:
+not-found-back = Go back to the home page
```
caugner, line 19: What would be nice is if it detected multiple lines with common prefix and added a line between them, for readability.

l10n/fluent.js
LeoMcA: In l10n/README.md we explain the pseudolocalization locales

l10n/fluent.js
LeoMcA: qa was the old pseudolocale, now we've got multiple I migrated qa to qai

l10n/template.ftl
LeoMcA: It's ordered by manual strings, then extracted strings, but yes we could sort it

LeoMcA: Actually this isn't trivial, as we'd want to move comments with the strings, and if this it to make things useful for localisers manually using these files, preserve blocks (and insert scrape strings into those blocks, etc.) so I'd prefer to do this in a follow up

l10n/fluent.js
LeoMcA: it could be a lit html template, I'll add a comment

l10n/template.ftl
LeoMcA: I'd do this in a follow up, as it's a convenience rather than a requirement to make all strings localisable

l10n/mixin.js
LeoMcA: Yes (because we need to halt further rendering until we've got our fluent files loaded), but doing it here was the reason for my weird SymmetricContext could be undefined workaround (because when we imported files which imported the mixin on the server this got evaluated while it was undefined).

I've undone those changes to SymmetricContext and just put this in a client hook, which is where it belongs.

l10n/cli.js
```diff
@@ -0,0 +1,10 @@
+import { extract } from "./parser/extractor.js";
+import { generateQaaLocale } from "./parser/transform.js";
+
+const lint = process.argv.includes("--lint");
+await extract({ lint });
+
+const generatePseudo = process.argv.includes("--gen-pseudo");
+if (generatePseudo) {
+  await generateQaaLocale();
+}
```
caugner: [nit] Would be nice to add `--help` that lists these to options.

l10n/parser/extractor.js
```diff
@@ -21,20 +22,66 @@ import { Node, Project, SyntaxKind } from "ts-morph";
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url));
 
+const EDIT_WARNING = `WARNING: do not edit this file, it's automatically generated by ExtractL10nPlugin.
```
caugner: [nit] Isn't this essentially the header of the template file?

```suggestion
const TEMPLATE_HEADER = `WARNING: do not edit this file, it's automatically generated by ExtractL10nPlugin.
```

l10n/parser/extractor.js
```diff
@@ -21,20 +22,66 @@ import { Node, Project, SyntaxKind } from "ts-morph";
...
+    }
+  } else {
+    await writeFile(outputPath, output, "utf8");
+  }
+}
+
+/**
+ * @param {string} path Path to fluent file
+ */
+export async function getManualEntries(path) {
```
caugner: [nit] Maybe:

```suggestion
export async function getStaticEntries(path) {
```

test/unit/l10n/fixtures/comments.ftl
```diff
@@ -0,0 +1,6 @@
+# WARNING: this is a multi-
+# line warning comment
+
+# TODO: this is a todo comment
+
+# This is a normal comment
```
caugner: [nit] Missing traling new line

```suggestion
# This is a normal comment

```

caugner (approved):

caugner:
@LeoMcA FYI

> This branch has conflicts that must be resolved

caugner (approved):
LGTM, just one question: When new strings are added inline, are these automatically added to the Fluent file on pre-commit?

LeoMcA:
> When new strings are added inline, are these automatically added to the Fluent file on pre-commit?

I wanted to push a commit and say "they are now!" - but unfortunately it's a little more complicated.

I've added a pre-push hook lint run to catch any errors before CI, but in terms of running the extractor on the pre-commit hook, if you have files that are unstaged but which have strings to be scraped, they'll be pulled into the wrong commit.

We should be able to work around that by passing a list of files to the extractor to act on - but that's something I think makes sense for a follow up PR at this point.

.lefthook.yml
```diff
@@ -45,6 +45,8 @@ pre-push:
             run: npx stylelint '**/*.css' --cache --ignore-path .gitignore
           - name: tsc
             run: npx tsc
+          - name: lint l10n strings
+            run: npm run l10n -- --lint
```
caugner, line 49: Given how fast this runs, should we not also run it on pre-commit?

```
% time npm run l10n -- --lint

> @mdn/fred@2.2.1 l10n
> node l10n/cli.js --lint

npm run l10n -- --lint  0.64s user 0.07s system 113% cpu 0.627 total
```

caugner: The `glob` option supports multiple values, so we could run it only when relevant files were changed.

LeoMcA: The failure case is something like:

- add string to components/a.js
- add string to components/b.js
- stage a.js
- commit
- new strings from both a and b are in template.ftl

PR #454, opened 2025-07-25
feat(placement): new leaderboard format

### LINKED ISSUE #515: Prepare for standard placement sizes
In anticipation for supporting standard-sized placements, we have to increase the top banner spaces to accomodate the leaderborad height of 90 pixels. Native ads should be put in the space at the top, with a grey background.

The leaderboard placement should be centered wiothout overlays. Overlays are shown to the right of the leaderboard content and not obstruct the placement, instead they should be put to the right of the placement if there is space to show them, otherwise these will be hidden.

Top placements will be hidden on mobile screens like the current/yari implementation.

### LINKED ISSUE #516: Show house-ad style content as fallback
To prevent content-shifts, we keep the reserved ad space at the top and display a house-ad style content in place of the ad in places where we do not show ads.

argl:
### Description

Implement new leaderboard and skyscraper formats. These work along the existing natve placements. The cloud function has been amended to support these new formats, [here](https://github.com/mdn/yari/pull/13337) is the corresponding yari PR.

To avoid layout shifts, the top bar space had been increased to accomodate both top bar formats.

#### Wide screen
Screenshot 2025-08-10 at 17 56 55

#### Mid screen
Screenshot 2025-08-10 at 17 57 43

#### Small screen
Screenshot 2025-08-10 at 17 58 20

#### Native top
Screenshot 2025-08-10 at 18 03 20

#### Homepage
Screenshot 2025-08-10 at 18 03 42

### Motivation

Support new formats along with the new ones, leave a constant top space to avoid any layout shifts.

### Additional details

New formats: 728x90, 160x600

fixes #515 
fixes #516


caugner:
Is the "AD" link intentionally not shown in the corner of the ad?

argl:
Yes, there shall be no overlays on the new ad formats.

components/placement-sidebar/element.js
```diff
@@ -65,37 +66,69 @@ export class MDNPlacementSidebar extends PlacementMixin(LitElement) {
...
-          <div class="placement-content">
-            <strong class="placement-heading">${heading}</strong>
-            <span class="placement-copy">${copy}</span>
-            <span class="placement-cta external">${cta}</span>
+    // skyscraper format is not suitable for horizontal layout
+    if (imageFormat === "skyscraper" && this.horizontal) {
+      return nothing;
+    }
+
+    return imageFormat === "skyscraper"
```
LeoMcA, line 74: I'm a bit on the fence about this: doing the conditional on the top level makes the code cleaner, but could lead to us making changes to in format and not the other: I would consider doing the conditionals inline: i.e. `width=${imageFormat === "skyscraper" ? "160px" : "125px"}`, but I'll leave it up to you - you may have attempted this already and decided it was too messy.

components/page-layout/server.js
```diff
@@ -12,10 +12,16 @@ export class PageLayout extends ServerComponent {
...
+      context.renderer,
+    )
+      ? "dark"
+      : "";
+
     return html`
       <body class="page-layout">
         ${A11yMenu.render(context)}
-        <div class="page-layout__banner">
+        <div class="page-layout__banner" style="--color-scheme: ${colorScheme}">
```
LeoMcA: Can we do this the same way we do it in the breadcrumbs and nav?
https://github.com/mdn/fred/blob/8925717983febf01c3d1985c4a2d47b758abb008/components/breadcrumbs-bar/server.js#L19

components/page-layout/server.css
```diff
@@ -20,8 +20,15 @@
   z-index: var(--z-index-sticky-header);
 }
 
-@media (--screen-small-and-narrower) {
-  .page-layout__banner {
+.page-layout__banner {
+  height: calc(5.625rem + 1px);
+
+  color-scheme: var(--color-scheme, light);
```
LeoMcA: https://github.com/mdn/fred/blob/8925717983febf01c3d1985c4a2d47b758abb008/components/breadcrumbs-bar/server.css#L25-L28

components/placement-top/element.css
```diff
@@ -29,13 +29,24 @@
     var(--place-top-cta-color-dark)
   );
 
+  --top-placement-inner-padding: 0.125rem;
+
   min-height: calc(
     var(--top-placement-inner-height) + 2 * var(--top-placement-inner-padding)
   );
```
LeoMcA: These variables are a bit confused:
We set `--top-placement-innner-padding` here, and below in `.top-placement` (to the same value), but don't set `--top-placement-inner-height` on `.top-placement-leaderboard`, but still use it here for calculating the min height.

components/placement-top/element.js
```diff
@@ -36,12 +55,25 @@ export class MDNPlacementTop extends PlacementMixin(LitElement) {
     if (placementContext.status === "noads") {
       return nothing;
     }
+
+    // console.log("placementContext", placementContext);
```
LeoMcA:
```suggestion
```

components/placement-top/element.js
```diff
@@ -75,34 +107,62 @@ export class MDNPlacementTop extends PlacementMixin(LitElement) {
...
-            target="_blank"
-            rel="sponsored noreferrer"
-            ><div class="placement-content">
-              <img
-                src=${this.imgLink(image)}
-                aria-hidden=${!alt}
-                alt=${alt || ""}
-                height="50"
-              /><span>${copy}</span>
+    return imageFormat === "leaderboard"
```
LeoMcA, line 108: Same note as above about doing the conditional on the top level and duplicating vs inline

components/placement-sidebar/element.js
argl: The markup is slightly different, so I decided to separate on this level.

LeoMcA (approved):
lgtm!

PR #455, opened 2025-07-25
fix: randomize button label ids to prevent leaking non-unique element ids

argl:
### Description

The button component potentially leaked element ids, this fixes it.

### Motivation

Proper HTML markup


pepelsbey:
I wonder if we could use [`randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) for that.

argl:
> I wonder if we could use [`randomUUID()`](https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID) for that.

We could, but I prefer the extended character set / shorter strings.

components/button/pure.js
```diff
@@ -26,11 +26,12 @@ export default function Button({
   variant = "primary",
   action,
 }) {
+  const labelId = `label-${randomString()}`;
```
LeoMcA: We need a similar random ID in the dropdown component, let's put a function in `components/utils.js` or similar, and use in both places:

https://github.com/mdn/fred/blob/c7c9b603aae031efadec57dc12fccace29a4b921/components/dropdown/element.js#L71

components/button/pure.js
```diff
@@ -68,3 +69,12 @@ export default function Button({
         </button>
       `;
 }
+
+const charSet =
+  "abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789";
```
LeoMcA: Not an enormous fan of listing out the entire charset here, how's my `Math.random().toString(36).replace("0.", "")` approach from the dropdown component?

argl:
I have made the suggested changes.

LeoMcA (approved):
Nice, ty!

PR #456, opened 2025-07-26
enhance(homepage-search): use design colors

pepelsbey:
### Description

- Uses design color instead of the default browser ones
- Cleans up CSS and normalizes paddings

### Before/after

image

image

### DIFF
--- a/components/homepage-search/element.css
+++ b/components/homepage-search/element.css
@@ -1,36 +1,33 @@
 .mdn-homepage-search {
   display: flex;
 
-  gap: 0.5ch;
-  justify-content: center;
+  gap: 0.25em;
+  align-items: center;
 
-  padding: 1em;
+  padding: 0.8em 1em;
   margin: 0 auto;
 
   font-size: var(--font-size-large);
 
-  border-color: var(--color-border-primary);
-  border-style: solid;
-  border-radius: 2em;
+  cursor: pointer;
+
+  background-color: transparent;
+  border: 2px solid var(--color-border-primary);
+  border-radius: calc(infinity * 1px);
+
+  &:hover {
+    background-color: var(--color-background-secondary);
+  }
 
   &::before {
-    display: inline-block;
-
-    flex-shrink: 0;
-
-    width: 1.1em;
-    height: 1.1em;
+    width: 1em;
+    height: 1em;
 
     content: "";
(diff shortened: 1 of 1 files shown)


components/homepage-search/element.css
```diff
@@ -1,36 +1,33 @@
...
-  border-color: var(--color-border-primary);
-  border-style: solid;
-  border-radius: 2em;
+  cursor: pointer;
 
-  &::before {
-    display: inline-block;
+  background-color: transparent;
+  border: 2px solid var(--color-border-primary);
+  border-radius: calc(infinity * 1px);
```
LeoMcA, line 16: `calc(infinity * 1px)` isn't baseline green (Firefox 114 released 2023-06-06), and looks like a weird hack anyway - I had to pause for a moment to figure out what was going on.

components/homepage-search/element.css
```diff
@@ -1,36 +1,33 @@
 .mdn-homepage-search {
   display: flex;
 
-  gap: 0.5ch;
-  justify-content: center;
+  gap: 0.25em;
```
LeoMcA, line 4: nit: I personally feel like `0.5ch` is more expressive - given it's horizontal spacing, and should be relative to the font's width rather than height

LeoMcA (approved):
Approved with the `infinity` issue addressed:

components/homepage-search/element.css
pepelsbey: It’s not the only place we use infinity to achieve a reliable pill-shaped button, no matter how big it is (hence the infinity), so maybe we should file an issue and merge this one as is?

I would introduce something like `--radius-pill: 9999px` for cases like this, since infinity is not green enough yet.

pepelsbey: There’s a draft PR adding global vars for radius values https://github.com/mdn/fred/pull/472. I will complete it with all `infinity` cases replaced with the `--radius-full` once this PR is merged.

components/homepage-search/element.css
pepelsbey: The em doesn’t really represent the font’s height. It’s just a different and more common (well-understood) unit attached to the font size. Unless there’s an agreement for fred, I would keep it up to the author.

PR #460, opened 2025-07-28
fixes layout issues with prev/next navigation on small screens

argl:
### Description

Fix prev/next navigation on small screens

### Additional details
Before
image
After
image

### DIFF
--- a/components/prev-next/index.css
+++ b/components/prev-next/index.css
@@ -26,9 +26,33 @@
     }
   }
 
+  @media (--screen-xsmall) {
+    display: grid;
+    grid-template-areas:
+      "title title"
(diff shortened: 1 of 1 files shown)


LeoMcA (commented):
Gets a bit tight on smaller screens:
Screen Shot 2025-07-28 at 15 00 14

I'd suggest we reflow into a vertical column past a certain breakpoint. I also notice there's a weird overflowing underline between the icon and the text: can we remove that if it's easy?

argl:
Removed the overflow underline, changed to a better layout on extra small screens:
image

LeoMcA (approved):
Lovely!

PR #461, opened 2025-07-28
fix(left-sidebar): some items displayed horizontally

LeoMcA:
Examples:
- https://test.developer.allizom.org/en-US/docs/Web/HTTP/Guides/Proxy_servers_and_tunneling
- https://test.developer.allizom.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments

Attempted to keep flex, but it's not possible without using not-yet-baseline-green `:has` selector. Reverted to normal block display, which makes the icons less pretty when overflowing - but so be it, the sidebars need some more thorough attention.

### DIFF
--- a/components/left-sidebar/server.css
+++ b/components/left-sidebar/server.css
@@ -63,16 +63,10 @@
         }
       }
     }
-
-    ol {
-      padding-left: 1.5rem;
-    }
   }
 
   li {
-    display: flex;
-    align-items: center;
-    justify-content: stretch;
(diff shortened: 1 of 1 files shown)


argl:
I have a hard time understanding the 'current page' display logic, but I see that this is not a regression.
Why are these selected as  block, but subpages can be navigated individually?

imageimage

The actual issue has been solved though 👍

argl (approved):

PR #462, opened 2025-07-28
fix(playground): editor not minimizing in some browsers

LeoMcA:
`display: none` does the trick! Tested across Firefox (Nightly, Release), Chrome and Epiphany.


argl (approved):
LGTM, fix is working for me & my Safari

PR #472, opened 2025-07-30
feat(radius): add global radius vars

pepelsbey:
### Description

Introduces global vars for border-radius values:

- Normal: 0.25rem (4px) one for all cases (as discussed with Ruth: “we need only one value, a very small one”)
- Full: extra large to make sure that the object becomes circular or pill-shaped
- Addresses the issue with compatibility of the `infinity` method, [mentioned here](https://github.com/mdn/fred/pull/456#discussion_r2236585117)

### DIFF
--- a/components/homepage-search/element.css
+++ b/components/homepage-search/element.css
@@ -13,7 +13,7 @@
 
   background-color: transparent;
   border: 2px solid var(--color-border-primary);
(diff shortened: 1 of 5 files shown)


LeoMcA (approved):
looks good, thanks

PR #474, opened 2025-07-30
fix(ai-help): nav and sidebar overlap on mobile

LeoMcA:
caused by conflicts between variables in our legacy css, and in fred's css: so rename `--top-nav-height` to `--navigation-height` - which better fits the name of the component, anyway


argl (approved):
LGTM

PR #494, opened 2025-07-31
chore: add active user ping

### LINKED ISSUE #451: Port yari settings metrics ping
Sent by yari here: https://github.com/mdn/yari/blob/main/client/src/hooks.ts#L185
Consumed by rumba here: https://github.com/mdn/rumba/blob/main/src/api/ping.rs

The legacy code should still be sending the ping, but that'll only happen on legacy pages: we should send this for any active user on any page. So we also need to keep the localstorage key named the same, to ensure we don't send duplicate pings.

LeoMcA:
Fixes: https://github.com/mdn/fred/issues/451

Ports code from yari, without offline component.

Tested locally that:
- When logged out, no ping is sent
- When logged in, ping is sent once, but not on subsequent loads
- When deleting localstorage item, ping is sent once again

### DIFF
--- a/entry.client.js
+++ b/entry.client.js
@@ -11,3 +11,4 @@ import "./hooks/toc-highlight.js";
 import "./hooks/code-examples.js";
 import "./hooks/live-samples.js";
 import "./hooks/skip-to-search.js";
+import "./hooks/user-ping.js";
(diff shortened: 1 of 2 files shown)


argl (approved):
LGTM

Issue #465, opened 2025-07-29
Localised 404 page


LeoMcA:
We build and serve an en-US 404 page for all locales, which leads to some oddities:

- The links in the top menu are all to `/en-US`
- The page shows a locale switcher, with English as the only option
- This also stops us using localised strings, "server side", in this page: once that's fully implemented

For now, we could hide the locale switcher on 404 pages, but ideally, if the page exists in other locales, it should show those locales.

We should also render a 404 for each locale, and use the locale specific 404 in our cloud function. Alternatively, we could render the 404 "live" in our cloud function, but this could be more expensive, and should probably be part of a wider effort on that.

This is a particularly bad experience if clicking around the top navigation on a locale, since you'll:

1. Start on a locale
2. Click on a link in the top nav which doesn't include that locale
3. Get a 404, in English
4. Click on another link in the top nav, to carry on
5. End up on a documentation page, in English

mirunacurtean:
Tested this with https://test.developer.allizom.org/en-US/ and here are some of the results

**Localization testing:** 
**Homepage** - all Theme switcher locales 

**Top Menu navigation** 

DE - Check  articles are translated when possible, no 404s, Playground, Curriculum, Observatory, About links redirect to en-US (as expected)

ES 
Top Menu navigation
https://test.developer.allizom.org/es/docs/Web/HTML/Guides/Date_and_time_formats - 404 
https://test.developer.allizom.org/es/docs/Web/CSS/Properties
https://test.developer.allizom.org/es/docs/Web/CSS/CSS_Values_and_Units 
https://test.developer.allizom.org/es/docs/Web/CSS/Guides 
https://test.developer.allizom.org/es/docs/Web/CSS/Layout_cookbook/Column_layouts 
https://test.developer.allizom.org/es/docs/Web/CSS/Layout_cookbook/Center_an_element 
https://test.developer.allizom.org/es/docs/Web/CSS/Layout_cookbook/Card 
https://test.developer.allizom.org/es/docs/Web/JavaScript/Guide/Using_classes 
https://test.developer.allizom.org/es/docs/Web/API/File_System_API 
https://test.developer.allizom.org/es/docs/Web/API/HTML_DOM_API
https://test.developer.allizom.org/es/docs/Web/API/Web_Animations_API/Using_the_Web_Animations_API
https://test.developer.allizom.org/es/docs/Web/API/History_API/Working_with_the_History_API 
https://test.developer.allizom.org/es/docs/Web/URI
[...]

PR #512, opened 2025-08-06
fix(language-switcher): show locales that exist on 404

LeoMcA:
Relates to: https://github.com/mdn/fred/issues/465

Deployed it to testing testing, alongside the related yari and rari changes: https://github.com/mdn/yari/actions/runs/16778636629

Can see it take effect on a page like: https://test.developer.allizom.org/fr/docs/Learn_web_development/Getting_started (you should see that the locales in which the page *does* exist are shown in the language switcher, and it no longer links you to the `${locale}/404` page as before).


argl (approved):
Nice!

PR #524, opened 2025-08-11
fix(observatory): mark illustrative SVGs with `presentation` role

### LINKED ISSUE #521: Observatory has missing alt text on an image
Accessibility report: observatory has missing alt text on an image:

> When I scrolled past the scanner buttons, I noticed an image with no alt text. I don't know what this image is referring to because of a lack of information.

Perhaps this is the `landing-illustration`?

argl:
### Description

This add a11y attributes to illustrative SVGs and makes sure these make it through SVGO postprocessing.

### Motivation

A11y improvements

fixes #521


components/observatory-landing/server.js
```diff
@@ -109,7 +109,7 @@ export class ObservatoryLanding extends ServerComponent {
  */
 function LinesSVG({ className = "" }) {
   return svg`
-  <svg width="75" class="${className}" role="none" height="75" viewBox="-25 0 75 75" version="1.1" id="svg1">
+  <svg width="75" class="${className}" height="75" viewBox="-25 0 75 75" version="1.1" role="presentation">
```
caugner, line 112: According to https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/presentation_role, `none` is a synonym of `presentation`.

caugner (approved):
LGTM.

PS: Although some of these SVGs have semantics, I checked that usages of these SVGs already communicate the same via a visually-hidden text or similar.

Issue #525, opened 2025-08-11
Port offline mode from Yari


globau:

caugner:
Should be resolved via https://github.com/mdn/fred/pull/531, but @LeoMcA is still doing some testing before closing this and labeling `qa-needed`.

LeoMcA:
Tested by me and @mirunacurtean and we're both happy, so closing. Will follow up with some smaller issues which ought not to block release.

PR #531, opened 2025-08-11
feat(plus/offline): add offline mode using yari client

LeoMcA:
Relates to: https://github.com/mdn/fred/issues/525

Would advise reviewing commit-by-commit, as the first commit simply copies over the pwa files from yari (with prettier applied), and the second commit makes the real changes to support offline in fred.

Testable in the review environment (if you do the login dance with replacing the url), but I'll be deploying this specially to test too (to test the yari -> fred "migration").

You'll know it's working if, after downloading, yari starts rendering pages instead of fred. Disable "prefer offline" to get fred rendering pages until you pull the network cable.


caugner (approved):
LGTM, and tested.

1. After enabling MDN Offline Storage, status says "Up to date" (an old issue):
    image
2. After enabling auto-update, it starts downloading:
    image
3. Then it starts unpacking:
    image
4. After finishing unpacking, it shows "Up to date", while "Storage used" still shows "0 MiB":
    image
5. After refreshing the page, the button colors changed, and "Storage used" shows up as "257 MiB":
    image
6. After enabling "Prefer online content", and refreshing:
    image
7. Browsing with "Prefer online content" enabled seems to work (shows fred).
8. Browsing with "Prefer online content" disabled seems to work (shows yari from December).
9. After clicking "Clear data", downloading starts again (probably an old issue):
    image
10. Trying to cancel the download by disabling auto-update, and refreshing, doesn't seem to load (probably an old issue), but works after hard-reload, although status remains "Downloading…" (because we don't support cancelling the download/unpack process).
11. After disabling "MDN Offline Storage" again, this setting persist after a refresh.

vendor/yari/client/pwa/README.md
```diff
@@ -0,0 +1,6 @@
+# MDN Offline
+
+This is MDN's ServiceWorker that powers MDN Offline. It's not being build by
+default for local development. And for now it requires a running instance of
+[kuma](https://github.com/mdn/kuma) to work. We might add support for local
+development and update this README.
```
caugner, line 6: Nit: Remove the unimportant bits.

```suggestion
This is MDN's ServiceWorker that powers MDN Offline.
```

LeoMcA:
> After enabling MDN Offline Storage, status says "Up to date" (an old issue):

I had a go at fixing this in https://github.com/mdn/yari/pull/13386, but this causes problems of its own (the button looks clickable but doesn't do anything until the service worker has cached the static assets and activated). I'll have another attempt at fixing this later today.

package.json
```diff
@@ -85,8 +85,11 @@
     "@types/prismjs": "^1.26.5",
     "@types/webpack-hot-middleware": "^2.25.10",
     "@typescript-eslint/eslint-plugin": "^8.39.1",
+    "@zip.js/zip.js": "^2.7.72",
     "css-loader": "^7.1.2",
     "css-minimizer-webpack-plugin": "^7.0.2",
+    "dexie": "^4.0.11",
+    "downshift": "^9.0.10",
```
caugner, line 92: @LeoMcA Do you remember why you added downshift as a Fred dependency here?

PR #541, opened 2025-08-13
fix(a11y): add breadcrumb & sidebar padding

### LINKED ISSUE #476: Add better focus highlight to breadcrumbs and sidebar with keyboard navigation

### LINKED ISSUE #490: Desktop: Small cut off on the left side of the sidebar when focus is set on the filter

argl:
### Description

Make enough room around breadcrumb items to allow for the `focus-visible` border to not be clipped.

### Motivation

The previous version was looking broken (while still functional), yari parity

#### Before:
image

#### After:
image

fixes #476 
fixes #490

### DIFF
--- a/components/left-sidebar/server.css
+++ b/components/left-sidebar/server.css
@@ -145,6 +145,7 @@
   padding-top: 0.5rem;
   padding-right: 0.5rem;
   padding-bottom: 2rem;
+  padding-left: 2px;
 
   background: linear-gradient(
     to bottom,
--- a/components/left-sidebar/server.css
+++ b/components/left-sidebar/server.css
@@ -131,6 +131,12 @@
   }
 }
 
+.left-sidebar__content {
+  > ol {
+    padding-left: 2px;
+  }
+}
+
 .left-sidebar__header {
   position: sticky;
   top: 0;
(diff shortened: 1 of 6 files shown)


caugner (changes requested):
Can we also add `padding-left: 2px` to the `.left-sidebar` to fix the issue there, and to ensure the left side of the breadcrumbs is aligned with the sidebar?

image

This would also fix the sidebar filter:

image


PS: Technically, we could then add `.left-sidebar em { margin-left: -2px; }` (which didn't work previously).
PPS: We seem to have the same issue with the placement:
image

argl:
Ah interesting, the sidebar suffers the same problem. I'll see to that as well.

components/left-sidebar/server.css
```diff
@@ -131,6 +131,12 @@
   }
 }
 
+.left-sidebar__content {
+  > ol {
+    padding-left: 2px;
```
caugner, line 136: Why not add the `padding-left` once on `.left-sidebar`?

components/placement-sidebar/element.css
```diff
@@ -1,3 +1,7 @@
+:host {
+  padding-left: 2px;
+}
```
caugner: I think the padding should be outside of the custom element (`.reference-layout__toc` would be consistent with `.left-sidebar`).

caugner (approved):
Awesome, LGTM.

PR #542, opened 2025-08-13
fix(interactive-examples): improve a11y + fix unsupported icon

### LINKED ISSUE #522: Improve CSS Interactive Examples accessibility
- CSS interactive example options don’t announce - recommended to use buttons & not form fields
- CSS interactive examples have no context for form fields
- No feedback on CSS interactive examples to understand what has changed when different styles are selected

### LINKED ISSUE #12776: Unsupported indicator does not meet accessibility criteria
### Summary

Unsupported indicator does not meet accessibility criteria

On smaller screens there is only an orange outline and no icon.
On larger screens there is an orange outline and an icon but no textual information.
The icon isn't part of the accessibility tree and won't be announced by screen readers.

In no case is there any explanation what any of this means.
Users have to guess that this has to do with browser support status.

Instead I would expect:
- information in text
- everything is part of the accessibility tree
- not using color alone to convey meaning on smaller screens

### URL

https://developer.mozilla.org/en-US/docs/Web/CSS/hyphens#try_it

### Reproduction steps

1. I set `hyphen: unsupported` in the demo
2. I see the "this feature is unsupported" indication

### Expected behavior

I expected this feature to have text information and to be accessible

### Actual behavior

Feature is not accessible and doesn't explain itself

### Device

Other (specify below)

### Browser

Other (specify below)

### Browser version

Other version (specify below)

### Operating system

Other (specify below)

### Screenshot

On smaller screens:

Image

On larger screens:

Image

### Anything else?

_No response_

### Validations

- [x] I have read the [Community Participation Guidelines](https://www.mozilla.org/en-US/about/governance/policies/participation/).
- [x] I have verified that there isn't [already an issue](https://github.com/mdn/yari/issues) that reports the same bug to avoid creating a duplicate.
- [x] I have checked that this is a concrete bug. For Q&A open a [GitHub Discussion](https://github.com/orgs/mdn/discussions/categories/platform).

caugner:
### Description

Bundles a couple of fixes, mainly accessibility of interactive examples:

1. [fix(play): add aria-live to runner/console](https://github.com/mdn/fred/commit/f3f5982e54166dd114928f33e9cb6a2b7ef0a487) should cause screen readers to announce output/console changes (e.g. when clicking "Run")
2. [fix(interactive-example): make choices a list](https://github.com/mdn/fred/commit/8acad00906c49443fb4631089c7acbd206d5c758) should cause screen readers to announce the CSS interactive examples choices as a list
3. [fix(interactive-example): add aria-label to choices](https://github.com/mdn/fred/commit/2f352776531a7df8ad8cf5a20227138d33fb4565) should cause screen readers to announce the interactive examples as a group using their heading
4. [fix(interactive-example): disable Reset until choice update](https://github.com/mdn/fred/pull/542/commits/03b4151cc34a84c40cee856039a56bc153b81e40) should avoid some confusion for screen reader users
5. [fix(interactive-examples): announce unsupported value](https://github.com/mdn/fred/commit/ef28cd46d6e3b7b6d74a8e5e0f384a8dade0e312) should fix [this issue reported in yari](https://github.com/mdn/yari/issues/12776)
6. [fix(interactive-example): show alert icon if unsupported](https://github.com/mdn/fred/commit/dcde79fd3a4fee22dba096ce5144aef32c610fbd) fixes a regression I noticed while fixing the previous item

### Motivation

Improve accessibility of interactive examples, and fix the unsupported CSS property/value icon.

Fixes https://github.com/mdn/fred/issues/522.


components/play-editor/element.js
```diff
@@ -59,7 +59,6 @@ export class MDNPlayEditor extends LitElement {
       });
       this._editor.setState(state);
     }
-    this.dispatchEvent(new Event("update", { bubbles: true, composed: true }));
```
caugner: Note: We already dispatch the `update` event below using a `EditorView.updateListener`.

components/play-runner/element.js
```diff
@@ -160,6 +160,7 @@ export class MDNPlayRunner extends LitElement {
             ...(this.sandbox?.split(" ") || []),
           ]),
         ].join(" ")}
+        aria-live="polite"
```
caugner, line 163: Note: I don't know if this (`aria-live` on an `iframe`) has actually any effect.

components/play-console/element.js
```diff
@@ -114,7 +114,7 @@ export class MDNPlayConsole extends LitElement {
 
   render() {
     return html`
-      <ul>
+      <ul aria-live="polite">
```
caugner, line 117: Note: This should have an effect, but I wasn't able to confirm.

components/play-console/element.js
```diff
@@ -114,7 +114,7 @@ export class MDNPlayConsole extends LitElement {
...
         ${this._messages.map(
           (message) => html`<li><code tabindex="0">${message}</code></li>`,
```
caugner: Note: I have previously attempted to make this message "accessible" by adding `tabindex="0"`, but if `aria-live` works, then this should no longer be needed.

```suggestion
          (message) => html`<li><code>${message}</code></li>`,
```

components/interactive-example/with-choices.js
```diff
@@ -104,23 +111,29 @@ export const InteractiveExampleWithChoices = (Base) =>
...
+        <div class="template-choices" aria-labelledby=${id}>
           <header>
-            <h4>${decode(this.name)}</h4>
-            <mdn-button id="reset" @click=${this._reset} variant="secondary"
+            <h4 id=${id}>${decode(this.name)}</h4>
+            <mdn-button
+              id="reset"
+              @click=${this._reset}
+              variant="secondary"
+              .disabled=${!this.__choiceUpdated}
```
caugner, line 124: Should probably also add a `title` to explain the effect of activating this button.

components/interactive-example/with-choices.js
```diff
@@ -104,23 +111,29 @@ export const InteractiveExampleWithChoices = (Base) =>
...
               >${this.l10n`Reset`}</mdn-button
             >
           </header>
-          <div
+          <ul
```
caugner, line 128: Should probably also add an `aria-label` to explain this list.

components/interactive-example/element.css
```diff
@@ -273,25 +273,21 @@ mdn-ix-tab-wrapper {
...
-          content: "";
+          background-color: var(--color-unsupported);
 
-          background-image: url("../icon/triangle-alert.svg");
-          background-repeat: no-repeat;
-          background-position: center center;
-          background-size: 1rem;
-
-          opacity: 1;
+          mask-image: url("../icon/triangle-alert.svg");
```
caugner, line 288: Note: It looks like this had regressed in https://github.com/mdn/fred/commit/08a03380352cd09047fa964ac97ce1605a3b19b8.

components/interactive-example/with-choices.js
```diff
@@ -104,23 +111,29 @@ export const InteractiveExampleWithChoices = (Base) =>
...
             class="choice-wrapper"
             @click=${this.#choiceFocus}
             @focus=${this.#choiceSelect}
             @update=${this.#choiceUpdate}
```
caugner, line 132: Address https://github.com/mdn/fred/pull/542/files/dcde79fd3a4fee22dba096ce5144aef32c610fbd#r2273743349

```suggestion
            @update=${this.#choiceUpdate}
            aria-label=${this.l10n`Value select`}
```

components/interactive-example/with-choices.js
caugner: Resolved via 4e7fc4f0343f80b94b3686aff8c6df6d0c9313f1.

components/play-editor/element.js
LeoMcA: I'm almost certain I added this to fix a bug: some kind of race condition when values were being set before the editor had initialised, or something. I'll take a deeper look (along with if there's any alternative for the dependent changes)

caugner: Let me back those two changes out into a separate PR to unblock.

caugner: Done, extracted to https://github.com/mdn/fred/pull/549.

components/interactive-example/with-choices.js
```diff
@@ -133,11 +138,17 @@ export const InteractiveExampleWithChoices = (Base) =>
                     minimal="true"
                     .delay=${100}
                     .value=${code?.trim()}
+                    aria-label=${ifDefined(
+                      this.__choiceUnsupported[index]
+                        ? this
+                            .l10n`The current value is not supported by your browser.`
+                        : undefined,
+                    )}
```
caugner, line 155: Notes:

1. This could deserve a label when the current value is supported.
2. This might be eligible for some post-launch improvements (such as not selecting the current option as a side-effect of focus).

components/interactive-example/with-choices.js
caugner: Probably better to do as part of https://github.com/mdn/fred/pull/549. There are some other "Reset" buttons.

components/interactive-example/with-choices.js
LeoMcA: I think it not having a label implicitly suggests it is supported - so I wouldn't add one.

I think auto selecting the option when focusing is nice: why add an extra step to select the option? But happy to see it changed if we receive specific feedback about it being bad.

components/play-console/element.js
LeoMcA: Could `tabindex="0"` be hiding this change? If we had to choose, I'd go with the more "standards compliant" approach, which seems to be this one.

LeoMcA (approved):
Looks good to me, I'll leave it to you whether you want to remove `tabindex=0` and merge!

components/play-console/element.js
caugner: So let's remove `tabindex="0"` along with this change (i.e. apply [this suggestion](https://github.com/mdn/fred/pull/542/files#r2273738382))?

caugner:
Merged https://github.com/mdn/fred/pull/549 back into this one.

components/play-console/element.js
caugner: Removed in 9ce0b38e1f022ac0bf00f37f0f0fcf2799ff0a16.

caugner:
Merging now, as the changes here, and the merged changes were both were already approved.

PR #544, opened 2025-08-14
fix(site-search): add h1 for a11y

### LINKED ISSUE #535: Add h1 to site search results
The search results page has no `<h1>`.

caugner:
### Description

Adds an h1 to the "Search" page (aka "site-search"), ~~looking like the h1 on the "Page not found" page~~ visually hidden like on Google.

### Motivation

Improve accessibility, ~~and consistency~~.

Fixes https://github.com/mdn/fred/issues/535.

### DIFF
--- a/components/site-search/server.css
+++ b/components/site-search/server.css
@@ -6,19 +6,8 @@
   @media (--screen-layout-no-sidebar) {
     grid-template-columns: var(--layout-no-sidebar);
   }
+}
 
-  h1,
-  mdn-site-search {
-    grid-column: content;
-  }
-
-  /* Same as: `.not-found h1` */
-  h1 {
-    margin-top: 0;
-
-    font-size: var(--font-size-largest);
-    font-weight: normal;
-
-    line-height: var(--font-line-content);
-  }
+mdn-site-search {
+  grid-column: content;
 }
--- a/components/site-search/server.css
+++ b/components/site-search/server.css
@@ -6,8 +6,19 @@
   @media (--screen-layout-no-sidebar) {
     grid-template-columns: var(--layout-no-sidebar);
   }
-}
 
-mdn-site-search {
-  grid-column: content;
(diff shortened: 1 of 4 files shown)


components/site-search/server.css
```diff
@@ -6,8 +6,19 @@
...
+  }
+
+  /* Same as: `.not-found h1` */
+  h1 {
+    margin-top: 0;
+
+    font-size: var(--font-size-largest);
+    font-weight: normal;
+
+    line-height: var(--font-line-content);
```
argl: This looks very large, costing a lot of screen estate. Maybe cut back on the font size / line height? My understanding is that this is primarily an a11y / semantic issue, visually this is not really needed to use the page.

Putting it in front of the input field would also be a possibility, minimizing the impact on vertical space, but then the wording should be slightly different: "Search for [input field]", assuming that would work in all languages.

caugner: No strong opinions, happy to remove those styles.

This is how it would look without those three lines:
image

```suggestion
```

LeoMcA: Google seems to have a visually hidden `<h1>` on their results page, I wonder if we should take the same approach:

image

caugner: Visually hidden via d3090b4. 👍

components/site-search/server.js
```diff
@@ -11,6 +11,7 @@ export class SiteSearch extends ServerComponent {
     return PageLayout.render(
       context,
       html`<div class="site-search">
+        <h1 class="visually-hidden">${context.l10n`Search`}</h1>
```
LeoMcA, line 14: Perhaps "Search Results"? Provides a bit more of a clue there's results after the input

LeoMcA (approved):
One nit

argl (approved):
LGTM

PR #547, opened 2025-08-14
feat(menu): link to en-US if page not available in current locale

### LINKED ISSUE #546: Link to en-US if menu link not available in current locale

caugner:
### Description

Falls back to en-US for menu links not available in the current locale, with an indicator; using a hard-coded list of missing docs, along with a script to check if any were added.

_Note_: This PR also changes `Web/HTML#guides` to `Web/HTML/Guides` as agreed directly with @Rumyra.

### Motivation

Avoid linking to 404s.

### Additional details

pt-BR:
image

fr:
image

Fixes https://github.com/mdn/fred/issues/546.


caugner:
Marking as draft while I generalize for all menu links.

caugner:
> Marking as draft while I generalize for all menu links.

Abandoned the idea of generalizing for _all_ menu links, but at least added it for Curriculum to be somewhat consistent, see 1b81655f0bac025e3b488f4cb08db3b7c380c966:

> For the other links with icon, it is less trivial, because
> those are currently using an inline-flex with a gap, which
> would cause the indicator to be vertically centered, and
> have an unexpected gap next to them.
> 
> We also cannot add the indicator behind Blog, because there is
> not necessarily enough space, so this might break the layout.

components/menu/server.js
```diff
@@ -1,12 +1,44 @@
...
    * @param {import("@fred").Context} context
    */
   render(context) {
+    /**
+     * Renders a link to a page.
+     *
+     * @param {string} slug
+     * @param {string} text
+     * @param {string} [label]
+     * @returns
```
LeoMcA:
```suggestion
```

components/menu/server.js
```diff
@@ -352,59 +317,30 @@ export class Menu extends ServerComponent {
...
-                        </a>
-                      </li>
-                      <li>
-                        <a href=${`/${context.locale}/docs/Web/API/Push_API`}>
-                          Push API
-                        </a>
+                        ${link("Web/API/Geolocation_API", "Geolocation API")}
                       </li>
+                      <li>${link("Web/API/HTML_DOM_API", "HTML DOM API")}</li>
+                      <li>${link("Web/API/Push_API", " Push API ")}</li>
```
LeoMcA:
```suggestion
                      <li>${link("Web/API/Push_API", "Push API")}</li>
```

LeoMcA (approved):
nice stuff, a couple of small fixes:

PR #549, opened 2025-08-14
enhance(interactive-examples): disable choice Reset until updated

caugner:
### Description

Disable the "Reset" button on the CSS interactive examples with value choices.

### Motivation

Avoids confusion, especially for screen reader users.

Extracted from https://github.com/mdn/fred/pull/542.

Related to https://github.com/mdn/fred/issues/522.

### DIFF
--- a/components/interactive-example/with-choices.js
+++ b/components/interactive-example/with-choices.js
@@ -23,6 +23,7 @@ export const InteractiveExampleWithChoices = (Base) =>
     static properties = {
       __choiceSelected: { state: true },
       __choiceUnsupported: { state: true },
+      __choiceUpdated: { state: true },
     };
 
     /** @param {any[]} _args  */
@@ -32,6 +33,8 @@ export const InteractiveExampleWithChoices = (Base) =>
       this.__choiceSelected = -1;
       /** @type {boolean[]} */
       this.__choiceUnsupported = [];
+      /** @type {boolean} */
+      this.__choiceUpdated = false;
     }
 
     /** @param {MouseEvent} event  */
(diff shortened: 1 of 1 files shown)


argl (approved):
Nice!

caugner:
⚠️ Let's hold off merging until @LeoMcA has checked it doesn't cause a regression.

See: https://github.com/mdn/fred/pull/542#discussion_r2276934673

components/play-editor/element.js
```diff
@@ -59,7 +59,6 @@ export class MDNPlayEditor extends LitElement {
       });
       this._editor.setState(state);
     }
-    this.dispatchEvent(new Event("update", { bubbles: true, composed: true }));
```
LeoMcA: Yes, as I suspected, this causes the weird race condition to come back: I think it's to do with the order the components get loaded in. Let me debug a bit more, and either come up with a different solution for the reset, or a different solution to this race.

You should be able to reproduce by [loading up the kitchen sink](https://fred-pr549.review.mdn.allizom.net/en-US/docs/MDN/Kitchensink#try_it_3) and refreshing until you see:

image

LeoMcA (changes requested):

components/play-editor/element.js
LeoMcA: I was wrong, this is an unrelated issue: https://github.com/mdn/fred/issues/606

But I was right about this line being important, it ensures we save the state after formatting in the playground, however I've got a better solution for that in this PR: https://github.com/mdn/fred/pull/607

LeoMcA (approved):
Looks good, though as this PR stands it breaks saving state in playground after formatting: https://github.com/mdn/fred/pull/607 fixes that in a way compatible with this PR, so please review and merge that, before merging this!

caugner:
I'll merge this back into https://github.com/mdn/fred/pull/542.

PR #554, opened 2025-08-15
feat(placement): playground placement on right side

### LINKED ISSUE #553: Move the placement on the playground page to the right column
In preparation for new placement formats, we cannot support the "horizontal" vasiant of the sidebar placement any more because there is no horizontal skyscraper.

Solution is to put it on the right sidebar like on other pages, with appropriate logic for smaller screens.

Also show the "normal" variant of native placements instead of the horizontal version.

argl:
### Description

In anticipation of the new placement sizes, we move it to right side on the playground, also disabling the horizontal layout for the classic native placements.

### Motivation

New placement sizes

### Additional details

Native:
image

Skyscraper:
image

fixed #553


LeoMcA (commented):
I like the fact the playground currently expands to the full width of the page: it seems like once place you *do* want to maximise screen usage.

There's also some extra padding on the right of the banner which seems off.

argl:
Good point, I'll change that.

LeoMcA (commented):
apologies, started poking around the css to review, but needed to make changes to see if they'd work: ended up making quite a few changes, which I've opened as  PR against this branch for your review: https://github.com/mdn/fred/pull/608

LeoMcA (approved):

PR #561, opened 2025-08-18
feat: better print/pdf support for docs and observatory

### LINKED ISSUE #540: Improve printing behaviour of Observatory results page
We got feedback requesting PDF export for observatory results. Rather than doing that directly, it was decided that improving the print behaviour would be the way to go since that inlies PDF support as well and is well supported by the web platform.

Make sure the tabs regardless of the current tab selection all print out at once, remove auxiliary stuff like navigation as much as needed.

argl:
### Description

Add print css styles on global level and for observatory specifically. This is far from perfect, but a long way from the previous basically unusable state. The print specific CSS is not a lot, so no reason to ship it separately/demand.

### Motivation

Better print behaviour/PDF export on observatory results (including collateral improvements for reference pages)

### Additional details

Here is a demo PDF exported via the print dialog:
[Scan results - HTTP Observatory _ MDN.pdf](https://github.com/user-attachments/files/21837302/Scan.results.-.HTTP.Observatory._.MDN.pdf)

Here is a reference page PDF exported via the print dialog:
[Boolean.pdf](https://github.com/user-attachments/files/21837366/Boolean.pdf)

fixes #540


LeoMcA (approved):
looks good!

PR #564, opened 2025-08-19
chore(article-footer): point "Learn how to contribute" link to "Getting started" page

bsmth:
### Description

Pointing to our MDN docs for contribution hints instead of to GitHub READMEs.

### Motivation

The contribution docs are more user-friendly, easier to navigate, and are more comprehensive than the GitHub README.

Similar to:

- [x] https://github.com/mdn/generic-content/pull/19


components/article-footer/server.js
```diff
@@ -36,11 +36,10 @@ export class ArticleFooter extends ServerComponent {
  */
 function Contribute(context) {
   return html`<a
-    class="article-footer__contribute external"
-    href="https://github.com/mdn/content/blob/main/CONTRIBUTING.md"
-    title=${context.l10n`This will take you to our contribution guidelines on GitHub.`}
+    class="article-footer__contribute"
+    href="/en-US/docs/MDN/Community/Getting_started"
```
caugner: This en-US link would also appear in other locales.

caugner (commented):
Is the status quo a regression versus prod?

bsmth:
> Is the status quo a regression versus prod?

There's no regression, this is just a chore for future. We can consider it low-prio

components/article-footer/server.js
bsmth: What's the best approach here? I'm just copying what we're doing in other places

caugner: Two options:

1. If this page is already translated in all locales, then we can do:
  ```suggestion
      href="/en-US/docs/MDN/Community/Getting_started"
  ```
2. If it isn't, then we'd have to maintain a list of locales where it's available, and link to en-US elsewhere, like we do in the menu.

caugner: It seems to be available in all locales:

```suggestion
    href=${`/{context.locale}/docs/MDN/Community/Getting_started`}
```

components/article-footer/server.js
```diff
@@ -36,11 +36,10 @@ export class ArticleFooter extends ServerComponent {
  */
 function Contribute(context) {
   return html`<a
-    class="article-footer__contribute external"
-    href="https://github.com/mdn/content/blob/main/CONTRIBUTING.md"
-    title=${context.l10n`This will take you to our contribution guidelines on GitHub.`}
+    class="article-footer__contribute"
+    href=${`/{context.locale}/docs/MDN/Community/Getting_started`}
+    title=${context.l10n`This will take you to our contribution documentation.`}
```
caugner: Nit: I think this title no longer makes sense, as it just duplicates the text.

```suggestion
```

caugner (approved):

components/article-footer/server.js
caugner: @pepelsbey Any thoughts on this one?

bsmth: I'm +1 on removing, does this mostly make sense for external links?

caugner: Actually, I realize that this link opens in a new tab. If we keep that, then the title should have a suffix "(opens in new tab)".

In general, I would say a `title` makes sense only where the link text is ambiguous.

pepelsbey: I don't think we should be using titles at all. They are unavailable on touch devices, barely accessible on desktop (too small, don't scale), and give us false “they will hover and figure it out” safety. If you can't figure out what the link or button does by looking at it (or reading the primary label), it’s not right. Removing titles will force us to think harder and make better decisions.

bsmth: Thanks, both. I think this is ready, now!

components/article-footer/server.js
```diff
@@ -36,11 +36,9 @@ export class ArticleFooter extends ServerComponent {
  */
 function Contribute(context) {
   return html`<a
-    class="article-footer__contribute external"
-    href="https://github.com/mdn/content/blob/main/CONTRIBUTING.md"
-    title=${context.l10n`This will take you to our contribution guidelines on GitHub.`}
+    class="article-footer__contribute"
+    href=${`/${context.locale}/docs/MDN/Community/Getting_started`}
     target="_blank"
```
caugner: Does this still need to open in a new tab? I don't think we open internal links in a new tab anywhere else.

```suggestion
```

caugner (changes requested):
Last round.

components/article-footer/server.js
bsmth: > Does this still need to open in a new tab?

nope! Let's get rid of it

caugner (approved):

PR #566, opened 2025-08-19
fix: various components were using old text color vars

LeoMcA:
Cleans up old uses of text color variables, and fixes an issue with the leaderboard ad style on dark mode:

Before:

image

After:

image


argl (approved):
Nice cleanup!

PR #568, opened 2025-08-19
fix(light-dark): fix colors and theme switcher when polyfilled

LeoMcA:
Tested against Firefox ESR 115.27.0 and Safari iOS 16.4, and release Firefox and Safari (to ensure no regressions).

A couple of issues here:
- The `light-dark` polyfill was stepping on its own toes, and generating overlapping variables when we combined css files from each server component
  - This is fixed by adding a hash into the variables
- The theme switcher wasn't working because it was setting `color-scheme` directly from javascript
  - This is fixed by only setting the `data-theme` attribute, and using that to set `color-scheme` in css: allowing the polyfill to access it

Before:

Screenshot 2025-08-19 at 12-26-57 JavaScript

After:

Screenshot 2025-08-19 at 12-27-44 JavaScript

### DIFF
--- a/build/loaders/fix-light-dark.js
+++ b/build/loaders/fix-light-dark.js
@@ -1,6 +1,8 @@
 import crypto from "node:crypto";
 
 /**
+ * Adds a per-file identifier to the CSS variables used by the light-dark polyfill,
+ * to avoid conflicts between CSS files.
  * @this {import("@rspack/core").LoaderContext}
  * @param {string} source
  */
--- a/build/loaders/fix-light-dark.js
+++ b/build/loaders/fix-light-dark.js
@@ -1,0 +1,21 @@
+import crypto from "node:crypto";
+
+/**
+ * @this {import("@rspack/core").LoaderContext}
+ * @param {string} source
+ */
+export default function postCssLightDarkFix(source) {
+  const id = hash(this.resourcePath);
+  return source.replaceAll(
+    "--csstools-light-dark-toggle--",
+    `--csstools-light-dark-toggle-${id}-`,
+  );
+}
+
+/**
+ * @param {string} str
+ * @returns {string}
+ */
+function hash(str) {
+  return crypto.createHash("sha256").update(str).digest("hex").slice(0, 8);
+}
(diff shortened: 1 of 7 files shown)


rspack.config.js
```diff
@@ -37,33 +37,38 @@ const OPTIMIZATIONS = {
...
-        [
-          "postcss-preset-env",
-          {
-            stage: 2,
-            minimumVendorImplementations: 2,
-          },
+const postcssLoaders = (lit = false) => [
+  {
+    loader: "./build/loaders/fix-light-dark.js",
+  },
```
caugner: Nit: Would be good to have a comment explaining (beyond git log) why this is necessary.

components/outer-layout/server.js
```diff
@@ -74,7 +74,7 @@ export class OuterLayout extends ServerComponent {
       <!doctype html>
       <html
         lang=${context.locale}
-        style="color-scheme: light dark;"
+        data-theme="light dark"
```
caugner, line 77: Note: We're already using [`data-theme`](https://github.com/search?q=repo%3Amdn%2Ffred%20%2Fdata-theme%2F&type=code) in other places. Are we sure this causes no conflicts?

caugner (approved):
LGTM overall, just a nit and a note. (A bit surprised about those `importLoaders` constant numbers.)

Tested that theme switching works in Firefox 115 on macOS.

build/loaders/fix-light-dark.js
```diff
@@ -0,0 +1,21 @@
...
+ * @this {import("@rspack/core").LoaderContext}
+ * @param {string} source
+ */
+export default function postCssLightDarkFix(source) {
+  const id = hash(this.resourcePath);
+  return source.replaceAll(
+    "--csstools-light-dark-toggle--",
+    `--csstools-light-dark-toggle-${id}-`,
+  );
+}
```
caugner: Nit: Some comment explaining this would be nice here as well.

components/outer-layout/server.js
LeoMcA: Will double check those instances we use it, but theoretically I'm sure: the theme switcher was already setting both `style="color-scheme: blah;"` and `data-theme="blah"` on `<html>` now it just does the latter (and so we update our pre-js default here to match).

LeoMcA: Yep, double checked on the curriculum homepage

Issue #572, opened 2025-08-19
Sidebar filter search results highlights should be more visible in dark mode


mirunacurtean:
### What information was incorrect, unhelpful, or incomplete?

1. Go to https://developer.allizom.org/de/docs/Learn_web_development/Core/Text_styling 
2. Select the sidebar filter ad type in a keyword. 
3. Observe the results. 

<img width="1835" height="1057" alt="Image" src="https://github.com/user-attachments/assets/2a58e43a-da48-43f3-b364-3b1a933bedbe" />

### What did you expect to see?

This would be an accessibility improvement

caugner:
We could maybe set `.left-sidebar mark { color: var(--color-area-link); }`.

<img width="749" height="439" alt="Image" src="https://github.com/user-attachments/assets/9d2d3f31-e90a-4a95-bcdd-22176c167a14" />

<img width="749" height="439" alt="Image" src="https://github.com/user-attachments/assets/3bdc8447-5ad5-4f02-b103-e929ac2f2fb4" />

globau:
Might be fixed by #588

PR #588, opened 2025-08-20
feat(sidebar-filter): refactor, always expand, increase font-size

### LINKED ISSUE #602: UI: Filter Icon
Image

On Safari 18.6 (Mac M3 Sequoia) the filter icon appears too big with the text appearing a lot closer to it.

Note: This doesn't happen on Chrome 139 or Firefox 141 for the same OS.

pepelsbey:
### Description

- Fixes the zoom-in behavior in mobile Safari
- Removes the “matches” word from the counter to save space for the query
- Removes the collapsing behavior: less jumping, easier to enter the field

Simplifies the implementation:

- Simpler (scoped) class names
- Grid layout instead of positioning
- CSS `:placeholder-shown` instead of `_active` state for hiding the clear button

### Relevant issues it fixes

Fixes https://github.com/mdn/fred/issues/602
- ~~https://github.com/mdn/fred/issues/572~~

### Before/after

<details>
<summary>Not in focus</summary>

before

after

</details>

<details>
<summary>In focus</summary>

before-focus

after-focus

</details>

### Zoom-in behavior in Safari

<details>
<summary>Before</summary>

https://github.com/user-attachments/assets/a531e53e-7c7d-4771-a54d-2ddae0b5e313

</details>

<details>
<summary>After</summary>

https://github.com/user-attachments/assets/0796bfe4-6f18-494c-b76c-753aea08b5e2

</details>


caugner:
@pepelsbey Can you please take a look at https://github.com/mdn/fred/issues/572 to see if it can be fixed as part of this?

Edit: Also, I'm noticing two things in the Zoom-in behavior After screen recording (at 0:04):

1. There are two expand/collapse icons.
2. There is some text overflowing/overlaying at the bottom.

Assuming you can repro those, can you please check whether those are new regressions in this PR, and how they could be fixed?

image

components/sidebar-filter/element.js
```diff
@@ -152,45 +140,29 @@ class MDNSidebarFilter extends L10nMixin(LitElement) {
...
+        autocomplete="off"
+        class="input"
+        type="text"
+        placeholder=${this.l10n`Filter`}
+        .value=${this.query}
+        @input=${this._onInput}
+      />
+      ${this.matchCount === undefined
+        ? ""
+        : html` <span class="counter"> ${this.matchCount} </span> `}
```
caugner, line 157: 1. This looks like an accessibility regression, given that the number isn't explained (or was this already ignored by screen-readers before?).
2. Nit: We should be able to remove the white-space around `${this.matchCount}`.

caugner (changes requested):
Thanks for your work on these improvements. I can see you've put quite some effort into it. 🙌 

Before going into the details, I think it might be helpful for future PRs to first open an issue for triage. That way, we can prioritize and align before too much work goes in.

To make this PR easier to review, could you consider separating the fixes into separate commits, or adding more context in the PR description, or via review comments? That would help me understand the motivation, and effect of each change more clearly.

I've looked at [the preview](https://fred-pr588.review.mdn.allizom.net/en-US/docs/Web/CSS) in comparison to [the main branch](https://fred.review.mdn.allizom.net/en-US/docs/Web/CSS).

---

> Fixes the zoom-in behavior in mobile Safari

From the screen recordings, it looks like the zoom-in on focus is no longer happening after your changes.

Is that the expected behavior on iOS? Do you know what caused the zoom originally, and what in your change prevents it now?

> Removes the “matches” word from the counter to save space for the query

From the screenshots, it's not clear to me that space is actually a problem. I did notice that the sidebar filter font size looks larger on this branch (do you know why?), which could contribute. In practice, I'd expect filter terms to be short, since the list usually narrows down quickly.

Was this change based on feedback, or can you share a practical example?

Even if we remove the word "matches" visually, it might still make sense to keep it for screen readers, or otherwise convey what the number represents.

> Removes the collapsing behavior: less jumping, easier to enter the field

It looks like the filter is now always expanded, and more prominent (especially with the increased font size).

Could you explain a bit more about the "jumping" issue, and entering the field was difficult? Was this a regression in fred, or something reported by the community?

My concern here is that making the filter more prominent could distract from the rest of the sidebar. Since most users likely use the sidebar without the filter, I'm not sure we want to draw extra attention to it.

pepelsbey:
> There is some text overflowing/overlaying at the bottom.

This is another issue, not a regression. I would file it separately, it's a different component and would prefer to keep this PR focused on the field, not the whole sidebar.

components/sidebar-filter/element.js
pepelsbey: The counter text wasn’t accessible before. Missing “matches” word doesn’t make any difference in this regard.

pepelsbey:
> From the screen recordings, it looks like the zoom-in on focus is no longer happening after your changes. Is that the expected behavior on iOS? Do you know what caused the zoom originally, and what in your change prevents it now?

Zoom-focusing is an annoying behavior that only happens for input fields with the `font-size` smaller than 16px. This is the main reason for the increased font size.

pepelsbey:
> From the screenshots, it's not clear to me that space is actually a problem. I did notice that the sidebar filter font size looks larger on this branch (do you know why?), which could contribute. In practice, I'd expect filter terms to be short, since the list usually narrows down quickly.

The number of matches is a questionable feature in general. It’s hard for me to imagine a use case where the number of items found matters. Leaving only the number not only leaves more space for typing but also partially removes distraction.

pepelsbey:
> Could you explain a bit more about the "jumping" issue, and entering the field was difficult? Was this a regression in fred, or something reported by the community? My concern here is that making the filter more prominent could distract from the rest of the sidebar. Since most users likely use the sidebar without the filter, I'm not sure we want to draw extra attention to it.

Simplified filter behavior is a side-effect of the refactoring. Our sidebars are usually quite long and could be hard to navigate. I don’t see how a more prominent filter would distract users from the sidebar. On the contrary, it would only highlight its existence and would help navigate sidebars.

Rumyra:
I'm happy with the styling change

components/sidebar-filter/element.js
caugner: Using VoiceOver, when I type into the filter, and let it start reading (Control + Option + A), it reads "forty-two matches", but with your change, it only reads "forty-two", without indicating what this number stands for. To me this indicates an a11y regression.

caugner:
To address the concerns around confusion and distraction, may I suggest making it look like a filter (only bottom-border, no border-radius)?

image

We could also put the filter icon on the right, and have it replaced with the "x".

image

pepelsbey:
> may I suggest making it look like a filter (only bottom-border, no border-radius)?

Microsoft made a mistake with some Windows versions, making text fields look like not text fields (the same underlined text), and regretted it very much. Let's avoid making the same mistake and keep text fields consistent.

components/sidebar-filter/element.js
pepelsbey: The number of found options is not announced in any way while you're typing in the focused field, making this span next to it rather useless. Yes, you can technically move to the next text node, but it’s not a behavior I would expect. I think we should solve this issue separately, or just drop the number of found items. But it’s not an a11y regression from my point of view (it wasn't accessible in the first place).

caugner (approved):
🎁

PR #592, opened 2025-08-20
fix(build): add public assets, respect BUILD_OUT_ROOT

### LINKED ISSUE #574: Static resources including favicon missing from fred build
Yari would copy [assets from `client/public`](https://github.com/mdn/yari/tree/main/client/public) into the build output dir.

We need to do the same before re-enabling [yari's prod build workflow](https://github.com/mdn/yari/actions/workflows/prod-build.yml). We have uploaded them manually for now.

We're sure we don't need all these assets, but we're not entirely sure which are necessary. Narrowing that down is lower priority than getting the workflow working without manual intervention.

Solutions could be:
- Replicating this copy in fred's rspack
  - I'm not sure this is the best solution for everything, but probably a good idea for things fred actually relies on - such as the favicon (and the service worker, which we manually copy around in the yari build)
- Copying the contents of the folder into `out` manually in the build workflow in yari
  - Might be the best short term solution, but doesn't handle the hashed items: though these shouldn't be referred to anywhere

### LINKED ISSUE #576: Fred has no opensearch.xml
Yari declared it:

https://github.com/mdn/yari/blob/9177f0c8498a070bd338bbb3d1dff0de16d4f475/ssr/render.tsx#L178C1-L183C13
https://github.com/mdn/yari/blob/9177f0c8498a070bd338bbb3d1dff0de16d4f475/client/public/opensearch.xml

Fred doesn't.

LeoMcA:
### Description

- Copy over assets from https://github.com/mdn/yari/tree/main/client/public which are either referred to in fred, or see requests in our cloud logs (due to being cached on external sites, like some of the `favicon-*.png` files).
- Add missing link tags to some of these assets
- Make build process respect `BUILD_OUT_ROOT`, and copy public assets and service worker directly into root, rather than relying on yari build process to do this
  - Requires changes in yari: https://github.com/mdn/yari/pull/13419

Writing this makes me realise I should document how `/public` works and what it's for: will follow up.

Partially fixes https://github.com/mdn/fred/issues/574
Fixes https://github.com/mdn/fred/issues/576


build.sh
```diff
@@ -7,9 +7,7 @@ if [ -z "$CONTENT_ROOT" ]; then
 fi
 
 npm install
-mkdir -p out/static
 export BUILD_OUT_ROOT=out
 npm run rari build
 npm run build
 node build/ssr.js
```
caugner: Nit: While we're here

```suggestion
npm run ssr
```

.github/workflows/_deploy.yml
```diff
@@ -236,8 +236,6 @@ jobs:
         working-directory: mdn/fred
         run: |
           npm run ssr
```
caugner: Nit: Unnecessary multiline

```suggestion
        run: npm run ssr
```

build/env.js
```diff
@@ -0,0 +1,7 @@
+import path from "node:path";
+import { fileURLToPath } from "node:url";
+
+const __dirname = path.dirname(fileURLToPath(import.meta.url));
```
caugner: Nit: Not sure if we can use `import.meta.dirname` given it's still experimental in Node.js 22.

```suggestion

const __dirname = import.meta.dirname;
```

build/render.js
```diff
@@ -1,23 +1,24 @@
...
 
 import path from "node:path";
-import { fileURLToPath } from "node:url";
 
-import { render as distRender } from "../dist/ssr/index.js";
+import { BUILD_OUT_ROOT } from "./env.js";
 
-const __dirname = path.dirname(fileURLToPath(import.meta.url));
+const { render: distRender } = /** @type {import("../entry.ssr.js")} */ (
+  await import(path.resolve(BUILD_OUT_ROOT, "static", "ssr", "index.js"))
```
caugner, line 12: Silly question: If `static/ssr/index.js` is in BUILD_OUT_ROOT, does this mean we publish this file?

caugner (approved):
Changes LGTM, just a couple of nits, and one silly question.

build/render.js
LeoMcA: Yes: it's true at the moment too, we don't need to, but there shouldn't be any harm in doing so. I'll file an issue to output this somewhere else.

build/env.js
LeoMcA: Looks stable to me, as of v22: https://nodejs.org/docs/latest-v22.x/api/esm.html#importmetadirname

Regardless, AFAIK an API being experimental in a node version refers to the fact that its interface might change: so if it becomes stable in a later version, and its interface hasn't changed, it's effectively stable in those earlier versions too.

build/render.js
LeoMcA: https://github.com/mdn/fred/issues/595

build/env.js
caugner: Technically, we require `>=20.0.0` in our `package.json`, so ESLint might complain about us using an API that was marked as stable only after that.

If ESLint is properly checking the `eslint-plugin-n` rules in this repo, then it should give us a `no-unsupported-features/node-builtins` error if we try and use it. However, in that case, a solution would be to bump our Node requirement to `>=22.16.0`.

LeoMcA: > Technically, we require >=20.0.0 in our package.json

Are you sure about that? :wink: 

https://github.com/mdn/fred/blob/main/package.json#L126

caugner: Apologies, yes, I meant `>=22.0.0`, and `import.meta.dirname` is still experimental in 22.0.0.

caugner (approved):

PR #597, opened 2025-08-20
fix(search-modal): handle modifier keys

### LINKED ISSUE #583: Search doesn't handle modifier keys
When searching, old frontend (Yari) [handles modifier keys correctly](https://github.com/mdn/yari/blob/5def85c0c123a6feb709c7944d146dd6bbed5ede/client/src/search.tsx#L476-L492), but new frontend (Fred) does not:

https://github.com/mdn/fred/blob/607b72fb86653d0ecc733b7810fedb560ce8c747/components/search-modal/element.js#L108-L114

I would like to be able to e.g. Meta+Enter to open search results in new tabs.

caugner:
Allows opening search results in new tabs.

### Description

Add support for opening quicksearch results with modifier keys.

### Motivation

Allows opening results in a new tab.

Fixes https://github.com/mdn/fred/issues/583.

### DIFF
--- a/components/search-modal/element.js
+++ b/components/search-modal/element.js
@@ -78,6 +78,27 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
         event.preventDefault();
         this._select(this._selected + 1);
         break;
+
+      case "Enter": {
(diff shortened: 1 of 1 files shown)


LeoMcA (approved):
nice!

PR #598, opened 2025-08-20
fix(sidebar-filter): increase contrast of matches

### LINKED ISSUE #572: Sidebar filter search results highlights should be more visible in dark mode

caugner:
### Description

Increase the contrast of the sidebar filter matches by changing the background color (both of the matches, and the match count) to that of matches in the search dialog.

### Motivation

Improve accessibility with the dark theme.

### Additional details

Test URL: https://fred-pr598.review.mdn.allizom.net/en-US/docs/Web/HTML

| Before | After |
|--------|--------|
| image | image |
| image | image |

Fixes https://github.com/mdn/fred/issues/572.

### DIFF
--- a/components/left-sidebar/server.css
+++ b/components/left-sidebar/server.css
@@ -126,7 +126,7 @@
 
   mark {
     /* Sidebar filter highlight */
-    color: unset;
+    color: var(--color-area-link);
     background-color: var(--color-background-blue);
   }
 }
--- a/components/left-sidebar/server.css
+++ b/components/left-sidebar/server.css
@@ -126,8 +126,8 @@
 
   mark {
     /* Sidebar filter highlight */
-    color: var(--color-area-link);
-    background-color: var(--color-background-blue);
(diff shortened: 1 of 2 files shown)


pepelsbey:
I would suggest using the same color for highlight as we already use in quick search. These are very similar use cases that would benefit from consistency.

image

caugner:
> I would suggest using the same color for highlight as we already use in quick search. These are very similar use cases that would benefit from consistency.

Good idea, and done in [b3d4014](https://github.com/mdn/fred/pull/598/commits/b3d40142b47c8764f3ceeff5a6c8f9e726e6dd8b).

argl (approved):

PR #620, opened 2025-08-22
feat(writer-mode): reload when content edited

### LINKED ISSUE #603: Autoreload when content edited
For when fred is being used in the content repo to preview changes.

We can poll the index.json and reload the page when it changes.

LeoMcA:
Fixes https://github.com/mdn/fred/issues/603

Uses [`document.visibilityState`](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState) to ensure we don't hammer rari when a user has multiple tabs open.

Uses two methods to avoid https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Too_much_recursion if recursively calling one.

### DIFF
--- a/components/writer-reload/element.js
+++ b/components/writer-reload/element.js
@@ -1,6 +1,14 @@
 import { LitElement, nothing } from "lit";
 
+/**
+ * Custom element which reloads the page when the document has changed.
+ */
 export default class MDNWriterReload extends LitElement {
+  constructor() {
+    super();
+    this._state = "";
+  }
+
   connectedCallback() {
     super.connectedCallback();
     this._pollForChanges();
@@ -33,7 +41,7 @@ export default class MDNWriterReload extends LitElement {
       if (document.visibilityState === "visible") {
         await this._reloadIfChanged();
       }
-      await wait(1000);
+      await timeout(1000);
     }
   }
 }
@@ -41,7 +49,7 @@ export default class MDNWriterReload extends LitElement {
 /**
  * @param {number} ms
  */
-async function wait(ms) {
+async function timeout(ms) {
   return new Promise((resolve) => setTimeout(resolve, ms));
 }
 
--- a/components/writer-reload/element.js
+++ b/components/writer-reload/element.js
@@ -1,0 +1,48 @@
+import { LitElement, nothing } from "lit";
+
+export default class MDNWriterReload extends LitElement {
+  connectedCallback() {
+    super.connectedCallback();
+    this._pollForChanges();
+  }
+
+  render() {
+    return nothing;
+  }
+
+  async _reloadIfChanged() {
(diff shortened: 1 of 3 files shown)


components/writer-reload/element.js
```diff
@@ -0,0 +1,48 @@
...
+  async _reloadIfChanged() {
+    const url = new URL(
+      "index.json",
+      new URL(location.pathname + "/", location.origin),
+    );
+    const res = await fetch(url);
+    if (res.ok) {
+      const state = await res.text();
+      if (!this._state) {
+        this._state = state;
```
caugner, line 30: Nit: Should `_state` not be declared?

components/writer-reload/element.js
```diff
@@ -0,0 +1,48 @@
...
+    } else {
+      console.error("Failed to fetch document", res.status, res.statusText);
+    }
+  }
+
+  async _pollForChanges() {
+    while (true) {
+      if (document.visibilityState === "visible") {
+        await this._reloadIfChanged();
+      }
```
caugner, line 43: Nit: Shouldn't we only poll for changes while the page is visible, i.e. use the [visibilitychange](https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event) to toggle polling on and off? Admittedly the performance impact is probably negligible.

components/writer-reload/element.js
```diff
@@ -0,0 +1,48 @@
+import { LitElement, nothing } from "lit";
+
+export default class MDNWriterReload extends LitElement {
```
caugner, line 6: Nit: Add JSDoc explaining its purpose.

components/writer-reload/element.js
```diff
@@ -0,0 +1,48 @@
...
+      console.error("Failed to fetch document", res.status, res.statusText);
+    }
+  }
+
+  async _pollForChanges() {
+    while (true) {
+      if (document.visibilityState === "visible") {
+        await this._reloadIfChanged();
+      }
+      await wait(1000);
```
caugner: Nit: Looks like `timeout` would be a better name?

```suggestion
      await timeout(1000);
```

caugner (approved):

components/writer-reload/element.js
LeoMcA: You don't like to await wait? :stuck_out_tongue:

caugner: Wait, await wait?

components/writer-reload/element.js
LeoMcA: Yeah, I can't imagine it's a performance problem, and the logic becomes significantly more complex: we have to add state to check if we're currently polling, good chance for race conditions, etc.

PR #623, opened 2025-08-22
fix: ensure css images respect user theme choice

### LINKED ISSUE #570: Some images don't respond to theme
We have a number of images which respect the browser theme, rather than the user's theme:

e.g. baseline icon in light theme when browser pref is dark theme:
Image

These can be found with a search for `prefers-color-scheme`.

They should instead use a conditional media query based on the `data-theme` attribute, [like in the curriculum](https://github.com/mdn/fred/blob/dfaa8ba338347cb4cad1e0dd721c0189c235004a/components/curriculum/vars.css#L217).

Ideally we'd come up with a mixin/helper function, to not have to do all this extra syntax each time.

LeoMcA:
Fixes https://github.com/mdn/fred/issues/570 with a potentially contentious change.

The following CSS is the minimal I could come up with to ensure that a user's theme preference is respected over the browser theme:
```css
property: light;

@media (prefers-color-scheme: dark) {
  property: dark;
}

[data-theme="light"] & {
  property: light;
}

[data-theme="dark"] & {
  property: dark;
}
```

I feel this is too verbose to be putting everywhere we want to do this, and liable to be forgotten the next time we need to do this, so in this PR I add the `postcss-mixins` plugin, and define a mixin which allows us to write the following, instead of the above:
```css
@mixin light-dark property, light, dark;
```

I propose that the benefits from the much simpler syntax here outweigh the negatives of using something which isn't CSS: much like our use of `@custom-media`. And, much like `@custom-media`, this is a feature which very well may be coming to CSS, in the form of [custom functions](https://www.w3.org/TR/css-mixins-1/) (though the syntax would be better).

@caugner @argl assigned you both, as I'd like unanimous support (or at least no strong objections!) for this change. @argl could you particularly look at whether I've missed anything with the light/dark logic, I tested each permutation of browser light/dark, user theme default/light/dark, but I want to make sure I didn't miss anything.


rspack.config.js
```diff
@@ -59,6 +59,32 @@ const postcssLoaders = (lit = false) => [
...
+            "postcss-mixins",
+            {
+              mixins: {
+                "light-dark":
+                  /**
+                   * @param {any} _mixin
+                   * @param {string} property
+                   * @param {string} light
+                   * @param {string} dark
+                   */
```
caugner, line 72: Nit: Add a short example of how this mixin is used.

caugner (approved):

argl (approved):
Tested locally, no adverse effects noticed, nice!

PR #625, opened 2025-08-22
fix(play-runner): internal state could clobber runner state

### LINKED ISSUE #606: Race condition on interactive examples
Discovered here https://github.com/mdn/fred/pull/549#discussion_r2290512516 but not actually caused by that PR.

The most effective way to reproduce for me seems to be, set a slight connection throttle (I use "Wi-Fi"), and refresh the page lots until the example doesn't load correctly, like so:

image

The DOM state when it fails reveals something interesting:
```html
<mdn-play-runner defaults="ix-choice" src="https://267dced9-eecc-4e9c-a2a9-044c3de917f0.mdnyalp.dev/runner.html?uuid=267dced9-eecc-4e9c-a2a9-044c3de917f0&amp;state=hVJBbsMgEPwKIpdUquM0lXpw0l76DS4U1rAtBgtw4ijK37vYjhq1hx6QVjO7M7sjLtzmzvGGHxKojMEz1K%2BCa2jl4HIFo%2Bx6B4K%2FCc%2FYQeORKSdTopaFqlTwWaKHuDRRG3ZmrtisdmsFBx34LPiNTVERXScrI%2BiKdCGnGjtpINXLUKpbjNCGsXLBhE06mp%2FxE%2BpsSWC33QrO6nnHmpak6lAvB1HNH7lKiY7c%2FFmaXcrQh1RfJobBa2JciA07WcywL9xk0rDdy7YfJ8ACGpvvEY2pd%2FLcsNbBjEiHxlck0aWGKboZ4oR%2FDilje578Cb3jrsILv%2FqV1LxdUW3Y0yTQS63Rm4Y9z%2BbX6brPchyF924DKlgL3qIj2YYN0a2F%2BDdiYnU4lXBXi3OFusw97OmRwfIfig2OlZpsCM6WmglzJRF%2B%2FQY%3D"></mdn-play-runner>
```

The `src` attribute has been updated: meaning the [`_updateSrc` task](https://github.com/mdn/fred/blob/aa37d1d4411413403ed386afbf4be4f5803ef5ca/components/play-runner/element.js#L97) has run successfully, but the iframe hasn't navigated to the new location.

We do some [safe navigation on that iframe in the task](https://github.com/mdn/fred/blob/aa37d1d4411413403ed386afbf4be4f5803ef5ca/components/play-runner/element.js#L121):
```js
this._iframe.value?.contentWindow?.location.replace(src);
```

It's possible that we're attempting to set that location before `contentWindow` exists, because of a slow load over the network. This would also explain why I've not been able to reproduce it locally: as the network is near-instant.

LeoMcA:
Fixes https://github.com/mdn/fred/issues/606 (though you might have to trust me on that...)

I never came up with a way of reproducing the above issue locally organically. I did however figure out that the issue isn't `this._iframe.value?.contentWindow?.location.replace(src);` as this works even before the blank runner has loaded, when the iframe's src is `about:blank`.

The one way I was able to trigger - what I assume is the same issue - locally, was by artificially adding a random query param to the iframe's src on every render. This produced the same result, DOM state and network panel state as the bug on prod.

This fix avoids that problem, and thoroughly cleans up the code. We had some rather complex code here because we wanted to avoid adding to the browser history when updating the iframe's src. We used the `location.replace` approach to do this before, but we can just ensure the iframe gets removed and re-added to the DOM with the Lit `keyed` directive.

This means we can also remove the complexity around ensuring we `location.replace` on a non-`about:blank` page in Safari, [to ensure it sends a `Referer` header](https://github.com/mdn/yari/commit/bfb93f509e7df80efb282ab918ac243b698aafb4). Since we're just inserting a new iframe, the Referer is always sent.

Also updates the local server to allow running the playground on https: this was necessary for testing against safari (on another machine) locally.

Tested across latest release version of Firefox, Chromium and Safari.

### DIFF
--- a/components/env/index.js
+++ b/components/env/index.js
@@ -9,6 +9,7 @@ export const PLAYGROUND_BASE_HOST = parseString(
   "PLAYGROUND_BASE_HOST",
   "mdnplay.dev",
 );
+export const PLAYGROUND_LOCAL = parseBool("PLAYGROUND_LOCAL", false);
 
 export const FXA_SIGNIN_URL = parseString(
   "FXA_SIGNIN_URL",
--- a/components/play-runner/element.js
+++ b/components/play-runner/element.js
@@ -1,10 +1,11 @@
 import { Task } from "@lit/task";
 import { LitElement, html } from "lit";
 import { ifDefined } from "lit/directives/if-defined.js";
+import { keyed } from "lit/directives/keyed.js";
 import { createRef, ref } from "lit/directives/ref.js";
 
 import { ThemeController } from "../color-theme/controller.js";
-import { PLAYGROUND_BASE_HOST } from "../env/index.js";
+import { PLAYGROUND_BASE_HOST, PLAYGROUND_LOCAL } from "../env/index.js";
 import { compressAndBase64Encode } from "../playground/utils.js";
 
 import styles from "./element.css?lit";
@@ -24,7 +25,7 @@ export class MDNPlayRunner extends LitElement {
     srcPrefix: { type: String, attribute: "src-prefix" },
     allow: { type: String },
     sandbox: { type: String },
-    src: { reflect: true },
+    _src: { state: true },
   };
 
   static styles = styles;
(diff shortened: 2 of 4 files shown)


argl (approved):
Nice improvement, simplify!

caugner (approved):

PR #641, opened 2025-08-27
fix(npm): js/styles/assets weren't being found

### LINKED ISSUE #594: Respect BUILD_OUT_ROOT in build/render.js
Split from https://github.com/mdn/fred/pull/592

The code there initially had `build/render.js` referencing `process.env.BUILD_OUT_ROOT`: this works for local development and the build in yari, but the tests caught that this doesn't work when packaging for npm: as the runtime variable is still used.

It needs to set to a relative path in the npm package, based on the build-time value of BUILD_OUT_ROOT.

LeoMcA:
Fixes https://github.com/mdn/fred/issues/594

Tested locally with `npm pack`

### DIFF
--- a/build/env.js
+++ b/build/env.js
@@ -5,5 +5,8 @@ const nodeModule = defaultOut.endsWith(
   path.join("node_modules", "@mdn", "fred", "out"),
 );
 
-export const BUILD_OUT_ROOT =
-  (nodeModule && defaultOut) || process.env.BUILD_OUT_ROOT || defaultOut;
+export const RARI_BUILD_ROOT = process.env.BUILD_OUT_ROOT || defaultOut;
+
+// When running from an npm package, this needs to be the default output directory,
+// so we can load the pre-built assets
+export const FRED_BUILD_ROOT = nodeModule ? defaultOut : RARI_BUILD_ROOT;
--- a/build/env.js
+++ b/build/env.js
@@ -1,3 +1,9 @@
 import path from "node:path";
+
+const defaultOut = path.join(import.meta.dirname, "..", "out");
+const nodeModule = defaultOut.endsWith(
+  path.join("node_modules", "@mdn", "fred", "out"),
+);
+
 export const BUILD_OUT_ROOT =
-  process.env.BUILD_OUT_ROOT || path.join(import.meta.dirname, "..", "out");
+  (nodeModule && defaultOut) || process.env.BUILD_OUT_ROOT || defaultOut;
--- a/build/render.js
+++ b/build/render.js
@@ -2,23 +2,23 @@ import { readFile } from "node:fs/promises";
 
 import path from "node:path";
 
-import { BUILD_OUT_ROOT } from "./env.js";
+import { FRED_BUILD_ROOT } from "./env.js";
 
 const { render: distRender } = /** @type {import("../entry.ssr.js")} */ (
-  await import(path.resolve(BUILD_OUT_ROOT, "static", "ssr", "index.js"))
+  await import(path.resolve(FRED_BUILD_ROOT, "static", "ssr", "index.js"))
 );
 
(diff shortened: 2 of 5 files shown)


build/env.js
```diff
@@ -1,3 +1,9 @@
 import path from "node:path";
+
+const defaultOut = path.join(import.meta.dirname, "..", "out");
+const nodeModule = defaultOut.endsWith(
+  path.join("node_modules", "@mdn", "fred", "out"),
+);
+
 export const BUILD_OUT_ROOT =
-  process.env.BUILD_OUT_ROOT || path.join(import.meta.dirname, "..", "out");
+  (nodeModule && defaultOut) || process.env.BUILD_OUT_ROOT || defaultOut;
```
caugner: Doesn't this mean it will always build into `node_modules/@mdn/fred/out` when run from the content repo, even if `BUILD_OUT_ROOT` is set? Equivalent to:

```suggestion
  nodeModule ? defaultOut : process.env.BUILD_OUT_ROOT || defaultOut;
```

Isn't the most important step here that we `path.resolve()` the given `process.env.BUILD_OUT_ROOT`?

```suggestion
  process.env.BUILD_OUT_ROOT ? path.resolve(process.env.BUILD_OUT_ROOT) : defaultOut;
```

caugner: Here's what I tested:

```patch
# node_modules/@mdn/fred/build/env.js
 import path from "node:path";
 export const BUILD_OUT_ROOT =
-  process.env.BUILD_OUT_ROOT || path.join(import.meta.dirname, "..", "out");
+  process.env.BUILD_OUT_ROOT ? path.resolve(process.env.BUILD_OUT_ROOT) : path.join(import.meta.dirname, "..", "out");
+console.log({ BUILD_OUT_ROOT})
```

Before, it gave:

```
{ BUILD_OUT_ROOT: 'build' }
```

With this change, it gave:

```
{ BUILD_OUT_ROOT: '/path/to/mdn/content/build' 
```

LeoMcA: And that's the wrong path, BUILD_OUT_ROOT shouldn't be set based on the environment variable when running from the npm package, it must be the path that BUILD_OUT_ROOT was set to when building the npm package i.e. `/path/to/mdn/content/node_modules/@mdn/fred/out`

LeoMcA: I'll split `BUILD_OUT_ROOT` into two variables internally, something like `RARI_BUILD_ROOT` and `FRED_BUILD_ROOT` to make this distinction clearer

caugner: To clarify, are we dropping support for BUILD_OUT_ROOT when running from the npm package?

LeoMcA: I'm not really sure what you mean: we never supported it. `BUILD_OUT_ROOT` sets where rspack outputs built js/css/assets, where fred looks for those files, and where our ssr script looks for rari files.

When running as an npm package, rspack has already built into `BUILT_OUT_ROOT`, and so once-packaged and installed the files are in `node_modules/@mdn/fred/out`. Fred needs to therefore load those files from that directory, the current value of `BUILD_OUT_ROOT` is inconsequential.

caugner: I'm referring to the `start`/`start:fred` scripts in content that set `BUILD_OUT_ROOT=build`:

https://github.com/mdn/content/blob/a2b29d9159294f1437e0adf49cdf3019e9c1c24b/package.json#L40-L41

But maybe those were just accidentally copied from the `build` script:

https://github.com/mdn/content/blame/a2b29d9159294f1437e0adf49cdf3019e9c1c24b/package.json#L23

caugner (approved):

build/env.js
LeoMcA: It's possible we still need to set it, otherwise rari complains: I'll check when I open my `content` PR to update the fred version (and add `FRED_WRITER_MODE=true`)

PR #642, opened 2025-08-27
feat(env): add runtime variables

### LINKED ISSUE #604: Writer's toolbar doesn't respect .env
Reported by @pepelsbey on slack:

> The “Open in editor” link didn’t work with my existing EDITOR=code config in the .env. There was an error in the server’s logs. Should it be updated?

LeoMcA:
Partially fixes https://github.com/mdn/fred/issues/604

Making `WRITER_MODE` a runtime variable ensures that if the toolbar is being shown in the client, then the `_editor` API is enabled.

Tested locally with `npm pack`.

### DIFF
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -40,7 +40,7 @@ jobs:
         if: steps.release.outputs.release_created
         run: npm publish --access public
         env:
-          FRED_ALLOW_RUNTIME: true
+          FRED_ALLOW_RUNTIME_ENV: true
           # TODO: probably want to disable this, but it doesn't seem to work in content at the moment:
           FRED_LEGACY: true
           NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -40,7 +40,7 @@ jobs:
         if: steps.release.outputs.release_created
         run: npm publish --access public
         env:
-          FRED_WRITER_MODE: true
+          FRED_ALLOW_RUNTIME: true
           # TODO: probably want to disable this, but it doesn't seem to work in content at the moment:
           FRED_LEGACY: true
           NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
--- a/.github/workflows/npm-publish.yml
+++ b/.github/workflows/npm-publish.yml
@@ -40,7 +40,7 @@ jobs:
         if: steps.release.outputs.release_created
         run: npm publish --access public
         env:
-          FRED_ALLOW_RUNTIME_ENV: true
+          FRED_RUNTIME_ENV: true
           # TODO: probably want to disable this, but it doesn't seem to work in content at the moment:
           FRED_LEGACY: true
           NODE_AUTH_TOKEN: ${{ secrets.NPM_AUTH_TOKEN }}
--- a/components/env/README.md
+++ b/components/env/README.md
@@ -1,0 +1,25 @@
+# Environment Variables
+
+If a component or other part of Fred requires a variable set from the environment, you should define it in the [`./index.js`](./index.js) file, with a safe default value.
+
+## Defaults
+
+We set safe defaults for prod, unless the risk from doing - and having this set everywhere, across local dev etc. - outweighs the risk of it not being set on prod.
+
+E.g. we don't set `GLEAN_ENABLED` to the prod default, as we don't want to send telemetry pings for prod from other environments, and we'll notice in our telemetry if it's not set in prod.
+
+## Naming
+
+Within Fred, variables are unprefixed, however the environment variable name is prefixed with `FRED_` to avoid conflicts with other environment variables. For example, if you define a variable `MY_VAR` in Fred, the environment variable to set would be `FRED_MY_VAR`.
+
+## Secrets
+
+Rspack will bundle all environment variables prefixed with `FRED_` into the bundle, which is exposed client side. Therefore you should never set secrets through `FRED_` environment variables.
+
+As secrets can only be used in server-side components, you should simply set them through a non-`FRED_`-prefixed environment variable, and access them using the `process.env` object.
+
+## Build-time vs runtime
+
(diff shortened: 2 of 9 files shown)


components/env/utils.js
```diff
@@ -0,0 +1,39 @@
+import { ALLOW_RUNTIME, runtimeVariables } from "./runtime.js";
+
+/**
+ * @param {string} name
+ * @param {boolean} fallback
+ * @param {boolean} runtime
+ */
+export function parseBool(name, fallback, runtime = false) {
```
caugner: Let's make this an options object, to avoid confusion (I was confused when I saw the `parseBool("WRITER_MODE", false, true);` call above).

```suggestion
export function parseBool(name, fallback, { runtime = false }) {
```

PS: Maybe `false` should also be the default value of `fallback`?

components/env/utils.js
```diff
@@ -0,0 +1,39 @@
...
+    return fallback;
+  }
+}
+
+/**
+ * @param {string} name
+ * @param {string} fallback
+ * @param {boolean} runtime
+ */
+export function parseString(name, fallback, runtime = false) {
```
caugner: dito

components/outer-layout/server.js
```diff
@@ -86,6 +93,11 @@ export class OuterLayout extends ServerComponent {
             content="width=device-width, initial-scale=1.0"
           />
           <title>${context.pageTitle || "MDN"}</title>
+          ${ALLOW_RUNTIME
+            ? unsafeHTML(`<script>process = {
+  env: ${JSON.stringify(env)}
+};</script>`)
```
caugner, line 99: Nit: Should this be setting `__MDNEnv` instead?

```suggestion
            ? unsafeHTML(`<script>globalThis.__MDNEnv = ${JSON.stringify(env)};</script>`)
```

components/env/runtime.js
```diff
@@ -0,0 +1,6 @@
+import { parseBool } from "./utils.js";
+
+/** @type {string[]} */
+export const runtimeVariables = [];
+/** Overriden to prod default (false) in rspack config, set to true so it works in dev server by default. */
+export const ALLOW_RUNTIME = parseBool("ALLOW_RUNTIME", true);
```
caugner: Nit: How likely would this expose anything sensitive, if it was accidentally enabled in prod?

components/env/utils.js
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/**
+ * @param {string} name
+ * @param {boolean} runtime
+ * @returns {string | undefined}
+ */
+function getEnv(name, runtime = false) {
+  name = `FRED_${name}`;
+  if (runtime && ALLOW_RUNTIME) {
```
caugner: Nit: Could we make this variable clearer?

(Rationale: "ALLOW_RUNTIME" sounds less like it could hurt to enable.)

```suggestion
  if (runtime && EXPOSE_FRED_ENV) {
```

caugner (changes requested):
Some nits, and one request.

PS: What does runtime, or runtime variables refer to here? Should this be explained somewhere, such as in the README.

components/env/utils.js
LeoMcA: > Maybe false should also be the default value of fallback?

No, this should always be set explicitly

components/outer-layout/server.js
LeoMcA: It can't set `__MDNEnv` because rspack has already overwritten all references to that in the js bundle with the `FRED_` env vars set at compile time.

Setting `process.env` also means we don't have to do any kind of special handling for server components or the `server.js` itself, which fetch runtime variables from `process.env` directly

components/env/runtime.js
LeoMcA: We already expose all `FRED_` variables in the client bundle

components/env/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Defaults
+
+We set safe defaults for prod, unless the risk from doing - and having this set everywhere, across local dev etc. - outweighs the risk of it not being set on prod.
+
+E.g. we don't set `GLEAN_ENABLED` to the prod default, as we don't want to send telemetry pings for prod from other environments, and we'll notice in our telemetry if it's not set in prod.
+
+## Naming
+
+Within Fred, variables are unprefixed, however the environment variable name is prefixed with `FRED_` to avoid conflicts with other environment variables. For example, if you define a variable `MY_VAR` in Fred, the environment variable to set would be `FRED_MY_VAR`.
```
caugner, line 13: (I like to use the full variable name everywhere, to make it easier to find usages.)

.github/workflows/npm-publish.yml
```diff
@@ -40,7 +40,7 @@ jobs:
         if: steps.release.outputs.release_created
         run: npm publish --access public
         env:
-          FRED_WRITER_MODE: true
+          FRED_ALLOW_RUNTIME_ENV: true
```
caugner: Nit: Make make it more concise?

```suggestion
          FRED_RUNTIME_ENVS: true
```

Or:

```suggestion
          FRED_RUNTIME_VARS: true
```

components/env/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Secrets
+
+Rspack will bundle all environment variables prefixed with `FRED_` into the bundle, which is exposed client side. Therefore you should never set secrets through `FRED_` environment variables.
+
+As secrets can only be used in server-side components, you should simply set them through a non-`FRED_`-prefixed environment variable, and access them using the `process.env` object.
```
caugner: Nit: Maybe make this a warning?


```suggestion
Rspack will bundle **all** environment variables prefixed with `FRED_` into the bundle, which is exposed client side.

> [!WARNING]
>  **Never set secrets** through `FRED_` environment variables.
>
> As secrets can only be used in server-side components, you should simply set them through a non-`FRED_`-prefixed environment variable, and access them using the `process.env` object.
```

caugner (approved):

components/env/README.md
LeoMcA: Too much of a change for this PR, but you make a good point... file an issue? :stuck_out_tongue:

PR #637, opened 2025-08-26
fix(live-samples): id can contain dots

LeoMcA:
fixes https://github.com/mdn/fred/issues/590

PR #652, opened 2025-08-28
fix(live-samples): avoid handling code examples twice

### LINKED ISSUE #651: Duplicated output in live samples
### MDN URL

https://developer.mozilla.org/en-US/docs/Learn_web_development

### What specific section or headline is this issue about?

Playground/Live runner in CSS guide section

I'm reading guides in Mdn learn with Chrome these days. The behavior of playground/live runner was normal yesterday, the code was not repeated. But that gets weird as the title said today. Is there an update ruining something? The bug can also be reproduced in Edge.
For example, at [CSS: overflowing content](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Overflow#css_tries_to_avoid_data_loss):
Image
And in the playground, the code is repeated twice:
Image
The bug can be found everywhere the playground/live runner exist, in mdn guides.

The code should not be repeated.

_No response_

_No response_

caugner:
### Description

Reverts https://github.com/mdn/fred/pull/637, and escapes the live id.

### Motivation

- Some code examples had both classes, so they were handled twice, and output appeared twice.
- Deduplication seemed like an option, but is overall more complex than just escaping.

### Additional details

Tested on the following pages:

- http://localhost:3000/en-US/docs/Learn_web_development/Core/Styling_basics/Overflow#css_tries_to_avoid_data_loss (from #651)
- http://localhost:3000/en-US/docs/Web/API/URL/parse_static#examples (from #590)

Fixes #651.

### DIFF
--- a/hooks/live-samples.js
+++ b/hooks/live-samples.js
@@ -11,8 +11,9 @@ for (const iframe of document.querySelectorAll("iframe[data-live-id]")) {
       /** @type {MDNCodeExample[]} */
       const codeExamples = [];
 
+      const escapedId = liveId.replaceAll(".", String.raw`\.`);
       for (const element of document.querySelectorAll(
-        `.live-sample___${liveId}, .live-sample---${liveId}`,
+        `.live-sample___${escapedId}, .live-sample---${escapedId}`,
       )) {
         const { MDNCodeExample, upgradePre } = await import(
           "../components/code-example/element.js"
--- a/hooks/live-samples.js
+++ b/hooks/live-samples.js
@@ -1,4 +1,3 @@
-/* eslint-disable unicorn/prefer-query-selector */
 /**
  * @import { MDNCodeExample } from "../components/code-example/element.js";
  */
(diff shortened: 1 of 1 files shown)


hooks/live-samples.js
```diff
@@ -12,12 +12,14 @@ for (const iframe of document.querySelectorAll("iframe[data-live-id]")) {
...
-      for (const element of [
+      // Using Set, as some code examples have both classes.
+      const elements = new Set([
         // use getElementsByClassName as liveId can contain dots,
         // which breaks querySelectorAll
         ...document.getElementsByClassName(`live-sample___${liveId}`),
         ...document.getElementsByClassName(`live-sample---${liveId}`),
-      ]) {
+      ]);
+      for (const element of elements.values()) {
```
caugner: @LeoMcA Maybe a better alternative would be to go back to `querySelectorAll()`, but escape the `liveId`?

LeoMcA: Yes, I was just thinking the same

caugner: cf714edad7be8c173341223edf461f23be36c2a0

LeoMcA (approved):

PR #656, opened 2025-08-28
feat(baseline): make open state sticky

### LINKED ISSUE #655: Make Baseline banner open state sticky
If the Baseline banner is opened it should stay open across all other pages. This is for users who are targeting a particular Baseline year.

LeoMcA:
Fixes https://github.com/mdn/fred/issues/655

Requires a yari change to update the CSP hash.

Tested across Firefox, Chromium and Gnome Web (WebKit).

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ We place this JS in `entry.inline.js`, and it's inlined on page load.
 Rspack also generates the necessary CSP hash when doing a prod build with `npm run build`.
 
 If this code is component-specific, it can be imported with `?source&csp=true` and used to set the value of `static inlineScript` in a Server Component.
-Remebmer to add an additional entry to the CSP hashes in yari when doing so.
+Remember to add an additional entry to the CSP hashes in yari when doing so.
 
 ### Layout
 
--- a/README.md
+++ b/README.md
@@ -51,6 +51,9 @@ We need to run some JS as soon as possible at page load, to avoid layout shifts 
 We place this JS in `entry.inline.js`, and it's inlined on page load.
 Rspack also generates the necessary CSP hash when doing a prod build with `npm run build`.
 
+If this code is component-specific, it can be imported with `?source&csp=true` and used to set the value of `static inlineScript` in a Server Component.
+Remebmer to add an additional entry to the CSP hashes in yari when doing so.
+
 ### Layout
 
 See [the layout README](./components/layout/README.md).
--- a/README.md
+++ b/README.md
@@ -51,8 +51,29 @@ We need to run some JS as soon as possible at page load, to avoid layout shifts 
 We place this JS in `entry.inline.js`, and it's inlined on page load.
 Rspack also generates the necessary CSP hash when doing a prod build with `npm run build`.
 
-If this code is component-specific, it can be imported with `?source&csp=true` and used to set the value of `static inlineScript` in a Server Component.
+If this code is component-specific, it can be [imported with `?source&csp=true`](#custom-imports) and used to set the value of `static inlineScript` in a Server Component.
 Remember to add an additional entry to the CSP hashes in yari when doing so.
+
+### Custom Imports
+
+We support a range of non-standard imports in our JavaScript. This includes:
+
+#### `?source`
+
+Imports the raw source of the file as a string.
+
+```js
+import text from "./some-file.txt?source";
+```
+
+#### `&csp=true`
+
+Logs a CSP hash for the source of the file during the production build.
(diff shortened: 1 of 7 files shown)


caugner:
> Requires a yari change to update the CSP hash.

Converting to draft, as we shouldn't accidentally merge this before adding the CSP hash in yari.

Can you open that yari PR? 🙏

LeoMcA:
yep, done: https://github.com/mdn/yari/pull/13443

components/baseline-indicator/hook.js
caugner, line 1: Should probably better wrap in `try...catch` like below.

entry.inline.js
```diff
@@ -12,3 +12,35 @@ try {
 } catch (error) {
   console.warn("Unable to set nop", error);
 }
+
+try {
+  if (localStorage.getItem("baseline-indicator") === "open") {
```
caugner: Nit: Not sure it makes sense to wrap everything, would rather just try-catch the localStorage access.

```js
function isOpen() {
  try {
    return localStorage.getItem("baseline-indicator") === "open";
	} catch {
    return false;
  }
}
```

caugner (changes requested):

caugner (changes requested):
Sorry, but I'm strongly against adding this amount of **inline** JavaScript for such a marginal feature.

Both theme and nop are essential to our site and their inline JavaScript is minimal, this one('s) isn't.

@LeoMcA Is there any way we can avoid the inflated inline JavaScript and the layout shift at the same time?

entry.inline.js
caugner, line 1: Rather than adding this here, could we add this inline script directly in `components/baseline-indicator/server.js` behind the `<details>`?

LeoMcA: Yes, I'm already working on a solution like this

caugner: Awesome, please re-request review once you have pushed the changes.

README.md
```diff
@@ -51,6 +51,9 @@ We need to run some JS as soon as possible at page load, to avoid layout shifts
 We place this JS in `entry.inline.js`, and it's inlined on page load.
 Rspack also generates the necessary CSP hash when doing a prod build with `npm run build`.
 
+If this code is component-specific, it can be imported with `?source&csp=true` and used to set the value of `static inlineScript` in a Server Component.
+Remebmer to add an additional entry to the CSP hashes in yari when doing so.
```
caugner: Nit: typo

```suggestion
Remember to add an additional entry to the CSP hashes in yari when doing so.
```

components/server/index.js
```diff
@@ -31,16 +35,20 @@ export class ServerComponent {
...
         componentsUsed.delete(this.name);
         componentsWithStylesInHead.delete(this.name);
       }
       return nothing;
     }
 
+    const inline = this.inlineScript;
+    const res = inline
+      ? html`${renderResult}${unsafeHTML(`<script>${inline}</script>`)}`
+      : renderResult;
```
caugner: Nit: maybe:

```suggestion
    const { inlineScript } = this;
    if (inlineScript) {
      renderResult = html`${renderResult}${unsafeHTML(`<script>${inlineScript}</script>`)}`;
    }
```

components/server/index.js
```diff
@@ -31,16 +35,20 @@ export class ServerComponent {
       componentsUsed.add("legacy");
     }
 
-    const res = new this().render(...args);
-
-    if (!res || res === nothing) {
+    const renderResult = new this().render(...args);
```
caugner: I think it's okay to use `let` in this case, and overwrite it below:

```suggestion
    let renderResult = new this().render(...args);
```

LeoMcA:
Yari PR to add CSP hash: https://github.com/mdn/yari/pull/13505

README.md
```diff
@@ -51,6 +51,9 @@ We need to run some JS as soon as possible at page load, to avoid layout shifts
 We place this JS in `entry.inline.js`, and it's inlined on page load.
 Rspack also generates the necessary CSP hash when doing a prod build with `npm run build`.
 
+If this code is component-specific, it can be imported with `?source&csp=true` and used to set the value of `static inlineScript` in a Server Component.
```
caugner: Are there upstream docs about the `source` and `csp` parameter that we could link here?

caugner (approved):

README.md
LeoMcA: This isn't an upstream feature, but something specific to our rspack config: I've added a minimal explanation of that to the README which can be expanded on in the future.

caugner: Cool, thanks.

caugner (approved):

Issue #665, opened 2025-08-29
Improve discoverability of "Run" action for Playground links


caugner:
From @jakearchibald via Slack:

> Is there a way to make MDN playground links auto-run? Right now it seems somewhat unclear that it hasn't already ran. It might be worth putting a big "run" button in the centre of the canvas for the first run.

caugner:
Shared MDN Playgrounds don't auto-run for security reasons, but this means the output is just blank, and there is no indication that the user needs to press the "Run" button.

We could highlight the "Run" button, and/or duplicate it in the center of the output area.

jakearchibald:
> and/or duplicate it in the center of the output area.

I think this is the right answer, as I feel the user's attention is drawn to the output area.

OnkarRuikar:
Instead of a second run button how about we ask users to click on the 'Run' button?

>  ![Image](https://github.com/user-attachments/assets/18ff15e2-c146-4748-806d-d9c84f8f2812)

We already prompt users to report suspicious playground code. So we can add one more message to ask  them to click on run button.

This can be easily handled by adding following 6 lines of code to the on going PR: https://github.com/mdn/fred/pull/669 

<details>
<summary>Small code to add the message</summary>

```diff
diff --git a/components/playground/element.js b/components/playground/element.js
index 994c31c..d1e2e86 100644
--- a/components/playground/element.js
+++ b/components/playground/element.js
@@ -70,6 +70,7 @@ export class MDNPlayground extends L10nMixin(LitElement) {
         this._autoRun = true;
         controller.runOnChange = true;
         this._storeSession();
+        this.requestUpdate();
       }
     }
   }
@@ -377,6 +378,11 @@ ${"```"}`,
                   @toast-click=${this._reportOpen}
                 ></mdn-toast>`
               : nothing}
+            ${this._gistId && !this._autoRun
+              ? html`<mdn-toast
+                  .toastMessage=${"Click 'Run' button to render the output."}
+                ></mdn-toast>`
+              : nothing}
             <mdn-play-runner></mdn-play-runner>
             <div class="playground__console">
               <div>${this.l10n`Console`}</div>
```

</details>

jakearchibald:
@OnkarRuikar that's ok, but given how unusual it is to look at a playground and not see the result, I feel this will be the likely first interaction on the page, so it need to be big and cover the canvas, and be easy to resolve.

Fwiw, most other playgrounds I've seen just run the code by default. https://jsbin.com has an "auto run" checkbox.

PR #666, opened 2025-08-29
fix(button): increase contrast

caugner:
### Description

Increases the contrast between the background color with/without hover, all while ensuring a visible hover effect.

### Motivation

Resolve accessibility issues.

### Additional details

As of 120d223b1dfbc394c4f56ff4ffae004f4fa11e70:

### WCAG contrast ratios

| State                  | Light (main → branch)                                                                                                                                                                              | Dark (main → branch)                                                                                                                                                                                |
| ---------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| YES default (no hover) | [3.81:1](https://webaim.org/resources/contrastchecker/?fcolor=089141&bcolor=F7F7F8) 🔴 → [**7.74:1**](https://webaim.org/resources/contrastchecker/?fcolor=065B29&bcolor=F7F7F8) 🟢                  | [10.49:1](https://webaim.org/resources/contrastchecker/?fcolor=70EBA3&bcolor=212426) 🟢                                                                                                              |
| YES hover              | [3.66:1](https://webaim.org/resources/contrastchecker/?fcolor=089141&bcolor=EAF6EF) 🔴 → [**7.44:1**](https://webaim.org/resources/contrastchecker/?fcolor=065B29&bcolor=EAF6EF) 🟢                  | [9.56:1](https://webaim.org/resources/contrastchecker/?fcolor=70EBA3&bcolor=0C311C) 🟢 → [**5.54:1**](https://webaim.org/resources/contrastchecker/?fcolor=70EBA3&bcolor=065B29) 🟠                   |
| NO default (no hover)  | [5.02:1](https://webaim.org/resources/contrastchecker/?fcolor=D31232&bcolor=F7F7F8) 🟠                                                                                                              | [6.35:1](https://webaim.org/resources/contrastchecker/?fcolor=FC7D92&bcolor=212426) 🟠                                                                                                               |
| NO hover               | [4.53:1](https://webaim.org/resources/contrastchecker/?fcolor=D31232&bcolor=FFE5EA) 🟠 → [**5.67:1**](https://webaim.org/resources/contrastchecker/?fcolor=B7102C&bcolor=FFE5EA) 🟠                  | [6.24:1](https://webaim.org/resources/contrastchecker/?fcolor=FC7D92&bcolor=4B0B16) 🟠                                                                                                               |

[...]

PR #669, opened 2025-08-31
fix(playground): set auto run on clear

### LINKED ISSUE #663: Clear in shared Playground doesn't clear properly
STR:

1. Go to https://developer.mozilla.org/en-US/play?id=rCfusnHbQaseoVnpt5%2BOn3b12l6jKOf8oDyb9smSb5ueTnvaFpg87%2FSZ4fWjiysCSJBwRrmdXiMXJnTQ
2. Press "Clear"
3. Type "Hello world" into the HTML field

Expectations:

1. After (2), the Playground should be fully cleared.
2. After (3), the Playground should show "Hello world" in the output.

Actual:

1. After (2), I still see the "Seeing something inappropriate?" message/button.
2. After (3), the Playground output remains empty until I press "Run" manually.

OnkarRuikar:
- Fix https://github.com/mdn/fred/issues/663

### Issues

The PR addresses the following issues in the playground:

1. In a freshly loaded playground, after clearing, it doesn't auto-run the code.

~2. After clearing, the report message(`Seeing something inappropriate?`) toast doesn't go away.~
~3. The report message doesn't go away after closing the report dialog.~
~4. There is no close/dismiss button on the report toast/banner.~

### Solution

- For the first issue, when the clear button is clicked, set the `_autoRun` and `runOnChange` variables.
~- For the second issue, use `${this._gistId && this._showReportToast` condition and hide the toast message on clear button click.~
~- For the third and fourth issues, define a new toast element named MDNToast to show dismissable messages. [Toasts are lightweight, dismissible notifications](https://getbootstrap.com/docs/5.0/components/toasts/).~

Fix #663

### DIFF
--- a/components/playground/element.js
+++ b/components/playground/element.js
@@ -80,6 +80,7 @@ export class MDNPlayground extends L10nMixin(LitElement) {
       controller
     ) {
       controller.clear();
+      this._autoRun = true;
       this._storeSession();
       this.requestUpdate();
       const urlWithoutSearch = new URL(location.href);
--- a/components/play-controller/element.js
+++ b/components/play-controller/element.js
@@ -102,6 +102,7 @@ export class MDNPlayController extends LitElement {
   }
 
   clear() {
+    this.runOnChange = true;
     this.initialCode = undefined;
     this.srcPrefix = "";
     this.reset();


OnkarRuikar:
:warning: we can fix https://github.com/mdn/fred/issues/665 in this PR itself!

components/button/server.css
```diff
@@ -95,7 +95,7 @@
 
   &[data-variant="plain"] {
     &:hover {
-      background-color: light-dark(var(--color-gray-80), var(--color-gray-20));
+      background-color: light-dark(var(--color-gray-80), var(--color-gray-40));
```
caugner: What is the reason for this change?

components/toast/element.js
```diff
@@ -0,0 +1,52 @@
...
+  close() {
+    this.remove();
+  }
+
+  render() {
+    return html`
+      <div class="toast-container">
+        <div class="toast-message" @click=${this.click}>
+          ${this.toastMessage}
+        </div>
```
caugner: Clickable divs are an antipattern.

caugner:
@OnkarRuikar Thank you for taking the time to work on these changes.

For all the issues that you're addressing in this PR, and that don't have an issue yet, can you please create one? I think they should first go through triage, before we can accept fixes.

That said, if you could extract the fixes for https://github.com/mdn/fred/issues/663 into a separate PR, that would be great to land that fix earlier.

components/playground/element.js
```diff
@@ -79,7 +84,10 @@ export class MDNPlayground extends L10nMixin(LitElement) {
       confirm(this.l10n`Do you really want to clear everything?`) &&
       controller
     ) {
+      this._reportToastHide();
       controller.clear();
+      this._autoRun = true;
+      controller.runOnChange = true;
```
caugner: This should probably happen as part of `controller.clear()`.

```suggestion
```

components/playground/element.js
```diff
@@ -355,10 +371,11 @@ ${"```"}`,
...
           <section class="playground__runner-console">
-            ${this._gistId
-              ? html`<mdn-button @click=${this._reportOpen} variant="plain">
-                  Seeing something inappropriate?
-                </mdn-button>`
+            ${this._gistId && this._showReportToast
+              ? html`<mdn-toast
+                  .toastMessage=${"Seeing something inappropriate?"}
+                  @toast-click=${this._reportOpen}
+                ></mdn-toast>`
```
caugner: Overall, I think the added complexity isn't justified here for being able to hide that message.

Instead, maybe that message could be less prominent, and be more clearly identifiable as a button. But that's better discussed as part of an issue.

components/toast/element.js
```diff
@@ -0,0 +1,52 @@
+import { LitElement, html } from "lit";
+
+import exitIcon from "../icon/cancel.svg?lit";
+
+import "../button/element.js";
+import styles from "./element.css?lit";
+
+export class MDNToast extends LitElement {
```
caugner: We want to avoid creating components for single use cases.

components/playground/element.js
```diff
@@ -254,7 +262,15 @@ ${"```"}`,
...
 
+  _reportToastHide() {
+    if (this._showReportToast == true) {
+      this._showReportToast = false;
+      this.requestUpdate();
+    }
+  }
+
   _reportOpen() {
+    this._reportToastHide();
```
caugner: I could imagine hiding the report button when a report was submitted, but not necessarily already when it was clicked once.

caugner (commented):
Skimmed over the changes, and left some comments here and there.

OnkarRuikar:
> For all the issues that you're addressing in this PR, and that don't have an issue yet, can you please create one? I think they should first go through triage, before we can accept fixes.

Created separate issues:
- https://github.com/mdn/fred/issues/709
- https://github.com/mdn/fred/issues/710
- https://github.com/mdn/fred/issues/711

> That said, if you could extract the fixes for https://github.com/mdn/fred/issues/663 into a separate PR, that would be great to land that fix earlier.

Let's fix #663 here itself and rest of the stuff in separate PRs after triages there.

components/button/server.css
OnkarRuikar: It is being handled in https://github.com/mdn/fred/pull/666#issuecomment-3242258117 . No not required anymore.

components/toast/element.js
OnkarRuikar: This is not part of this PR anymore.

components/playground/element.js
OnkarRuikar: This is not part of this PR anymore.

components/toast/element.js
OnkarRuikar: This is not part of this PR anymore.

components/playground/element.js
OnkarRuikar: Let's discuss this in its own issue: https://github.com/mdn/fred/issues/711

components/playground/element.js
OnkarRuikar: done

OnkarRuikar: done

caugner (approved):

PR #672, opened 2025-08-31
fix(sidebar): only in English indicator wrapping and alignment

### LINKED ISSUE #622: Sidebar entries wrap weirdly with en-US indicator

On https://developer.mozilla.org/es/docs/MDN/Community/Our_repositories, the "(inglés)" indicators are not positioned ideally for sidebar entries that wrap:

Image

I would expect it more to look like this:

Image

This is caused by `.left-sidebar a { display: inline-flex; }`, but that might have some purpose.

_No response_

OnkarRuikar:
- fix https://github.com/mdn/fred/issues/622

### Description

Refer to the following screenshot of [our  repositories page](https://developer.mozilla.org/es/docs/MDN/Community/Our_repositories):

There are two issues:
1. The `(inglés)` indicator is not justified on the right if the sidebar item text is short.
2. The indicator is wrapped.

## Solutions

1. Make sure the sidebar entry gets 100% width and the inline-flex items are space-between justified.
2. Add nowrap to the indicator text.

### Additional details

After fix:


components/only-in-en-us/global.css
```diff
@@ -7,6 +7,8 @@ html a.only-in-en-us::after {
 
   vertical-align: super;
 
+  white-space: nowrap;
```
caugner: Yeah, this is probably a good idea.

components/left-sidebar/server.css
```diff
@@ -88,7 +88,13 @@
   a,
   span:not(.highlight-container) {
     display: inline-flex;
+
+    justify-content: space-between;
+
+    width: 100%;
```
caugner: This is an interesting alternative to showing the en-US indicator inline (at the end of the text, like in content).

For example, here's how it would impact [/pt-BR/docs/Web/CSS/Layout_cookbook](https://developer.mozilla.org/pt-BR/docs/Web/CSS/Layout_cookbook):

| Before | After |
|--------|--------|
| image | image | 

I'll leave it to @mdn/content-team to make the call here.

LeoMcA (changes requested):
The sidebar css is fraught with danger, we really need to fix the markup itself at some point.

This change causes icons to always wrap:

| Before | After |
| - | - |
| Screenshot 2025-09-01 at 14-16-25 Background Fetch API | Screenshot 2025-09-01 at 14-20-46 Background Fetch API |

OnkarRuikar:
> The sidebar css is fraught with danger, we really need to fix the markup itself at some point.
> This change causes icons to always wrap:

:( At this point we can do only `nowrap` :
> 



Other better arrangement 
```
SomeAPI🗑️⚠️      (ingles)
SomeAPI          (ingles)
SomeAPI🗑️        (ingles)
```
 requires changes at multiple places where `sidebarHTML` is generated.

LeoMcA:
I think we can also change `display: inline-flex;` to `display: inline-block;` here:

https://github.com/mdn/fred/blob/9c1e59a270e18327a895a6d14fce9c4778753f03/components/left-sidebar/server.css#L88-L90

Which makes wrapping more in-line with the icons:
image

caugner:
> I think we can also change `display: inline-flex;` to `display: inline-block;` here:
> 
> https://github.com/mdn/fred/blob/9c1e59a270e18327a895a6d14fce9c4778753f03/components/left-sidebar/server.css#L88-L90
> 
> Which makes wrapping more in-line with the icons:

Indeed, what Leo suggests seems to do the trick with a single line change:

```patch
 a, 
 span:not(.highlight-container) {
-  display: inline-flex;
+  display: inline-block;
```

OnkarRuikar:
>> I think we can also change `display: inline-flex;` to `display: inline-block;` here:
>> Which makes wrapping more in-line with the icons:
>
> Indeed, what Leo suggests seems to do the trick with a single line change:

The `display: inline-block;` puts the anchored text and the status icons on separate lines:
```css
.left-sidebar a, .left-sidebar span:not(.highlight-container) {
    display: inline-block;
 }
```

https://github.com/user-attachments/assets/e0e2a95b-ecec-4f46-b780-bcf457f392ac

As shown in the video, `display: inline` (the default value) puts everything on the same line. That is as expected, right?

LeoMcA:
> As shown in the video, display: inline (the default value) puts everything on the same line. That is as expected, right?

Yes, I think you're onto something here.

There's a slight shift to the padding around the sidebar elements with the `inline-block` -> `inline` switch which I don't entirely understand (could it be to do with margins collapsing differently?):

| `inline-block` | `inline` |
| - | - |
| image | image |

If we can fix this, and test across a range of different sidebars, this might go a long way to fix some of our sidebar woes.

OnkarRuikar:
>  our sidebar woes.

Yes it like playing minesweeper :(
minesweeper-nostalgia

> There's a slight shift to the padding around the sidebar elements with the `inline-block` -> `inline` switch which I don't entirely understand (could it be to do with margins collapsing differently?):

No, it's how the height is derived. In case of `inline`, the `height` is set to `auto` (which is font size). In case of `inline-block`, the `height` is set to `line-height`(1.5 of font-size in our case). As we make the `<a>` and `<span>` inline, the `padding-block: 0.25rem` doesn't  work. Lets move  the padding to the parent `<li>` element. I've [committed](https://github.com/mdn/fred/pull/672/commits/99c1a9db2d8f786c36d7a592c9381282cbb5b265) the changes.

<details>
<summary>Some sidebars after the padding change</summary>









</details>

LeoMcA:
> Yes it like playing minesweeper :(

I think it's more like whack-a-mole, no? Fix one problem, and another pops up!

> No, it's how the height is derived. In case of inline, the height is set to auto (which is font size). In case of inline-block, the height is set to line-height(1.5 of font-size in our case). As we make the `<a>` and `<span>` inline, the padding-block: 0.25rem doesn't work. Lets move the padding to the parent `<li>` element.

A-ha, I see. This is good, but one issue I see is it's made our `<summary>` and `<em>`'s a bit shorter, which affects the hover state of the first and the normal state of the second:

| before | after |
| - | - |
| Screenshot 2025-09-16 at 10-06-11 -moz-image-rect - CSS MDN | Screenshot 2025-09-16 at 10-06-07 -moz-image-rect - CSS MDN |

I think we can just apply the same `padding-block` to our `<summary>` and `<em>`s?

Also I think the `padding-block` on the `li` should be moved up a couple of lines so it also applies to `:last-child`

OnkarRuikar:
I've applied the padding-block to `em`, `strong`, and ` li` elements.

<table>
<tr>
 <td>before fix</td>
 <td>after fix</td>
</tr>
<tr>
 <td></td>
 <td> </td>
</tr>
</table>

caugner:
> I think we can just apply the same `padding-block` to our `<summary>` and `<em>`s?

I don't think it's that easy: Afaik the padding is currently on the link, which is intentional to increase the clickable area.

Maybe we should migrate away from `<em>` in rari, although we might need still need to put the link and its icons together in a container.

LeoMcA:
> I don't think it's that easy: Afaik the padding is currently on the link, which is intentional to increase the clickable area.

:sob: my sidebar nightmares back from originally implementing this in fred are going to come back

> Maybe we should migrate away from `<em>` in rari, although we might need still need to put the link and its icons together in a container.

IMO the icons should be inside the `<a>` - I think it would solve most of the problems here, and I'm not sure there's any reason to *not* allow clicking on the icon to click on the sidebar element.

Since we can make backwards incompatible changes now fred is out, is there any reason to continue generating the sidebar HTML in rari, or could we just output structured data and turn it into HTML in fred? It would tighten this sidebar structure -> HTML structure -> CSS loop and make improving/fixing all these issues a lot easier

caugner:
> IMO the icons should be inside the `<a>` - I think it would solve most of the problems here, and I'm not sure there's any reason to _not_ allow clicking on the icon to click on the sidebar element.

Icons inside the link would mean that the en-US indicator would show after the icons.

caugner:
We all appreciate the effort that went into this PR, but it looks very much like we cannot resolve this issue in Fred without causing other issues, so this needs to wait until we change the sidebar structure upstream in Rari, which will require some planning and coordination.

Closing this PR for now.

PR #673, opened 2025-08-31
chore(copy-button): add checkmark to copied indicator

### LINKED ISSUE #612: Copy button could have a check mark when copied
From @yashrajbharti  https://github.com/mdn/fred/issues/610

> Nit: copy button could have a check mark when copied? Not totally needed but just a personal preference as icon + text combination has a better UX
> 
> Image

OnkarRuikar:
- fix https://github.com/mdn/fred/issues/612

### Description

To make it more apparent, it is suggested to add a check mark along with `Copied` text. The PR uses a heavy check mark emoji.

This is how it looks after the change:

https://github.com/user-attachments/assets/364d9142-de64-4517-b3ee-be81148d2989

### DIFF
--- a/components/copy-button/element.js
+++ b/components/copy-button/element.js
@@ -2,6 +2,7 @@ import { LitElement, html } from "lit";
 
 import { L10nMixin } from "../../l10n/mixin.js";
 import { MDNButton } from "../button/element.js";
+import check from "../icon/check.svg?lit";
 
 export class MDNCopyButton extends L10nMixin(LitElement) {
   static properties = {
@@ -12,6 +13,7 @@ export class MDNCopyButton extends L10nMixin(LitElement) {
   constructor() {
     super();
     this.variant = "primary";
+    this.copiedSuccessfully = false;
     /** @type {Element | undefined} */
     this.copiesFrom = undefined;
   }
@@ -21,7 +23,7 @@ export class MDNCopyButton extends L10nMixin(LitElement) {
    */
   async _copy({ target }) {
     if (target instanceof MDNButton) {
-      let copiedSuccessfully = true;
+      this.copiedSuccessfully = true;
       try {
         const text = this.copiesFrom?.textContent?.trim();
         if (text) {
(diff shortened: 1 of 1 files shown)


components/copy-button/element.js
```diff
@@ -36,7 +36,7 @@ export class MDNCopyButton extends L10nMixin(LitElement) {
       }
 
       this._message = copiedSuccessfully
-        ? this.l10n`Copied`
+        ? this.l10n`Copied ✔️`
```
caugner: We would probably want to use a Lucide icon instead: https://lucide.dev/icons/check

Cf. other `mdn-button`s with icons.

OnkarRuikar: Done. See the OP for the demo.

components/copy-button/element.js
```diff
@@ -12,6 +13,7 @@ export class MDNCopyButton extends L10nMixin(LitElement) {
   constructor() {
     super();
     this.variant = "primary";
+    this.copiedSuccessfully = false;
```
LeoMcA: Our convention would be that as a private property/state in the component, this should be `_copiedSuccessfully`.

Then it should also be added to the `static properties` object so Lit knows to re-render if its value updates, like `_copiedSuccessfully: { state: true }`. It still works at the moment because the `_message` state changes at the same time, triggering an update: but we should guard against future state bugs.

LeoMcA (commented):
Just one small change required, then I think we're good to go - nice stuff!

components/copy-button/element.js
OnkarRuikar: done

Rumyra:
I approve the UI changes ✅

components/copy-button/element.js
```diff
@@ -21,7 +24,7 @@ export class MDNCopyButton extends L10nMixin(LitElement) {
    */
   async _copy({ target }) {
     if (target instanceof MDNButton) {
-      let copiedSuccessfully = true;
+      this._copiedSuccessfully = true;
```
LeoMcA: One tiny change: I noticed while testing what I thought was a weird flash when clicking, recording a profile showed it was the check appearing momentarily before the text changed to "copied", because we set it to true before the async await (and it renders in that time).

Some very low resolution screenshots from the profiler which just about show this:

image
image
image

If we move this line down to L31 after the await, so it reads:
```
        if (text) {
          await navigator.clipboard.writeText(text);
          this._copiedSuccessfully = true;
        }
```
We fix that problem.

LeoMcA (changes requested):
Apologies, caught one more change, but otherwise looks good!

caugner (approved):

components/copy-button/element.js
OnkarRuikar: good catch! :eye:

LeoMcA (approved):
Lovely stuff, thank you!

PR #674, opened 2025-08-31
fix(play-console): preserve whitespace

### LINKED ISSUE #668: Interactive example collapses whitespace in output
### MDN URL

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim

### What specific section or headline is this issue about?

Try it box

`console.log("    a      a      ")`
`>"a a"`

`>a      a` 

Same link

> The trim() method of String values removes whitespace from **both ends of this string** and returns a new string, without modifying the original string.

_No response_

### MDN metadata

<details>
<summary>Page report details</summary>

* Folder: `en-us/web/javascript/reference/global_objects/string/trim`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/global_objects/string/trim/index.md
* Last commit: https://github.com/mdn/content/commit/544b843570cb08d1474cfc5ec03ffb9f4edc0166
* Document last modified: 2025-07-10T09:07:55.000Z

</details>

OnkarRuikar:
- fix https://github.com/mdn/fred/issues/668

## Description

The output lines are displayed only in `<code>` blocks, so the HTML rendering is collapsing the whitespaces. As the content is only the text log strings, using only inline `<pre>` blocks solves the problem.

## Additional details

Code used to test:
```js
const greeting = "   ad   a       bd  a   ";
console.log(greeting);
console.log(greeting);
```
### Before fix

> 

> 

### After fix

> 

> 

Fixes #668


components/play-console/element.css
```diff
@@ -29,7 +29,12 @@ li {
   }
 }
 
-code {
+pre {
+  display: inline-block;
+
+  margin: 0;
```
caugner, line 44: Unfortunately, this doesn't look good with messages spanning multiple lines:

image

To avoid this, we could set:

```css
  display: inline;
  white-space: preserve;
```

image

And if we want to keep the indent for subsequent lines, we also need:

```css
li {
  display: flex;
  gap: 0.5ch;
}
```

OnkarRuikar: I tried solution with `li::marker` but Safari hasn't implemented the marker with content.
We don't need  flex if we  use `vertical-align` on ::before. Check the latest commit.

components/play-console/element.css
```diff
@@ -25,11 +25,17 @@ li {
   padding: 0 0.5em;
 
   &::before {
+    vertical-align: top;
     content: "> ";
   }
 }
 
-code {
+pre {
```
caugner: The change causes long lines to no longer break:

| Where | Screenshot |
|--------|--------|
| Local | image |
| Prod | image | 

Maybe we should use neither `code` nor `pre`, and just set `white-space: preserve`?

caugner (changes requested):

components/play-console/element.css
OnkarRuikar: Just `white-space: preserve` isn't cutting it. The `>` and `####...` are getting rendered on different lines.
We need `<li>` to not wrap and the text content to wrap. Let me know how the [latest commit](https://github.com/mdn/fred/pull/674/commits/9dfc2638a412f918a240a3b442bc3d3d4d5785f7) fares:
> 

Code used
```js
console.log("   abcd  d   c  ");
console.log("   abcd  \nd   c  ");
console.log("a".repeat(300));
```

components/play-console/element.css
```diff
@@ -23,13 +23,23 @@ ul {
...
   &::before {
+    vertical-align: top;
     content: "> ";
   }
 }
 
 code {
+  display: inline-block;
+
+  width: 98%;
```
caugner:
```suggestion
  width: calc(100% - 0.75rem);
```

components/play-console/element.css
```diff
@@ -23,13 +23,23 @@ ul {
 
 li {
   padding: 0 0.5em;
+  white-space: nowrap;
 
   &::before {
+    vertical-align: top;
     content: "> ";
```
caugner, line 35:
```suggestion
    display: inline-block;

    width: 0.75rem;

    vertical-align: top;

    content: "> ";
```

caugner (changes requested):
Let's use a fixed width for the line indicator, and reuse that width for calculating the `code` with.

caugner (approved):

Issue #667, opened 2025-08-29
Migrate content repo to use fred by default


caugner:
From @Josh-Cena via Discord:

> when would the deploy preview start using fred?

---

The content repo still uses yari:

- when running `yarn start`
- when deploying to the review environment

We should fix that.

PR #679, opened 2025-09-01
feat(server): allow specifiying port from env var

LeoMcA:
### Description

Allow specifying ports for server from environment variable

### Motivation

Good practice, allows running multiple fred servers at once. Necessary for content to maintain 5042 port.

### Additional details

Also fixes a bug where a runtime variable would fallback to a double prefixed variable, e.g. `FRED_FOO` would fall back to `FRED_FRED_FOO`.
Also modifies the server output to not output a link to the playground server: you wouldn't want to click on it, so it's potentially confusing.

Part of https://github.com/mdn/fred/issues/667.

### DIFF
--- a/components/env/index.js
+++ b/components/env/index.js
@@ -4,7 +4,13 @@ export const PLAYGROUND_BASE_HOST = parseString(
   "PLAYGROUND_BASE_HOST",
   "mdnplay.dev",
 );
-export const PLAYGROUND_LOCAL = parseBool("PLAYGROUND_LOCAL", false);
+export const PLAYGROUND_LOCAL = parseBool("PLAYGROUND_LOCAL", false, {
+  runtime: true,
+});
+export const PORT = parseString("PORT", "3000", { runtime: true });
+export const PLAY_PORT = parseString("PLAY_PORT", "3001", {
+  runtime: true,
+});
 
 export const FXA_SIGNIN_URL = parseString(
   "FXA_SIGNIN_URL",
--- a/components/env/index.js
+++ b/components/env/index.js
@@ -1,4 +1,4 @@
-import { parseBool, parseString } from "./utils.js";
+import { parseBool, parseInt, parseString } from "./utils.js";
 
 export const PLAYGROUND_BASE_HOST = parseString(
   "PLAYGROUND_BASE_HOST",
@@ -7,8 +7,8 @@ export const PLAYGROUND_BASE_HOST = parseString(
 export const PLAYGROUND_LOCAL = parseBool("PLAYGROUND_LOCAL", false, {
   runtime: true,
 });
-export const PORT = parseString("PORT", "3000", { runtime: true });
-export const PLAY_PORT = parseString("PLAY_PORT", "3001", {
+export const PORT = parseInt("PORT", 3000, { runtime: true });
+export const PLAYGROUND_PORT = parseInt("PLAYGROUND_PORT", 3001, {
   runtime: true,
 });
 
--- a/components/env/utils.js
+++ b/components/env/utils.js
@@ -22,6 +22,17 @@ export function parseBool(name, fallback, options) {
 
 /**
  * @param {string} name
+ * @param {number} fallback
+ * @param {Options} [options]
+ */
+export function parseInt(name, fallback, options) {
+  const stringValue = getEnv(name, options);
+  const numberValue = stringValue ? Number.parseInt(stringValue, 10) : fallback;
+  return Number.isNaN(numberValue) ? fallback : numberValue;
+}
+
+/**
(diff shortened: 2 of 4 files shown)


components/env/index.js
```diff
@@ -4,7 +4,13 @@ export const PLAYGROUND_BASE_HOST = parseString(
   "PLAYGROUND_BASE_HOST",
   "mdnplay.dev",
 );
-export const PLAYGROUND_LOCAL = parseBool("PLAYGROUND_LOCAL", false);
+export const PLAYGROUND_LOCAL = parseBool("PLAYGROUND_LOCAL", false, {
+  runtime: true,
+});
+export const PORT = parseString("PORT", "3000", { runtime: true });
+export const PLAY_PORT = parseString("PLAY_PORT", "3001", {
```
caugner: Nit: Consistency.

```suggestion
export const PLAYGROUND_PORT = parseString("PLAY_PORT", "3001", {
```

Edit: Ideally, we would parse these as integers (eg. `parseInteger`).

components/env/utils.js
```diff
@@ -36,10 +36,10 @@ export function parseString(name, fallback, options) {
  */
 function getEnv(name, options = {}) {
   const { runtime } = { runtime: false, ...options };
-  name = `FRED_${name}`;
+  const envName = `FRED_${name}`;
```
caugner: Nit: Could reduce confusion in line 42, where `envName` is next to `getEnv(name)`.


```suggestion
  const fullName = `FRED_${name}`;
```

Alternatively, `name` could be `suffix`.

```suggestion
  const envName = `FRED_${suffix}`;
```

server.js
```diff
@@ -322,14 +322,14 @@ export async function startServer() {
...
     console.log(
-      `Server started at ${http2 ? "https" : "http"}://localhost:3000`,
+      `Server started at ${http2 ? "https" : "http"}://localhost:${PORT}`,
     );
   });
 
-  const playServer = play.listen(3001, () => {
-    console.log(`Play server started at http://localhost:3001`);
+  const playServer = play.listen(PLAY_PORT, () => {
+    console.log(`Playground server started on port ${PLAY_PORT}`);
```
caugner: Nit: Could make clearer that this is not the Playground front end?

```suggestion
    console.log(`Playground back end started on port ${PLAY_PORT}`);
```

caugner (approved):

caugner:
> Also fixes a bug where a runtime variable would fallback to a double prefixed variable, e.g. `FRED_FOO` would fall back to `FRED_FRED_FOO`.
> Also modifies the server output to not output a link to the playground server: you wouldn't want to click on it, so it's potentially confusing.

FYI If you add a line `fix(env): ...` in the commit message (outside of a list), it will appear in the release notes.

LeoMcA:
> FYI If you add a line fix(env): ... in the commit message (outside of a list), it will appear in the release notes.

Oh that's cool, good to know: apologies for clobbering a few of those from your PRs

components/play-runner/element.js
```diff
@@ -101,7 +106,7 @@ export class MDNPlayRunner extends LitElement {
       const url = new URL(
         `${prefix}/runner.html`,
         PLAYGROUND_LOCAL
-          ? location.origin.replace("3000", "3001")
+          ? location.origin.replace(PORT.toString(), PLAYGROUND_PORT.toString())
```
caugner, line 109: Nit: Generally, I would tend to use `String(var)`, as it also works if var is accidentally undefined.

```suggestion
          ? location.origin.replace(String(PORT), String(PLAYGROUND_PORT))
```

Or more shorter:

```suggestion
          ? location.origin.replace(`${PORT}`, `${PLAYGROUND_PORT}`)
```

server.js
```diff
@@ -322,14 +322,14 @@ export async function startServer() {
...
     console.log(
-      `Server started at ${http2 ? "https" : "http"}://localhost:3000`,
+      `Server started at ${http2 ? "https" : "http"}://localhost:${PORT}`,
     );
   });
 
-  const playServer = play.listen(3001, () => {
-    console.log(`Play server started at http://localhost:3001`);
+  const playServer = play.listen(PLAYGROUND_PORT, () => {
+    console.log(`Playground backend started on port ${PLAYGROUND_PORT}`);
```
caugner, line 332: (FWIW, I personally prefer the writing "backend" and "frontend". The [Curriculum](https://developer.mozilla.org/en-US/curriculum/) uses "front-end", and our [launch blog post](https://developer.mozilla.org/en-US/blog/launching-new-front-end/) uses "front end". 🤷 )

caugner (approved):

PR #687, opened 2025-09-02
chore(compat-table): Port release label fix from yari

dipikabh:
### Description

Porting the change in https://github.com/mdn/yari/pull/13060 from yari to Fred

### Motivation

To make the label in the BCD table time-agnostic


LeoMcA (approved):
thank you!

PR #691, opened 2025-09-03
fix(env): remove inline script if no runtime variables are set

LeoMcA:
https://github.com/mdn/content/pull/40961/files#r2314389165

Making FRED_RUNTIME_ENV itself a runtime variable got a bit too... weird for my liking, so this is an alternative fix which should avoid errors if FRED_RUNTIME_ENV was set at build, but no variables were set at runtime.

### DIFF
--- a/components/env/README.md
+++ b/components/env/README.md
@@ -26,3 +26,5 @@ Rspack will bundle **all** environment variables prefixed with `FRED_` into the 
 By default, variables are baked into our bundle at build time. However, you can define variables as runtime environment variables, which can be changed when running a version of Fred built with `FRED_RUNTIME_ENV=true`.
 
 This is used in our npm package, so we can e.g. set `FRED_WRITER_MODE=true` in the `content` repo, without having to bake that into the npm package for all consumers.
+
+Runtime environment variables fall back to their build-time values if not set, and then to the default defined in code.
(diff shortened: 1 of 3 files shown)


components/outer-layout/server.js
```diff
@@ -93,9 +91,9 @@ export class OuterLayout extends ServerComponent {
             content="width=device-width, initial-scale=1.0"
           />
           <title>${context.pageTitle || "MDN"}</title>
-          ${RUNTIME_ENV
+          ${RUNTIME_ENV && runtimeEnvEntries.length > 0
```
caugner, line 94: Nit: I would avoid storing entry arrays. Was there a reason not to just add the following condition? 
```suggestion
          ${RUNTIME_ENV && Object.keys(env).length > 0
```

components/env/README.md
```diff
@@ -26,3 +26,5 @@ Rspack will bundle **all** environment variables prefixed with `FRED_` into the
 By default, variables are baked into our bundle at build time. However, you can define variables as runtime environment variables, which can be changed when running a version of Fred built with `FRED_RUNTIME_ENV=true`.
 
 This is used in our npm package, so we can e.g. set `FRED_WRITER_MODE=true` in the `content` repo, without having to bake that into the npm package for all consumers.
+
+Runtime environment variables fall back to their build-time values if not set, and then to the default defined in code.
```
caugner, line 30: Nit: Could mention the specific use case.

```suggestion
Runtime environment variables fall back to their build-time values if not set (e.g. when using `fred-ssr` in content workflows), and then to the default defined in code.
```

caugner (approved):

PR #693, opened 2025-09-03
fix(doc): use pageTitle as `<title>`

caugner:
### Description

Fixes the `<title>` to use `doc.pageTitle`, not `doc.title`.

### Motivation

Consistency with yari.


LeoMcA (approved):
We should probably clean this up in rari too, so we always set pageTitle on the root: https://github.com/mdn/rari/issues/302

PR #700, opened 2025-09-04
fix(scrim-inline): add missing style

### LINKED ISSUE #686: Scrimba embeds under "test you skills" disappeared

The scrimba embeds on the following pages have disappeared since Fred release: 

https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/CSS_layout/Test_your_skills/Flexbox
https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting/Test_your_skills/Arrays
https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting/Test_your_skills/Functions
https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting/Test_your_skills/Variables
https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Test_your_skills/Box_model

The heading and content explaining the "interactive challenges" still exists. Please add the scrimba embeds back. 

Expected to see the scrimba embeds. 

Image

_No response_

caugner:
### Description

Add missing styles for `<mdn-scrim-inline>` elements.

### Motivation

Ensures that the scrims are shown again.

Fixes https://github.com/mdn/fred/issues/686 (together with https://github.com/mdn/content/pull/41024).

### DIFF
--- a/components/scrim-inline/element.css
+++ b/components/scrim-inline/element.css
@@ -2,6 +2,13 @@
 
 :host {
   display: block;
+
+  width: 100%;
+  max-width: 36rem;
+  aspect-ratio: 1.5;
+
+  margin: 0.5rem auto;
+
   overflow: hidden;
 }
 
--- a/components/scrim-inline/global.css
+++ b/components/scrim-inline/global.css
@@ -7,13 +7,3 @@
     url("./assets/BarlowCondensed-SemiBold.woff2") format("woff2");
   font-display: block;
 }
-
-mdn-scrim-inline {
-  display: block;
-
-  width: 100%;
-  max-width: 36rem;
-  aspect-ratio: 1.5;
-
-  margin: 0.5rem auto;
-}
--- a/components/scrim-inline/global.css
+++ b/components/scrim-inline/global.css
@@ -7,3 +7,13 @@
     url("./assets/BarlowCondensed-SemiBold.woff2") format("woff2");
   font-display: block;
 }
+
+mdn-scrim-inline {
+  display: block;
+
+  width: 100%;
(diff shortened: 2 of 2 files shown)


pepelsbey:
I wonder why we are exposing these styles globally. They could be scoped to the component: both font declaration and the element’s styles (via the `:host` selector). This is generally preferable.

caugner:
> I wonder why we are exposing these styles globally. They could be scoped to the component: both font declaration and the element’s styles (via the `:host` selector). This is generally preferable.

Good catch, I moved it in 6c4e1dc9, and it still works.

pepelsbey (approved):
Looks good! Thank you :)

LeoMcA:
> I wonder why we are exposing these styles globally. They could be scoped to the component: both font declaration and the element’s styles (via the :host selector). This is generally preferable.

No it's not: the styles affect layout, so will cause a layout shift on browsers with don't support DSD, or if the element isn't being SSRed (and the latter seems to be true of this element, as it happens on DSD-supporting browsers).

@caugner please could you fix this?

caugner:
@LeoMcA Good point, but since the layout shift appears to be happening below the fold, is this really an issue? I believe the CLS metric only [considers the current viewport](https://web.dev/articles/cls#layout-shift-score).

LeoMcA:
> since the layout shift appears to be happening below the fold, is this really an issue?

@caugner Yes, layout shifts are a suboptimal experience regardless of where they appear on page. We also don't guarantee this component appears below the fold: they could be placed higher in content, they're high on the page on the curriculum homepage, and we regularly link to headings.

PR #703, opened 2025-09-04
fix(content-section): restrict fragment link style to dt

### LINKED ISSUE #702: Doc fragment links not discernible in `<dd>` elements

Same-doc fragment links inside `<dd>` elements inherit styles, cannot be distinguished from regular text except on hover.

STR:

1. Navigate to https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout#delay
2. Hover over "Reasons for delays longer than specified"

Color used for all `<a>` elements inside the `<dd>`s.

_No response_

We have a CSS style rule for `:is(.content-section dl) a[href^="#"]` which seems to be the culprit. Looks like we want to remove the link styles for these cases (e.g., `#delay`):

```
<dt id="delay">
  <a href="#delay"><code>delay</code> …
```

Image

caugner:
### Description

Restricts the fragment link style to `<dt>` elements (effectively excluding `<dd>`).

### Motivation

Makes fragment links in `<dd>` visible.

#### Before

image

#### After

image

Fixes https://github.com/mdn/fred/issues/702.

### DIFF
--- a/components/content-section/server.css
+++ b/components/content-section/server.css
@@ -251,7 +251,7 @@
       margin: 1rem 0 2rem;
     }
 
-    a[href^="#"] {
+    dt a[href^="#"] {
       color: inherit;
 
       &:hover::before {


Josh-Cena:
IMO, anchor links should still have the underline, because the underline is a complement of the color, so whenever the color is different, so should the underline. Also IMO, if I declared `- [xxx](#yyy)` in content, I will expect there to be a discernible link. The link style should only be removed if this anchor link is autogenerated.

caugner:
> IMO, anchor links should still have the underline, because the underline is a complement of the color, so whenever the color is different, so should the underline. Also IMO, if I declared `- [xxx](#yyy)` in content, I will expect there to be a discernible link. The link style should only be removed if this anchor link is autogenerated.

That's a related, but distinct concern.

We [intentionally](https://github.com/mdn/fred/blob/89244ab6f0c022efe12cced7bfbfe792143dedba/components/content-section/server.css#L63) style fragment links differently.

/cc @pepelsbey

Josh-Cena:
Yeah I know it's a separate rule, I'm saying that's not really the purpose of the underline, which was an accessibility device in the first place. Also what about manual links in DTs? I think a better approach is to generate class names in Rari for autogenerated links, and target the class name instead

pepelsbey:
> I think a better approach is to generate class names in Rari for autogenerated links, and target the class name instead

@Josh-Cena Do you mind filing a separate issue for that? I'm not a big fan of the current implementation, either. I think we should discuss it further to find a suitable solution.

bsmth (approved):
LGTM locally on:

http://localhost:3000/en-US/docs/Web/API/Window/setTimeout#delay

argl (approved):

PR #712, opened 2025-09-05
fix(playground): overlay run button for shared playground

### LINKED ISSUE #665: Improve discoverability of "Run" action for Playground links

OnkarRuikar:
- fix https://github.com/mdn/fred/issues/665

### Description

Need a prominent Run button on canvas for shared playgrounds because the shared code could be untrustworthy.

### Motivation

In the current state it is not apparent that users are supposed to click run button to get the output and begin autorun for the shared playgrounds.\
It was requested to have big button spanning the entire canvas.

> @OnkarRuikar that's ok, but given how unusual it is to look at a playground and not see the result, I feel this will be the likely first interaction on the page, so it need to be **big and cover the canvas**, and be easy to resolve.

### Additional details

For testing used this link: http://localhost:3000/en-US/play?id=rCfusnHbQaseoVnpt5%2BOn3b12l6jKOf8oDyb9smSb5ueTnvaFpg87%2FSZ4fWjiysCSJBwRrmdXiMXJnTQ

or

https://fred-pr712.review.mdn.allizom.net/en-US/play?id=rCfusnHbQaseoVnpt5%2BOn3b12l6jKOf8oDyb9smSb5ueTnvaFpg87%2FSZ4fWjiysCSJBwRrmdXiMXJnTQ

### before change

> before

### After changes

> with warning

#665


LeoMcA (commented):
I like the implementation, I think the design could be a little better - could we use the same icon we use in the top nav next to the playground link, obviously a little bigger, so it's almost like a video play icon overlay?

OnkarRuikar:
> could we use the same icon we use in the top nav next to the playground link,

Done. Used the `circle-play` icon.

> obviously a little bigger,

Made the size bigger and responsive.

> so it's almost like a video play icon overlay?

We play videos and run code so shouldn't it be called "MDN Runground"? :thinking: \
To not give false impression of existence video content in the output, can we use " ▶ Run" instead of just the icon?

https://github.com/user-attachments/assets/5990322e-4936-4e6e-ab63-9e0c84e79c21

LeoMcA:
Let me... "Run" this past our content team and see what they think

caugner:
I could imagine making the button smaller, and adding some sort of warning below, so that the user understands the risks of pressing run.

OnkarRuikar:
> I could imagine making the button smaller, and adding some sort of warning below, so that the user understands the risks of pressing run.

"Risk of pressing run", we don't want to scare users so much that they won't even run the code.  We already have report banner at  the top, how  about  we reword it like the following?

caugner:
https://fred-pr712.review.mdn.allizom.net/en-US/play?id=rCfusnHbQaseoVnpt5%2BOn3b12l6jKOf8oDyb9smSb5ueTnvaFpg87%2FSZ4fWjiysCSJBwRrmdXiMXJnTQ

caugner (changes requested):
The button seems to overflow weirdly now.

Could the button take the whole area of the iframe, and look like the other buttons on the page in terms of border/background color and hover effect?

OnkarRuikar:
> Could the button take the whole area of the iframe, and look like the other buttons on the page in terms of border/background color and hover effect?

@caugner Do you mean without the white padding around?  Like the following:

https://github.com/user-attachments/assets/0565f784-c93e-4156-8eb3-39b7c7c83921


For dark theme it's a bit tricky, because the button sits on top of a white iframe. So we need to set the background color manually. What theme color variable should I use for `background-color: var(--color-XXX)`? All the theme colors involve light shades and not a single one has pitch dark color. Or should I create a new theme variable just for this button with values white|black?

caugner:
> Do you mean without the white padding around? Like the following:

Yes, exactly like that.

> For dark theme it's a bit tricky, because the button sits on top of a white iframe. So we need to set the background color manually.

Could we use `background-color: var(--color-background-page)`?

OnkarRuikar:
Ok now the button covers entire output iframe. And it appears like other buttons around:
 

https://github.com/user-attachments/assets/e82c983e-94e3-45bf-bcda-609132d78574

Preview Link:
https://fred-pr712.review.mdn.allizom.net/en-US/play?id=rCfusnHbQaseoVnpt5%2BOn3b12l6jKOf8oDyb9smSb5ueTnvaFpg87%2FSZ4fWjiysCSJBwRrmdXiMXJnTQ

caugner (changes requested):
Thanks, UI is looking good, with these remaining nits:
- Use the same `border-radius` as `mdn-play-runner` (`.25rem`).
- Use `flex-grow: 1` like `mdn-play-runner`, and hide the `mdn-play-runner` conditionally if it the overlay button is shown (so that the button takes the same space, and we should no longer need `position: absolute`, `inset: ...`, and `z-index`).
- Use `border-color: currentcolor` like `mdn-button`.

image
image

caugner:
I wonder if we could/should add a text in the button below the "Run" label along these lines:

> This shared Playground was authored by a user, not MDN.
> Review the code carefully. Running it might expose your IP address.
> Do not enter sensitive information.

caugner:
> Use `border-color: currentcolor` like `mdn-button`.

@OnkarRuikar Let's also apply this change, to make it easier to identify the button as such.

OnkarRuikar:
How about now?
> with warning

components/playground/element.js
```diff
@@ -399,7 +401,40 @@ ${"```"}`,
...
+                      ${this.l10n(
+                        "playground-run-warning-line-three",
+                      )`Running malicious code can expose you to serious risks, including theft of data, IP address, browser fingerprint, current session, etc.`}
+                    </li>
+                    <li>
+                      ${this.l10n(
+                        "playground-run-warning-line-four",
+                      )`Never enter passwords, payment details, address, or any sensitive information.`}
+                    </li>
+                  </ul>
```
caugner: Discussed this with Leo. Let's make this a `<div>` with two lines:

```html
<div>
  This is a user-shared playground.<br />
  Always inspect the code before running it.
</div>
```

With this CSS:

```css
color: var(--color-text-secondary);
line-height: var(--font-line-normal);
margin-top: 0.5rem;
```

caugner (changes requested):

components/playground/element.js
OnkarRuikar: done

components/playground/element.js
```diff
@@ -399,7 +401,22 @@ ${"```"}`,
...
-            <mdn-play-runner></mdn-play-runner>
+            ${this._autoRun
+              ? nothing
+              : html`<mdn-button
+                  class="overlay-run-button"
+                  @click=${this._run}
+                  variant="plain"
+                >
+                  <div>${circlePlay} ${this.l10n("playground-run")`Run`}</div>
+                  <div>
```
caugner: Let's give these CSS classes:


```suggestion
                  <div class="overlay-run-button--header">${circlePlay} ${this.l10n("playground-run")`Run`}</div>
                  <div class="overlay-run-button--body">
```

components/playground/element.css
```diff
@@ -102,10 +102,35 @@
...
+
+        &.hidden {
+          display: none;
+        }
+      }
+
+      .overlay-run-button {
+        border-color: currentcolor;
+
+        div:first-child {
```
caugner:
```suggestion
        .overlay-run-button--header {
```

PS: Once we target the CSS class, we can unnest, i.e. pull this block (and the following one) out of `.overlay-run-button`.

components/playground/element.css
```diff
@@ -102,10 +102,35 @@
...
+          font-size: clamp(1.5rem, 3vw, 3rem);
+
+          svg {
+            width: clamp(1.5rem, 3vw, 3rem) !important;
+            height: clamp(1.5rem, 3vw, 3rem) !important;
+            vertical-align: bottom;
+          }
+        }
+
+        div:nth-child(2) {
```
caugner:
```suggestion
        .overlay-run-button--body {
```

components/playground/element.js
```diff
@@ -399,7 +401,22 @@ ${"```"}`,
...
+                    This is a user-shared playground.<br />
+                    Always inspect the code before running it.
+                  </div>
+                </mdn-button>`}
+            <mdn-play-runner
+              class=${this._autoRun ? "" : "hidden"}
```
caugner: I think `""` causes an empty attribute to be added, and `nothing` should avoid this.

```suggestion
              class=${this._autoRun ? nothing : "hidden"}
```

caugner (changes requested):

l10n/locales/fr.ftl
```diff
@@ -272,6 +272,7 @@ playground-share = Partager
 playground-clear = Effacer
 playground-reset = Réinitialiser
 playground-seeing-something-inappropriate = Voir quelque chose d'inapproprié ?
+playground-user-shared-warning = Il s'agit d'un terrain d'essai partagé par un·e utilisateur·ice.<br>Inspectez toujours le code avant de l'exécuter.
```
caugner, line 275: cc @mdn/fr-content for review

caugner (approved):

l10n/locales/fr.ftl
tristantheb: Looks good 👍🏻

caugner:
Thank you @OnkarRuikar! 🎉

Issue #713, opened 2025-09-05
Investigate decline in Scrimba traffic


s-sood:
### What information was incorrect, unhelpful, or incomplete?

Since Fred launch, the traffic to Scrimba via MDN has dropped 20-25%. To investigate what is causing this decline we need to answer the following questions using telemetry data: 
- Has the traffic on curriculum declined after Fred launch? 
- What is the impact of main menu navigation changes on traffic per topic/area? 
- Could it be that we accidently removed the scrim embeds on "test your skills" pages? github.com/mdn/fred/issues/686

### What did you expect to see?

<img width="1119" height="428" alt="Image" src="https://github.com/user-attachments/assets/ea64dcdd-6e57-43f5-a202-ebb9d3dc532b" />

caugner:
> Has the traffic on curriculum declined after Fred launch?

Yes, Curriculum traffic has declined significantly: 62 % less page views, 68 % less users.

See: https://sql.telemetry.mozilla.org/queries/96599?p_Range=d_last_90_days&p_referrer=%25&p_resolution=WEEK&p_url=%25%2Fcurriculum%2F%25#238439 (Mozilla-internal)

> What is the impact of main menu navigation changes on traffic per topic/area?

To answer this, we will need to measure the menu clicks. I will open a PR today to add this measurement.

> Could it be that we accidently removed the scrim embeds on "test your skills" pages?

Yes, there was a bug (missing styles) that prevented the Scrim embeds from showing up on those pages. The fix landed in the night from Friday to Saturday.

caugner:
> What is the impact of main menu navigation changes on traffic per topic/area?

@s-sood Note: I have updated [this query](https://sql.telemetry.mozilla.org/queries/108065?p_days=21&p_mid=2025-08-19&p_url=%25) (Mozilla-internal), which allows you to compare page views in the 21 days before/after the Fred launch by several dimensions, including topic (e.g. "curriculum").

caugner:
I extracted https://github.com/mdn/fred/issues/738 for measuring menu clicks, and opened a PR that implements this measurement.

caugner:
Reducing to p1, as https://github.com/mdn/fred/pull/743 lands tonight.

caugner:
Unfortunately, there was a bug that likely caused the Scrimba fallback banner to only show up if fetching the placements failed, and prevented it from showing when no ad was served.

See: https://github.com/mdn/fred/pull/838

caugner:
Closing as resolved.

Issue #647, opened 2025-08-27
[meta] Review Glean measurements


caugner:
Only a few Glean measurements were ported from yari, and we should probably:

- [x] Check that all relevant measurements were ported
- [ ] Check that no irrelevant measurements were ported

Missing measurements will most likely be noticed when we're requested to update some dashboards.

caugner:
All missing measurements are now tracked in issues.

PR #743, opened 2025-09-10
fix(glean): make click measurements backward compatible

caugner:
### Description

1. Remove `gleanClick` options, accepting a single `id` (equivalent to `source` in yari).
2. Align measurement IDs with yari.
3. Replace `data-glean` attributes (no effect in fred) with `data-glean-id`.

### Motivation

Make the Fred click measurements work, and backward compatible with yari click measurements.

Related to https://github.com/mdn/fred/issues/647 and https://github.com/mdn/fred/issues/713.


argl (approved):
nice!

PR #739, opened 2025-09-10
feat(menu): measure clicks

caugner:
### Description

Adds click measurements to the menu.

### Motivation

We want to understand how the new menu performs compared to the old one.

### Additional details

Ran the following JavaScript snippet in the browser console on  http://localhost:3000/en-US/docs/Web/HTML/Reference/Elements:

```js
[...document.querySelectorAll('[data-glean-id]')].map(el => el.getAttribute('data-glean-id')).toSorted().join("\n")
```

This yielded the following result:

```
menu_click_link: top-level -> /en-US/blog/
menu_click_menu: all -> /en-US/docs/Web
menu_click_menu: css -> /en-US/docs/Web/CSS
menu_click_menu: html -> /en-US/docs/Web/HTML
menu_click_menu: javascript -> /en-US/docs/Web/JavaScript
menu_click_menu: learn -> /en-US/docs/Learn_web_development
menu_click_menu: webapis -> /en-US/docs/Web/API
menu_click_submenu: about -> /en-US/about
menu_click_submenu: about -> /en-US/advertising
menu_click_submenu: about -> /en-US/community
menu_click_submenu: about -> https://github.com/mdn
menu_click_submenu: all -> /en-US/docs/Mozilla/Add-ons/WebExtensions
menu_click_submenu: all -> /en-US/docs/Web
menu_click_submenu: all -> /en-US/docs/Web/API/Performance
menu_click_submenu: all -> /en-US/docs/Web/Accessibility
menu_click_submenu: all -> /en-US/docs/Web/HTTP
menu_click_submenu: all -> /en-US/docs/Web/Media
menu_click_submenu: all -> /en-US/docs/Web/Privacy
menu_click_submenu: all -> /en-US/docs/Web/Progressive_web_apps
menu_click_submenu: all -> /en-US/docs/Web/Security
menu_click_submenu: all -> /en-US/docs/Web/URI
menu_click_submenu: all -> /en-US/docs/Web/WebDriver
menu_click_submenu: all -> /en-US/docs/WebAssembly
menu_click_submenu: css -> /en-US/docs/Web/CSS/CSS_Values_and_Units
menu_click_submenu: css -> /en-US/docs/Web/CSS/CSS_animations/Using_CSS_animations
menu_click_submenu: css -> /en-US/docs/Web/CSS/CSS_box_model/Introduction_to_the_CSS_box_model
menu_click_submenu: css -> /en-US/docs/Web/CSS/CSS_colors
[...]

PR #748, opened 2025-09-11
chore(deploy): enable Glean debug

caugner:
### Description

Enables Glean debug pings by default on review deployments.

### Motivation

Allows us to validate Glean measurements directly on the PR review deployment, without having to run the branch locally.

Related to https://github.com/mdn/fred/pull/739#pullrequestreview-3209811304.

### DIFF
--- a/.github/workflows/_deploy.yml
+++ b/.github/workflows/_deploy.yml
@@ -235,6 +235,9 @@ jobs:
 
       - name: Render (fred)
         working-directory: mdn/fred
+        env:
+          FRED_GLEAN_ENABLED: true
+          FRED_GLEAN_DEBUG: true
         run: npm run ssr
 
       - name: Authenticate with GCP


.github/workflows/_deploy.yml
```diff
@@ -235,6 +235,9 @@ jobs:
 
       - name: Render (fred)
         working-directory: mdn/fred
+        env:
+          FRED_GLEAN_ENABLED: true
+          FRED_GLEAN_DEBUG: true
```
LeoMcA, line 240: I think these need to be set in the "Build (fred)" step above, instead

caugner:
As discussed on Slack, Glean is already enabled:

https://github.com/mdn/fred/blob/98087383ecb9a50e43cd1da746ea8aa591e19939/.github/workflows/_deploy.yml#L122-L126C11

So we can just use [debugging options](https://mozilla.github.io/glean.js/debugging/options/) to enable console log or debug view on demand.

PR #751, opened 2025-09-11
fix(scrim-inline): use `data-glean-id`, not `data-glean`

caugner:
### Description

Uses `data-glean-id` instead of `data-glean` (there was one indirect usage via `dataset`).

### Motivation

Make the measurement work.

Follow-up of https://github.com/mdn/fred/pull/743.


components/scrim-inline/element.js
```diff
@@ -142,7 +142,8 @@ export class MDNScrimInline extends LitElement {
    */
   #toggle(e) {
     if (e.target instanceof HTMLElement) {
-      e.target.dataset.glean = `curriculum: scrim fullscreen -> ${this._fullscreen ? 0 : 1} id:${this._scrimId}`;
+      e.target.dataset["glean-id"] =
```
LeoMcA: In today's edition of "what were the spec authors thinking?!", you've gotta do this instead:

```suggestion
      e.target.dataset.gleanId =
```

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/dataset#in_javascript

The current code throws an exception

caugner: Ouch, good catch!

caugner:
I consider this change approved, as it is exactly what Leo suggested in https://github.com/mdn/fred/pull/751#discussion_r2341623891, so this is equivalent to him opening a PR with exactly this change, and myself approving it. In other words, we unanimously agree on this change.

Also, I verified the fix on https://fred-pr751.review.mdn.allizom.net/en-US/curriculum/, comparing with https://fred.review.mdn.allizom.net/en-US/curriculum/.

LeoMcA:
Apologies, yes: Thursday evening brain from me there - thanks for merging - I meant to approve.

Issue #752, opened 2025-09-11
Glean `data-glean-id` not working in shadow DOM


caugner:
1. Go to https://developer.mozilla.org/en-US/curriculum/
2. Run `Glean.setLogPings(true);`
3. Click on Scrim.

Expected: to see Glean measurement in console (+ network tab)

Actual: no measurement

---

<img width="1219" height="377" alt="Image" src="https://github.com/user-attachments/assets/c1b39041-c417-4743-aab0-a3a8b81b8fee" />

LeoMcA:
Fixing upstream in https://github.com/mozilla/glean.js/pull/2013

@caugner since we've shipped our own fix, perhaps we should still close this? Have the glean pings shown up on redash?

caugner:
> since we've shipped our own fix, perhaps we should still close this? Have the glean pings shown up on redash?

Yes, they have! 🙂

PR #753, opened 2025-09-11
fix(glean): trigger gleanClick for clicks in Shadow DOMs

LeoMcA:
Relates to https://github.com/mdn/fred/issues/752

Fix locally before we fix upstream.

Tested locally:
- we get pings in shadow roots: e.g. https://debug-ping-preview.firebaseapp.com/pings/mdn-dev/4721d6b1-38ee-4cd8-9375-c42531547c0c
- we don't get double pings outside of shadow roots

We'll need to revert this once it's fixed and released upstream.

### DIFF
--- a/hooks/glean-init.js
+++ b/hooks/glean-init.js
@@ -6,6 +6,7 @@ import {
   GLEAN_DEBUG,
   GLEAN_ENABLED,
 } from "../components/env/index.js";
+import { gleanClick } from "../utils/glean.js";
 
 const FIRST_PARTY_DATA_OPT_OUT_COOKIE_NAME = "moz-1st-party-data-opt-out";
 const GLEAN_APP_ID = "mdn-fred";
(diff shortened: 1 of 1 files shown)


caugner (approved):
Code LGTM, and tested on https://fred-pr753.review.mdn.allizom.net/en-US/curriculum/.

hooks/glean-init.js
```diff
@@ -26,3 +27,16 @@ Glean.initialize(GLEAN_APP_ID, uploadEnabled, {
   enableAutoElementClickEvents: true,
   channel: GLEAN_CHANNEL,
 });
+
+document.addEventListener("click", (event) => {
+  const composedTarget = event.composedPath()?.[0];
+  if (composedTarget !== event.target && composedTarget instanceof Element) {
+    const taggedElement = composedTarget.closest("[data-glean-id]");
+    if (taggedElement instanceof HTMLElement) {
+      const gleanId = taggedElement.dataset.gleanId;
```
caugner, line 36: Nit: Maybe more concise, but could be more bits compressed.
```suggestion
      const { gleanId } = taggedElement.dataset;
```

LeoMcA: Ha, I always forget I can do this: destructing only enters my mind when I have more than 1 variable to extract from an object

PR #768, opened 2025-09-16
fix(server-component): exclude empty components from used set

LeoMcA:
Noticed this while working on inlining the baseline sticky code in its component.

We still add a component rendering nothing to the used set, which means we load its CSS (and potentially inline JS depending on how I implement that) - when there's no reason to. This changes our server component logic to remove the component from the sets tracking its use if it renders nothing.

The tangible effect of this is, e.g.
- On pages without a Baseline banner we don't load the banner css
- On en-US pages we don't load the translated content banner css

### DIFF
--- a/components/server/index.js
+++ b/components/server/index.js
@@ -1,4 +1,4 @@
-import { html } from "lit";
+import { html, nothing } from "lit";
 
 import { stylesForComponents } from "../outer-layout/utils.js";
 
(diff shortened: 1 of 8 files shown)


components/server/index.js
```diff
@@ -12,11 +12,12 @@ export class ServerComponent {
    * @template {typeof ServerComponent} T
    * @this {T}
    * @param {Parameters<InstanceType<T>["render"]>} args
-   * @returns {ReturnType<InstanceType<T>["render"]> | import("@lit").TemplateResult}
+   * @returns {ReturnType<InstanceType<T>["render"]> | import("@lit").TemplateResult | import("@lit").nothing }
    */
   static render(...args) {
     const { componentsUsed, componentsWithStylesInHead, compilationStats } =
       asyncLocalStorage.getStore() || {};
```
caugner, line 19: Why don't we need to initialize these?

```suggestion
      asyncLocalStorage.getStore() || {
        componentsUsed: new Set(),
        componentsWithStylesInHead: new Set(),
        compilationStats: new Set(),
      };
```

caugner (approved):
LGTM, just the uninitialized vars confuse me a bit.

components/server/index.js
LeoMcA: Something's gone horrendously wrong if `getStore()` actually resolves to undefined: we should probably just bail out with an error rather than having to safe navigate everywhere. I'll merge this and create another PR to fix that.

LeoMcA: https://github.com/mdn/fred/pull/770

PR #781, opened 2025-09-17
fix(types): update to match fixed rari types

LeoMcA:
Opening as draft because this depends on https://github.com/mdn/rari/pull/314 and should be merged once that's published to npm and updated in fred.

This can be tested by doing:
1. Checking out that rari branch
2. Running:
3. `cargo run export-schema`
4. `cd rari-npm`
5. `cp ../schema.json schema.json`
6. `npm run generate-types`
7. `cd ../../fred`
8. `npm install ../rari/rari-npm`
9. `npm run tsc`

### DIFF
--- a/components/observatory-results/server.js
+++ b/components/observatory-results/server.js
@@ -10,7 +10,7 @@ export class ObservatoryResults extends ServerComponent {
    * @param {import("@fred").Context<import("@rari").SPAPage>} context
    */
   render(context) {
-    if (context.parents.length > 0) {
+    if (context.parents !== undefined && context.parents.length > 0) {
       const lastParent = context.parents.at(-1);
       if (lastParent) {
         lastParent.uri = "#";
--- a/components/reference-layout/server.js
+++ b/components/reference-layout/server.js
@@ -16,9 +16,8 @@ export class ReferenceLayout extends ServerComponent {
    */
   render(context) {
     const { doc } = context;
-    const [description, ...sections] = doc.body.map((section) =>
-      ContentSection.render(context, section),
-    );
+    const [description, ...sections] =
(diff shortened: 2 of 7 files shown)


components/observatory-results/server.js
```diff
@@ -10,7 +10,7 @@ export class ObservatoryResults extends ServerComponent {
    * @param {import("@fred").Context<import("@rari").SPAPage>} context
    */
   render(context) {
-    if (context.parents.length > 0) {
+    if (context.parents !== undefined && context.parents.length > 0) {
```
caugner, line 13: Nit: We do this elsewhere

```suggestion
    if ("parents" in context && context.parents.length > 0) {
```

components/reference-layout/server.js
```diff
@@ -16,9 +16,8 @@ export class ReferenceLayout extends ServerComponent {
    */
   render(context) {
     const { doc } = context;
-    const [description, ...sections] = doc.body.map((section) =>
-      ContentSection.render(context, section),
-    );
+    const [description, ...sections] =
```
caugner, line 19: (Unrelated, but the names `description` and `sections` here confused me a bit. We could just call them `firstSection` and `otherSections`.)

caugner (approved):

components/observatory-results/server.js
LeoMcA: I believe this *would* work with how our data is serialized, but thanks to the loveliness of javascript it would cause tsc to fail, as you could still have:
```js
parents = {
  context: undefined
}
```

LeoMcA:
Merging because these changes appear to be backwards compatible

PR #782, opened 2025-09-17
fix: only 'allow-modals' in interactive examples and live samples

LeoMcA:
Follow up to https://github.com/mdn/fred/pull/634

Only set `allow-modals` on interactive examples and live samples, rather than all instances of the play runner (which includes the playground).


components/interactive-example/with-tabs.js
```diff
@@ -51,7 +51,7 @@ export const InteractiveExampleWithTabs = (Base) =>
               <h4>${this.l10n`Output`}</h4>
               <mdn-play-runner
                 ${ref(this._runner)}
-                sandbox="allow-top-navigation-by-user-activation"
+                sandbox="allow-top-navigation-by-user-activation allow-modals"
```
caugner: Nit: Order alphabetically.

```suggestion
                sandbox="allow-modals allow-top-navigation-by-user-activation"
```

components/interactive-example/with-choices.js
```diff
@@ -163,6 +163,7 @@ export const InteractiveExampleWithChoices = (Base) =>
               <mdn-play-runner
                 ${ref(this._runner)}
                 defaults="ix-choice"
+                sandbox="allow-modals"
```
caugner, line 166: Nit: I don't think CSS choice examples need this.

```suggestion
```

components/live-sample-result/element.js
```diff
@@ -109,7 +109,9 @@ export class MDNLiveSampleResult extends L10nMixin(LitElement) {
           @mdn-play-runner-src=${this._runnerSrcUpdated}
           .code=${this.code}
           .allow=${this.allow}
-          .sandbox=${this.sandbox}
+          .sandbox=${[
+            ...new Set(["allow-modals", ...(this.sandbox?.split(" ") || [])]),
+          ].join(" ")}
```
caugner, line 114: Nit: Maybe assign this to a variable above, rather than inline?

caugner (approved):

components/interactive-example/with-choices.js
LeoMcA: That was actually one of the examples the original addition was requested for: https://developer.mozilla.org/en-US/docs/Web/CSS/break-after

PR #788, opened 2025-09-17
fix(server-components): use lit ssr html template tag

LeoMcA:
Changes all imports of `html` from `lit` in server components to be from `@lit-labs/ssr` instead.

`html` from `lit` includes comments necessary for hydration, but we'll never hydrate server components: so there's no need for these comments - `html` from `@lit-labs/ssr` doesn't include them.

This allows for some inline stuff I want to do in the `ServerComponent` class itself, cleans up some slightly messy HTML in our pages, and saves us a bit of bandwidth/storage space:

```
ls -lh out/index.html out-lit-ssr-html/en-us/docs/mdn/kitchensink/index.html      
-rw-r--r-- 1 leo leo 170K Sep 17 18:47 out-lit-ssr-html/en-us/docs/mdn/kitchensink/index.html
-rw-r--r-- 1 leo leo 187K Sep 17 18:51 out/index.html

ls -lh out/index.html.br out-lit-ssr-html/en-us/docs/mdn/kitchensink/index.html.br
-rw-r--r-- 1 leo leo 21K Sep 17 18:47 out-lit-ssr-html/en-us/docs/mdn/kitchensink/index.html.br
-rw-r--r-- 1 leo leo 22K Sep 17 18:51 out/index.html.br
```

Tested by doing a full build locally with `./build.sh`


caugner (approved):
LGTM, but we should also enforce this with ESLint.

PR #795, opened 2025-09-18
feat(glean): measure search usage

### LINKED ISSUE #750: Measure search usage
From discussion with @s-sood.

Port from yari:

## Quick search:

- Click on "Search" buttons in homepage hero, and header)
- Type in search field (measure only once)
- Click on result item

## Site search:

- Keywords with zero results
- Click on result item

caugner:
### Description

Adds measurements for "quick" search (modal), and "site" search (full-text search).

### Motivation

Ports measurements from yari, refines them, and adds some useful ones (e.g., zero site-search results).

### Additional details

The following measurements are added:

| ID | Event |
|--------|--------|
| `quick-search-open: {homepage,menu,keyboard -> {slash,ctrl-k}}` | User opens search modal from homepage, menu, or via keyboard (Slash or CTRL+K) |
| `quick-search-change: {type,paste,selection}` | User engages with search modal by typing, pasting, or text selection (on open) |
| `quick-search: results[<index>] -> <query> -> <url>` | User navigates to a result item |
| `quick-search: site-search -> <query>` | User navigates to site search |
| `site-search: no-results` | Site search returns empty search results |
| `site-search: results[<index>] -> <query> -> <url>` | User clicks a given result item in the site search | 

Fixes https://github.com/mdn/fred/issues/750.

### DIFF
--- a/components/search-modal/element.js
+++ b/components/search-modal/element.js
@@ -15,7 +15,6 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
   static styles = styles;
 
   static properties = {
-    _hasChanged: { state: true },
     _index: { state: true },
     _query: { state: true },
     _selected: { state: true },
@@ -29,6 +28,8 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
     this._query = "";
     this._selected = 0;
     this._shiftFocus = false;
+    /** Capture whether user has engaged with the search. */
+    this._hasEngaged = false;
   }
 
   async _loadIndex() {
@@ -66,8 +67,8 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
   _input({ inputType, target }) {
     if (target instanceof HTMLInputElement) {
       this._query = target.value;
-      if (!this._hasChanged && inputType.startsWith("insert")) {
-        this._hasChanged = true;
+      if (!this._hasEngaged && inputType.startsWith("insert")) {
+        this._hasEngaged = true;
         gleanClick(
           `quick-search-change: ${inputType === "insertFromPaste" ? "paste" : "type"}`,
         );
--- a/components/search-modal/element.js
+++ b/components/search-modal/element.js
@@ -104,6 +104,9 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
               metaKey,
             }),
           );
+          // Workaround: For some reason, the dispatched click
+          // does not cause Glean.js to capture it via `data-glean-id`.
+          gleanClick(`quick-search: keyboard -> ${this._query}`);
         }
         break;
       }
@@ -262,7 +265,7 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
                   <li ?data-selected=${this._selected === i} data-result=${i}>
                     <a
                       href=${url}
-                      data-glean-id=${`quick-search: modal -> ${this._query}`}
+                      data-glean-id=${`quick-search: click -> ${this._query}`}
                       ><span class="slug"
                         >${mdnUrl2Breadcrumb(url, this.locale)}</span
                       >
--- a/components/search-modal/element.js
+++ b/components/search-modal/element.js
@@ -107,7 +107,9 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
           );
           // Workaround: For some reason, the dispatched click
           // does not cause Glean.js to capture it via `data-glean-id`.
-          gleanClick(`quick-search: keyboard -> ${this._query}`);
+          if (item.dataset.gleanId) {
+            gleanClick(item.dataset.gleanId);
+          }
         }
         break;
       }
@@ -229,6 +231,9 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
 
   render() {
     const siteSearchIndex = this._queryIndex.value?.length || 0;
+    const searchUrl = this._query
+      ? `/${this.locale}/search?${new URLSearchParams({ q: this._query })}`
+      : null;
     return html`
       <dialog @keydown=${this._keydown} @focusin=${this._focus} closedby="any">
         <form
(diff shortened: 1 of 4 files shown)


components/search-modal/element.js
```diff
@@ -13,6 +14,7 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
   static styles = styles;
 
   static properties = {
+    _hasChanged: { state: true },
```
LeoMcA: This shouldn't be necessary: we're not re-rendering based on `_hasChanged`

components/search-modal/element.js
```diff
@@ -60,9 +62,15 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
...
 
   /** @param {InputEvent} event */
-  _input({ target }) {
+  _input({ inputType, target }) {
     if (target instanceof HTMLInputElement) {
       this._query = target.value;
+      if (!this._hasChanged && inputType.startsWith("insert")) {
+        this._hasChanged = true;
+        gleanClick(
+          `quick-search-change: ${inputType === "insertFromPaste" ? "paste" : "type"}`,
```
LeoMcA, line 73: This seems unnecessary information to capture: let's say we find the majority of users e.g. paste inputs rather than type them in, what would we do with that information?

components/site-search/element.js
```diff
@@ -327,14 +334,17 @@ export class MDNSiteSearch extends L10nMixin(LitElement) {
...
                     html`<li class="site-search-results__item">
                       <article>
                         <p class="site-search-results__path">
                           ${mdnUrl2Breadcrumb(result.mdn_url, this.locale)}
                         </p>
                         <h2 class="site-search-results__title">
-                          <a href=${result.mdn_url}>
+                          <a
+                            href=${result.mdn_url}
+                            data-glean-id=${`site-search: click -> ${1 + (results.metadata.page - 1) * results.metadata.size + index}`}
```
LeoMcA: Should we be gathering this for quicksearch too?

LeoMcA (approved):
I'd suggest we only capture information for a known question we have - I'm not sure what questions pings like `quick-search-change` could be answering, but I wasn't in the discussions - merge if there is one.

components/search-modal/element.js
caugner: So just initialize in the constructor instead?

components/search-modal/element.js
caugner: The main point here is that we're capturing search engagement, and while we do this, we can distinguish typing from pasting for free.

components/site-search/element.js
caugner: For quick search, we currently gather the query as the part behind `->`, so for backward-compatible, I didn't change that.

I think it's more relevant for site search, where we might change the algorithm, or update ElasticSearch, and we'll see the effect.

components/search-modal/element.js
LeoMcA: No need: the existence of it here does no initialization, just tells Lit to look out for changes to it to re-render. With no initialization in the constructor, it defaults to undefined, which is falsy: hence why your code works. You can add it for completeness if you want.

components/site-search/element.js
LeoMcA: I can imagine a future where we want to compare our simple title-based search in quicksearch with something more advanced, but we can always set up metrics at the time and A/B test or something.

components/search-modal/element.js
caugner: Fixed via [3c6eeef](https://github.com/mdn/fred/pull/795/commits/3c6eeef64218aadf75392e7f351015f8b84acc76), and renamed to `_hasEngaged` to make the intent clearer.

components/site-search/element.js
caugner: Found a mostly backward-compatible way of capturing the index for quick search and site search in a similar way, and added the url as well, as pages most frequently navigated to via page would be candidates for exposure in the menu.

See 4e1aa081a221c24ca3a39d2bdc28b2aa483b2c85.

caugner:
@LeoMcA PTAL, if you don't mind. I would like to get this right.

components/search-modal/element.js
LeoMcA: Just because we can doesn't mean we should: but I won't block on this

components/search-modal/element.js
```diff
@@ -105,7 +106,9 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
           );
           // Workaround: For some reason, the dispatched click
           // does not cause Glean.js to capture it via `data-glean-id`.
-          gleanClick(`quick-search: keyboard -> ${this._query}`);
+          if (item.dataset.gleanId) {
+            gleanClick(item.dataset.gleanId);
+          }
```
LeoMcA: Add `composed: true` to the `MouseEvent` above (so it escapes the shadow dom to our document event listener) and we can remove this workaround

LeoMcA (approved):
Looks good! Approved with the `composed: true` change.

components/search-modal/element.js
caugner: TIL, thanks, fixed in [325f93f](https://github.com/mdn/fred/pull/795/commits/325f93ffe4dfcbd2e0a34bd1a7e041d2078472a1).

components/search-modal/element.js
```diff
@@ -187,6 +187,10 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
       this.showModal();
       if (selection) {
         this._query = selection;
+        if (!this._hasEngaged) {
+          this._hasEngaged = true;
+          gleanClick("quick-search-change: selection");
+        }
```
LeoMcA, line 193: Ah nice, didn't realise we even supported this: perhaps you've convinced me this metric does matter - could build an argument for improving its discoverability

LeoMcA (approved):

Issue #684, opened 2025-09-02
Add option to auto open browser when launching


LeoMcA:
Request from content editorial meeting. Should probably be opt-in with an environment variable.

brysonbw:
@LeoMcA Created PR (#805)  to resolve - would appreciate your feedback and/or review

PR #805, opened 2025-09-19
feat(server): add auto open browser when launching dev server

brysonbw:
### Description

- Add env var to allow option to auto open browser when launching dev server
  - Added `utils/validate-url.js` file with `isURLValid()` for extra sanity before opening dev URL and possibly a useful helper for other use cases when needing to validate a url

- #684


brysonbw:
Could also use a popular lib [open](https://www.npmjs.com/package/open) if that’s the preferred approach. I just opted not to here to avoid adding another dependency to the project, but either way is smooth. Or if someone thinks differently that's smooth too

caugner:
@brysonbw Can you please disclose if you used AI tools to author this PR (and which)?

brysonbw:
@caugner Before I give my answer - could you clarify your intent behind asking this question?

server.js
```diff
@@ -323,9 +325,27 @@ export async function startServer() {
...
+      if (platform === "darwin") {
+        // macOS
+        exec(`open ${url}`);
+      } else if (platform === "win32") {
+        // Windows
+        exec(`start ${url}`);
+      } else {
+        // Linux
+        exec(`xdg-open ${url}`);
+      }
```
caugner: Even though the `url` is safe here, we should always avoid `exec()`.
Let's use `spawn()` instead.

```suggestion
      const command = platform === "win32" ? "start" : platform === "darwin" ? "open" : "xdg-open";
      spawn(command, [url]);
```

server.js
```diff
@@ -323,9 +325,27 @@ export async function startServer() {
   }
 
   const httpServer = app.listen(PORT, () => {
-    console.log(
-      `Server started at ${http2 ? "https" : "http"}://localhost:${PORT}`,
+    const scheme = http2 ? "https" : "http";
+    const url = `${scheme}://localhost:${PORT}`;
+    console.log(`Server started at ${url}`);
+    // Auto open browser
+    const OPEN_BROWSER_ON_START = JSON.parse(
```
caugner: Let's move this into `./components/env`, and parse the environment variable like others.

server.js
```diff
@@ -323,9 +325,27 @@ export async function startServer() {
...
+      process.env.OPEN_BROWSER_ON_START || "null",
     );
+    if (devMode && OPEN_BROWSER_ON_START && isURLValid(url, [scheme])) {
```
caugner: We already ensure that `PORT` is a number, so we don't need to validate the URL:

https://github.com/mdn/fred/blob/42d79e42187698697d080d2d5d7de721e96220e8/components/env/index.js#L10

utils/validate-url.js
caugner, line 1: Remove, see above.

server.js
```diff
@@ -1,3 +1,4 @@
+import { exec } from "node:child_process";
```
caugner:
```suggestion
import { spawn } from "node:child_process";
```

server.js
```diff
@@ -9,6 +10,7 @@ import openEditor from "open-editor";
 
 import { FRED_BUILD_ROOT } from "./build/env.js";
 import { PLAYGROUND_PORT, PORT, WRITER_MODE } from "./components/env/index.js";
+import { isURLValid } from "./utils/validate-url.js";
```
caugner:
```suggestion
```

caugner (changes requested):

caugner (approved):
Code LGTM, assuming this feature is desired.

Requesting a second opinion from @LeoMcA, and cc'ing @mdn/content-team.

server.js
```diff
@@ -323,9 +329,21 @@ export async function startServer() {
...
 
   const httpServer = app.listen(PORT, () => {
-    console.log(
-      `Server started at ${http2 ? "https" : "http"}://localhost:${PORT}`,
-    );
+    const scheme = http2 ? "https" : "http";
+    const url = `${scheme}://localhost:${PORT}`;
+    console.log(`Server started at ${url}`);
+    // Auto open browser
+    if (devMode && OPEN_BROWSER_ON_START) {
```
LeoMcA: The request for this came from content - so we can't gate this feature behind `devMode`, as that's `false` when running in `content`:
```suggestion
    if (OPEN_BROWSER_ON_START) {
```

LeoMcA (approved):
The request came from the content team, so I think we're good to go after this change:

PR #807, opened 2025-09-19
chore: add basic browser testing for kitchensink page

LeoMcA:
Add basic testing of kitchensink page using https://webdriver.io/


argl (approved):
Assuming some more specific tests will be added to the kitchensink test, nice scaffolding!

PR #817, opened 2025-09-22
fix(writer-reload): add rate limiting

### LINKED ISSUE #776: Local server sometimes constantly refreshes

Sometimes the local server seems to be constantly reloading the page. It's really distracting.

The local server not to be constantly reloading the page.

Image

_No response_

LeoMcA:
Fixes https://github.com/mdn/fred/issues/776

Adds an exponential back-off to reloads if the first comparison each page load is different.

### DIFF
--- a/components/writer-reload/element.js
+++ b/components/writer-reload/element.js
@@ -1,12 +1,30 @@
-import { LitElement, nothing } from "lit";
+import { LitElement, html } from "lit";
+
+import styles from "./element.css?lit";
 
 /**
  * Custom element which reloads the page when the document has changed.
  */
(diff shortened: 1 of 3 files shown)


components/writer-reload/element.js
```diff
@@ -26,22 +44,33 @@ export default class MDNWriterReload extends LitElement {
     const res = await fetch(url);
     if (res.ok) {
       const state = await res.text();
+      // eslint-disable-next-line unicorn/no-negated-condition
```
LeoMcA, line 47: Opted to override eslint here and below because I think this order makes more sense - it follows the ordering of operations.

argl (approved):

Issue #487, opened 2025-07-31
Theme and language dropdown slightly overlap breadcrumbs bar


globau:

brysonbw:
Created PR in attempt to resolve: [#829](https://github.com/mdn/fred/pull/829)

caugner:
IIRC the overlap was intentional.

/cc @pepelsbey

pepelsbey:
> IIRC the overlap was intentional.

It wasn't. I would expect dropdown to open at the breadcrumbs bar line, not slightly over it.

pepelsbey:
The reason of the overlap is that the opening buttons don't take the full height of the breadcrumbs toolbar. It tried to solve it at some point but it was a little complicated and I haven't finished the fix.

brysonbw:
@pepelsbey and @caugner Following my previous PR, and after being made aware that the issue is caused by the buttons not taking the full height of the breadcrumbs bar, I created a new PR (#864 ) in attempt to resolve. I would appreciate any feedback or thoughts

LeoMcA:
Fixed in https://github.com/mdn/fred/pull/864, thanks @brysonbw!

PR #829, opened 2025-09-26
fix (ui): fix theme and language dropdown slightly overlap breadcrumb…

brysonbw:
#### Changes
- Fix `Theme` and `Language` dropdowns overlapping breadcrumbs bar on open

#### Before 

image

image

#### After

image

image

Linked Issues: #487

### DIFF
--- a/components/color-theme/element.css
+++ b/components/color-theme/element.css
@@ -66,7 +66,7 @@
   width: max-content;
 
   padding: 0.75rem;
-  margin: 0;
+  margin: 0.29rem 0 0;
 
   background-color: var(--color-background-primary);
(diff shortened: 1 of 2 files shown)


caugner (changes requested):
Thank you for taking the time to work on this issue. 🙌 

Unfortunately, setting a margin like this is not an acceptable solution.

We need to make the buttons take the full height of the breadcrumb bar, which appears to not be this trivial, see: https://github.com/mdn/fred/issues/487#issuecomment-3346110656

I'm going to close this PR.

brysonbw:
For sure, all smooth

Issue #489, opened 2025-07-31
About menu: Mark "MDN on github" link as an external link


globau:

caugner:
Originally fixed in https://github.com/mdn/fred/commit/49c18ecfcafdf8b9e321fa972f621d8929113168, but regressed in https://github.com/mdn/fred/commit/3ed0da0c1cd112394220307e535d7a79be5bd4e4#diff-2171d05f10cb211525de19c26509a318b14a22a0d3575fb347f410c539c975e7L758.

brysonbw:
Created PR in attempt to resolve: [#830](https://github.com/mdn/fred/pull/830)

mirunacurtean:
Closing this as the issue is fixed.

<img width="1797" height="452" alt="Image" src="https://github.com/user-attachments/assets/b951fcfb-c07f-479b-b88d-9e9625478204" />

PR #830, opened 2025-09-26
fix(menu): mark "MDN on GitHub" link as external

brysonbw:
#### Changes
- Apply external class to external links in menu

#### Before 

image

#### After 

image

- #489


brysonbw:
I resembled what we have for the same github link in the footer here:

image

If we want to mark it as external in a different way, that’s smooth - I just thought I’d start by creating a PR with my initial approach.

caugner:
> #### After
> image

In this screenshot, the external link icon does not look properly aligned (horizontally + vertically).

If this was as trivial as adding the "external" class, then we would have already done so. The CSS needs to be reworked carefully here, maintaining the gap to the icon on the left, without applying this gap to the external icon on the right.

brysonbw:
Ok, smooth. Thanks for the feedback

caugner:
I'm fixing the spacing issue in https://github.com/mdn/fred/pull/861. Once that landed, it should be fairly easy to add the external link indicator.

caugner (changes requested):
The menu rendering logic has completely changed (see the merge conflict).

Note that the "MDN on GitHub" link now already gets the English indicator in translated locales (see https://developer.mozilla.org/fr/), which is added via `::after`.

So when you add the `external` class, you'll probably need special handling that adds the external link indicator after the English indicator.

Converting to draft for now.

components/menu/server.js
```diff
@@ -109,6 +109,7 @@ export class Menu extends ServerComponent {
                                             item.icon && "menu__panel-icon",
                                             context.locale !== "en-US" &&
                                               "only-in-en-us",
+                                            item.external && "external",
```
caugner: I think we can just derive this from the URL:

```suggestion
                                            !item.href.startsWith("/") && "external",
```

components/menu/constants.js
```diff
@@ -415,6 +415,7 @@ export const TABS = [
             href: "https://github.com/mdn",
             text: "MDN on GitHub",
             icon: "github",
+            external: true,
```
caugner:
```suggestion
```

components/menu/types.d.ts
```diff
@@ -23,6 +23,7 @@ interface SlugItem extends BaseItem {
 interface HrefItem extends BaseItem {
   href: string;
   icon?: string;
+  external?: boolean;
```
caugner:
```suggestion
```

components/menu/server.css
```diff
@@ -3,3 +3,8 @@
 @import url("base.css");
 @import url("mobile.css");
 @import url("desktop.css");
+
+.external::after {
+  margin-bottom: 0.1em;
+  margin-left: 0.5em;
+}
```
caugner: This shouldn't be necessary, as we already have this:

https://github.com/mdn/fred/blob/e4a22f181561ae9f853a5947c5a072ca1f1d967e/components/menu/base.css#L159-L161

```suggestion
```

caugner (changes requested):
LGTM overall, but let's generalize.

Note that in French, the external indicator now replaces the `(angl.)` indicator:

- Before:
  image
- After:
  image

But I think that's actually fine, as we don't have the `(angl.)` indicator on external links in content either.

components/menu/server.js
caugner: Actually, let's avoid the `only-in-en-us` class for external links:

```suggestion
                                            (!item.href.startsWith("/") &&
                                              "external") ||
                                              (context.locale !== "en-US" &&
                                                "only-in-en-us"),
```

caugner (changes requested):

caugner (approved):

LeoMcA (approved):
Thank you! Let me just check that @mdn/content-team  approves this change before merging.

pepelsbey:
Makes sense and looks good!

PR #832, opened 2025-09-26
chore(lint): regenerate package-lock to check validity

LeoMcA:
Adds step to lint workflow to regenerate package-lock.json if it differs, to check unintended changes aren't included e.g. when changing deps with a dirty `node_modules` dir.

Draft while I test it works on github actions.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -23,6 +23,27 @@ jobs:
           cache: npm
           node-version-file: ".nvmrc"
 
+      - name: Checkout base package-lock.json
+        uses: actions/checkout@v5
+        with:
+          path: base
+          ref: ${{ github.base_ref }}
+          sparse-checkout: |
+            package-lock.json
+          sparse-checkout-cone-mode: false
+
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yml
```diff
@@ -23,6 +23,27 @@ jobs:
...
+          sparse-checkout-cone-mode: false
+
+      - name: Check package-lock.json is correct
+        run: |
+          if ! diff -q package-lock.json base/package-lock.json; then
+            mv package-lock.json package-lock.json.branch
+            mv base/package-lock.json .
+            rm -rf node_modules
+            npm install
+            diff package-lock.json package-lock.json.branch
```
caugner, line 42: If I understand this correctly, then this assumes that running `npm install` results in the same `package-lock.json`. However, if a PR adds a dependency (with a subtree of transitive dependencies), then one simultaneous release of one of those transitive dependencies (while the PR is open) might cause this workflow to fail.

I also think this might fail if the PR branch is behind `main`, afaik `github.base_ref` points to the latest commit on `main`.

Issue #757, opened 2025-09-12
Local server fails to start when there’s a space in the parent folder name


pepelsbey:
When `yarn start` is launched from the content folder inside a parent folder with a space in its name, the server fails to start. The same is true for `npm start` directly in fred.

Yes, “just rename the folder” could fix it, but in some cases, it’s not possible. For example, Apple’s iCloud documents are located in the `/Users/pepelsbey/Library/Mobile Documents/com~apple~CloudDocs` folder that has a space in it.

### ✅ `/Users/pepelsbey/MDN/content`

<img width="2940" height="1838" alt="Image" src="https://github.com/user-attachments/assets/37cda322-08da-4041-83dc-3e457c51d87e" />

### ❌ `/Users/pepelsbey/M D N/content`

<img width="2940" height="1838" alt="Image" src="https://github.com/user-attachments/assets/3d46794b-3e6c-42bd-bd49-181fc3f58661" />

<details>
<summary>Full error log</summary>

```sh
npx yarn start
yarn run v1.22.22
$ yarn -s info:fred && yarn up-to-date-check && env-cmd --silent cross-env CONTENT_ROOT=files NODE_ENV=production FRED_PLAYGROUND_LOCAL=true FRED_PLAYGROUND_PORT=5043 FRED_PORT=5042 FRED_WRITER_MODE=true fred-server

🐥 This command is now using fred: https://github.com/mdn/fred
🐞 Please report any issues here:  https://github.com/mdn/fred/issues/new?template=bug.yml

$ node scripts/up-to-date-check.js
[server] node:internal/modules/cjs/loader:1386
[server]   throw err;
[server]   ^
[server] 
[server] Error: Cannot find module '/Users/pepelsbey/M'
[server]     at Function._resolveFilename (node:internal/modules/cjs/loader:1383:15)
[server]     at defaultResolveImpl (node:internal/modules/cjs/loader:1025:19)
[server]     at resolveForCJSWithHooks (node:internal/modules/cjs/loader:1030:22)
[server]     at Function._load (node:internal/modules/cjs/loader:1192:37)
[server]     at TracingChannel.traceSync (node:diagnostics_channel:322:14)
[server]     at wrapModuleLoad (node:internal/modules/cjs/loader:237:24)
[server]     at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:171:5)
[server]     at node:internal/main/run_main_module:36:49 {
[server]   code: 'MODULE_NOT_FOUND',
[...]

PR #837, opened 2025-09-29
fix(server): quote server script path

brysonbw:
#### Changes
- Add quotes around server path to allow starting when parent folder has spaces

Linked Issues: #757


brysonbw:
Simply wrapping path in quotes comes to mind to resolve but when running sever with bin shorthand `fred-server` in `content` repo `start` and `start:fred` scripts - the quotes are not interpreted by the shell (at least from what I understand) so paths with spaces in the parent folder still break.

Likewise, replacing `fred-server` with `node ../fred/scripts/server.js` allows it to start successfully when running `yarn start` in `content` repo.

But this is just my opinion, leaving comment and marking PR as draft to see if someone thinks making changes to `content` repo `start` and `start:fred` scripts would be unnecessary? Thoughts @LeoMcA?

scripts/server.js
```diff
@@ -4,7 +4,7 @@ import { fileURLToPath } from "node:url";
 import { rariBin } from "@mdn/rari";
 import { concurrently } from "concurrently";
 
-export const server = fileURLToPath(import.meta.resolve("../server.js"));
+export const server = `"${fileURLToPath(import.meta.resolve("../server.js"))}"`;
```
caugner: I think it would be slightly better to quote the use below on line 12:

```diff
-command: `node ${server}`,
+command: `node "${server}"`,
```

caugner (commented):
Looks promising, but haven't tried it out yet.

scripts/server.js
brysonbw: Ok, smooth

LeoMcA (approved):
Good stuff, thank you! Tested across paths with/without a space, and locally in `content` after running `npm pack` and installing

caugner (approved):

Issue #848, opened 2025-10-01
Archive mdn/yari repository


caugner:
Tracking issue for effort to archive the mdn/yari repo.

caugner:
I have now updated Pull Requests permissions on the mdn/yari repository to "Collaborators only".

PR #852, opened 2025-10-02
feat(legacy): inline yari

caugner:
### Description

Inlines all parts of yari required for MDN Plus.

### Motivation

We want to archive the mdn/yari repository.

### Additional details

Preview:

- [Plus overview page](https://fred-pr852.review.mdn.allizom.net/en-US/plus)
- [Updates](https://fred-pr852.review.mdn.allizom.net/en-US/plus/updates)
- [AI Help](https://fred-pr852.review.mdn.allizom.net/en-US/plus/ai-help)
- [Settings](https://fred-pr852.review.mdn.allizom.net/en-US/plus/settings)

Part of https://github.com/mdn/fred/issues/848.


legacy/index.tsx
```diff
@@ -1,12 +1,12 @@
 import ReactDOM from "react-dom/client";
```
caugner, line 1: I now realize that no longer having `@mdn/yari` in our dependencies means this transitive dependency is no longer installed.

In theory, we could use `vendor/yari` as a local dependency, but that would break Dependabot updates.

rspack.config.js
```diff
@@ -455,19 +451,6 @@ const legacyConfig = merge(
...
-        filename: "index.[contenthash].html",
-        template: "node_modules/@mdn/yari/client/public/index.html",
-      }),
-      new RspackManifestPlugin({
-        fileName: "asset-manifest.json",
-        generate: (_seed, files) =>
-          files
-            .map((file) => file.path)
-            .filter((path) => !path.endsWith(".map")),
-      }),
```
caugner, line 470: @LeoMcA Do you remember what these were for?

LeoMcA: These are needed for offline mode to work:

https://github.com/mdn/fred/blob/9c3a2f908562213d70817835b00ac28c7b461596/vendor/yari/client/pwa/src/service-worker.ts#L41-L46

package.json
```diff
@@ -100,7 +100,6 @@
     "css-loader": "^7.1.4",
     "css-minimizer-webpack-plugin": "^8.0.0",
     "dexie": "^4.4.2",
```
LeoMcA, line 102: I think we can also drop dexie

package.json
```diff
@@ -118,7 +117,6 @@
     "postcss-preset-env": "^11.2.1",
     "prettier": "^3.8.3",
     "resolve-url-loader": "^5.0.0",
-    "rspack-manifest-plugin": "^5.2.1",
```
LeoMcA, line 121: Let's add it back for offline mode

package.json
LeoMcA, line 99: I think we can drop this too

vendor/yari/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## What's here
+
+- **`client/src/`** — React/Lit source for the legacy bundle. Fred's [`legacy/index.tsx`](../../legacy/index.tsx) renders `<Plus />` from this tree. Subset kept: `plus/`, `ui/`, `document/`, `lit/`, `playground/`, `settings/`, `telemetry/`, `homepage/static-page/` (used by Plus docs), plus shared providers (`user-context.tsx`, `ui-context.tsx`, `placement-context.tsx`) and utilities. Yari's full SPA router and the homepage/blog/community/curriculum/observatory/etc. routes have all been removed.
+- **`client/pwa/src/`** — service worker source, used by the rspack `service-worker` entry in [rspack.config.js](../../rspack.config.js). Note: the offline pre-cache install handler currently no-ops because it expects a manifest format and path that fred doesn't emit. Fixing that would require additional work (see "Known limitations" below).
+- **`client/public/assets/`** — illustration/icon SVGs and PNGs referenced by Plus React components.
+- **`libs/play/`** — playground runner used at runtime by fred's [server.js](../../server.js) (`handleRunner`). Not bundled by rspack.
```
LeoMcA, line 12: Might be worth pointing out this was pre-existing (and we should come up with a better solution now it's in dex!)

rspack.config.js
```diff
@@ -418,7 +417,6 @@ const legacyConfig = merge(
     name: "legacy",
     entry: {
       index: "./legacy/index.tsx",
-      yari: "./node_modules/@mdn/yari/client/src/index.tsx",
```
LeoMcA, line 421: Ah yeah, no: this is a problem - we need this file (which has been cleaned up from the vendored code) in order to render the offline JSONs we get from the differy zip.

LeoMcA (changes requested):
Sorry, should've thought a bit more about this before beginning to review (or even encouraging you to do it!) - we need pretty much all of yari to render offline mode.

We could put everything back, but - even though it's a bit of a silly reason - I'm not sure I want to clutter up the git tree like that, especially with stuff we won't be wanting to edit.

Perhaps we should keep using the npm yari for now to render offline? That has the added advantage that we won't accidentally break something being used for that, and allows us to modify/slim/convert as much of the vendored code as we want.

Longer term it would be fun to get fred's server components running in-browser, and I have some ideas how to achieve that - so it doesn't have to be this way forever.

rspack.config.js
caugner: So if the differy zip would contain HTML files, would that solve our issues?

LeoMcA: Yes, though I reckon it would be quite large, and prone to frequent re-downloads (every dependency update would change some aspect of each page). We would perhaps also have to bundle the static assets in (HTML/CSS): we could cache them directly from the site itself, but they could diverge from the ones used in the differy build.

My gut says getting fred server components to run in browser would be easier (and quite interesting!)

Easiest would just be restoring this line and continuing to build the "offline yari" from npm.

caugner: > Easiest would just be restoring this line and continuing to build the "offline yari" from npm.

The benefit is very small if we still keep `@mdn/yari` as a dependency, unless it moves to `devDependencies`?

LeoMcA: yeah we could put it in dev or optional

LeoMcA: > My gut says getting fred server components to run in browser would be easier (and quite interesting!)

Very unrefined, but here's my <1hr hacked together attempt at rendering fred server components in the browser - it works! https://github.com/mdn/fred/pull/1546

caugner:
Closing for now, we can reopen later.

PR #853, opened 2025-10-02
chore(survey): add developer survey 2025

### LINKED ISSUE #849: Run developer survey (implement banner)
Task issue for running our developer survey on MDN

* To be run initially for two weeks (6th -19 October)
* 30% of the audience of the following directories: glossary, learn and web
* The banner will link out to the survey rather than integrated (like short surveys) - link to be provided
* The text in the banner should read "We’re running a survey to understand how you use MDN in your web development work. We’d love for you to take 10 minutes to share your feedback."

LeoMcA:
add support for linking to surveys, simplify force logic, fix layout issue on short pages

fixes https://github.com/mdn/fred/issues/849

added "needs content decision" because I think the movement of the survey below the description paragraph(s) was an unintentional change from yari -> fred, posted some images to slack to determine where we want to place it

I have some ideas about simplifying this logic even further to make adding surveys easier, and migrating some logic server-side/inline to avoid the content shift - but testing this is a bit of a pain, so I'd like to work on adding proper support for component-based testing first - and none of this should block this survey.

Screenshot:

Screen Shot 2025-10-03 at 12 30 45

### DIFF
--- a/components/reference-layout/server.css
+++ b/components/reference-layout/server.css
@@ -6,7 +6,7 @@
   grid-template-areas:
     "sidebar . header . toc"
     "sidebar . body   . toc";
-  grid-template-rows: min-content auto;
+  grid-template-rows: min-content 1fr;
   grid-template-columns: var(--layout-2-sidebars);
 
   justify-content: space-between;
--- a/components/reference-layout/server.js
+++ b/components/reference-layout/server.js
@@ -26,9 +26,10 @@ export class ReferenceLayout extends ServerComponent {
           <div class="reference-layout__header">
             ${WRITER_MODE ? WriterToolbar.render(context) : nothing}
             ${TranslationBanner.render(context)}
-            <mdn-survey></mdn-survey>
             <h1>${doc.title}</h1>
-            ${BaselineIndicator.render(context)} ${description}
+            ${BaselineIndicator.render(context)}
+            <mdn-survey></mdn-survey>
+            ${description}
           </div>
           <aside class="reference-layout__toc">
             ${ReferenceToc.render(context)}
--- a/components/reference-layout/server.js
+++ b/components/reference-layout/server.js
@@ -26,6 +26,7 @@ export class ReferenceLayout extends ServerComponent {
           <div class="reference-layout__header">
             ${WRITER_MODE ? WriterToolbar.render(context) : nothing}
             ${TranslationBanner.render(context)}
+            <mdn-survey></mdn-survey>
             <h1>${doc.title}</h1>
             ${BaselineIndicator.render(context)} ${description}
           </div>
@@ -34,7 +35,6 @@ export class ReferenceLayout extends ServerComponent {
             <mdn-placement-sidebar></mdn-placement-sidebar>
           </aside>
           <div class="reference-layout__body">
-            <mdn-survey></mdn-survey>
             ${sections} ${ArticleFooter.render(context)}
           </div>
         </main>
--- a/components/survey/element.css
+++ b/components/survey/element.css
@@ -9,7 +9,6 @@
 
   padding: 1rem;
   padding-left: 5rem;
-  margin-top: 1.6rem;
 
   background-color: var(--color-background-purple);
   border-radius: 0.5rem;
--- a/components/survey/element.css
+++ b/components/survey/element.css
@@ -48,7 +48,8 @@ mdn-button::part(icon) {
   color: var(--color-text-purple);
 }
 
-summary {
+summary,
+a {
   color: var(--color-text-purple);
   cursor: pointer;
 }
--- a/components/survey/element.css
+++ b/components/survey/element.css
@@ -1,3 +1,5 @@
+@import url("../external-link/global.css");
+
 :host {
   display: block;
 }
--- a/components/survey/element.css
+++ b/components/survey/element.css
@@ -9,6 +9,7 @@
 
   padding: 1rem;
   padding-left: 5rem;
+  margin-top: 1.6rem;
 
   background-color: var(--color-background-purple);
   border-radius: 0.5rem;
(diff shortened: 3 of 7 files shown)


components/survey/element.css
```diff
@@ -48,7 +48,8 @@ mdn-button::part(icon) {
   color: var(--color-text-purple);
 }
 
-summary {
+summary,
+a {
```
caugner: Nit: The link has no hover state (i.e. underline should disappear on hover, or it should be a button to increase engagement rate).

components/reference-layout/server.css
```diff
@@ -6,7 +6,7 @@
   grid-template-areas:
     "sidebar . header . toc"
     "sidebar . body   . toc";
-  grid-template-rows: min-content auto;
+  grid-template-rows: min-content 1fr;
```
caugner, line 9: Can you explain the effect of this?

components/reference-layout/server.css
caugner, line 1: Move survey up in reference-layout if this is definitely desired.

components/reference-layout/server.css
LeoMcA: It fixes this weird void space after the description:

Screen Shot 2025-10-03 at 12 12 20

After `auto -> 1fr`:

Screen Shot 2025-10-03 at 12 12 34

components/survey/element.css
```diff
@@ -9,6 +9,7 @@
 
   padding: 1rem;
   padding-left: 5rem;
+  margin-top: 1.6rem;
```
caugner: Nit: How do we get to 1.6rem, why not 1.5rem?

components/survey/element.js
```diff
@@ -121,22 +113,32 @@ export class MDNSurvey extends L10nMixin(LitElement) {
...
-      writeSurveyState(this._survey.bucket, this._surveyState);
-      this.#measure("opened");
+      this.#markOpened();
       this._isOpen = true;
       this.requestUpdate();
     }
   }
 
+  #markOpened() {
+    if (!this._survey || !this._surveyState || this._isOpen) return;
```
LeoMcA:
```suggestion
    if (!this._survey || !this._surveyState) return;
```

nit: not sure that `this._isOpen` check is necessary since we already test it in `#onToggle`, and it's a bit confusing to return if open when we're marking opened (especially when we don't modify that state here)

LeoMcA (commented):
Nice changes, thanks: approved with one nit

components/survey/element.js
caugner: In that case we should probably check it in `#onLinkClick` though.

Edit: In a way, I think it makes sense not to mark it opened when it is already open, so I don't think this condition is wrong.

components/survey/element.js
```diff
@@ -208,17 +209,25 @@ export class MDNSurvey extends L10nMixin(LitElement) {
...
-                <iframe
-                  title=${ifDefined(this._survey.question)}
-                  src=${this._source}
-                ></iframe>
-              `
-            : nothing}
-        </details>
+        ${this._survey.link
+          ? html`<a
+              href=${this._source}
```
caugner, line 215: Should probably mark this as an external link:
```suggestion
              class="external"
              href=${this._source}
```

caugner (approved):

components/reference-layout/server.js
```diff
@@ -26,6 +26,7 @@ export class ReferenceLayout extends ServerComponent {
           <div class="reference-layout__header">
             ${WRITER_MODE ? WriterToolbar.render(context) : nothing}
             ${TranslationBanner.render(context)}
+            <mdn-survey></mdn-survey>
             <h1>${doc.title}</h1>
```
caugner: It used to be shown below the header, see: https://github.com/mdn/yari/blob/0cfdd1b8445c0c6983ff4f2ce6375ae32308047a/client/src/document/index.tsx#L255

```suggestion
            <h1>${doc.title}</h1>
            <mdn-survey></mdn-survey>
```

components/survey/element.js
```diff
@@ -70,20 +66,16 @@ export class MDNSurvey extends L10nMixin(LitElement) {
    * @returns {Survey.Survey | undefined}
    */
   #findSurvey() {
+    const forcedSurvey = new URLSearchParams(location.search).get(
```
caugner, line 69: What was the reason to move from hash to search?

```suggestion
    const forcedSurvey = new URLSearchParams(location.hash).get(
```

components/survey/element.js
LeoMcA: But we never set `this._isOpen` when clicking the link, so checking it is a no-op in `#onLinkClick`.

This means that every link click sends a glean ping, but only the first summary click *per page load* sends a glean ping.

components/survey/element.css
LeoMcA: 1.5rem is the top margin of the title, but since its line height isn't 1, there's internal padding between the top of the character and the bounding box, so 1.5rem makes the banner look a bit lopsided visually: 1.6rem looked best to my eye, balancing the padding with aligning with the sidebar filter to the left and "In this article" heading to the right

components/survey/element.js
LeoMcA: appending a hash on an already-loaded page doesn't reload it, so the banner doesn't appear: also this feels more param-y than hash-y to me - I could imagine similar logic for forcing some other condition, and we'd want to be able to force every combination of A and B not just the exclusive ones

components/survey/element.js
caugner: > But we never set this._isOpen when clicking the link

We do now, but this would just avoid capturing the clicked event twice on one page load, which is inconsistent with the details toggling, where we capture each opening. I'll apply your suggested change.

components/survey/element.js
caugner: > I could imagine similar logic for forcing some other condition, and we'd want to be able to force every combination of A and B not just the exclusive ones

If you parse the hash as URLSearchParams, that shouldn't be a problem, but I guess in the end this is fine.

caugner (approved):
Final look:
image

PR #858, opened 2025-10-03
fix(content-section): preserve white space only in pre, not code

### LINKED ISSUE #847: regression: HTML whitespace rendered in code blocks due to white-space:pre-wrap
PR #418 added `white-space: pre-wrap` to all `code` elements in `.content-section`. Whitespace in `<code>` is not significant, and before this change it was collapsed like HTML whitespace anywhere else. Now with `white-space: pre-wrap`, every single whitespace character is rendered visually:

Example: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/browser_action#:~:text=browser_style
Source: https://github.com/mdn/content/blob/7e1296fc0722c86fb7e15487b5e9626597c7a2a0/files/en-us/mozilla/add-ons/webextensions/manifest.json/browser_action/index.md#L73-L77

Screenshot showing extraneous whitespace around the browser_style text

I recommend to revert the change to `code` from PR #418 to fix this issue.
Or at the very least add a selector to avoid matching `<code>` elements that have children.

caugner:
### Description

Preserve white space only in `pre`, not `code`.

### Motivation

Content uses non-significant white space in `<code>` without `<pre>`, and this was rendered on MDN.

Fixes https://github.com/mdn/fred/issues/847.

### DIFF
--- a/components/content-section/server.css
+++ b/components/content-section/server.css
@@ -82,8 +82,11 @@
   pre,
   code {
     font-family: var(--font-family-code);
-    white-space: pre-wrap;
     background-color: var(--color-background-secondary);
+  }
+
+  pre {
+    white-space: pre-wrap;
   }
 
   pre:not(.code-example pre) {


components/content-section/server.css
```diff
@@ -82,10 +82,13 @@
   pre,
   code {
     font-family: var(--font-family-code);
-    white-space: pre-wrap;
```
caugner, line 85: Note: GitHub still preserves spaces (but not breaks) in inline Markdown code:

```css
.markdown-body code {
  white-space: break-spaces;
}
```

argl (approved):

PR #871, opened 2025-10-06
fix(site-search): add divider between result preview snippets

### LINKED ISSUE #859: Broken summary for CORS
### Proposal

I am not sure what is going on here but something is getting stitched together incorrectly.

Image

argl:
### Description

Fixes the assembly of resutl highlight snippets with a divider.

### Motivation

Assembly was broken with missing whitespace

### Additional details

image

Fixes #859

### DIFF
--- a/components/site-search/element.css
+++ b/components/site-search/element.css
@@ -209,6 +209,10 @@
     color: inherit;
     background-color: var(--color-highlight-background);
   }
+
+  .divider {
+    color: var(--color-text-secondary);
+  }
 }
 
 .site-search-results__locale-indicator {
--- a/components/site-search/element.js
+++ b/components/site-search/element.js
@@ -361,7 +361,11 @@ export class MDNSiteSearch extends L10nMixin(LitElement) {
                         <p class="site-search-results__description">
                           ${result.highlight.body &&
                           result.highlight.body.length > 0
-                            ? result.highlight.body.map((b) => unsafeHTML(b))
+                            ? unsafeHTML(
+                                result.highlight.body.join(
+                                  ' <span class="divider">…</span> ',
+                                ),
+                              )
                             : result.summary}
                         </p>
                       </article>
--- a/components/site-search/element.js
+++ b/components/site-search/element.js
@@ -1,6 +1,7 @@
 import { Task } from "@lit/task";
 import { LitElement, html, nothing } from "lit";
 
+import { join } from "lit/directives/join.js";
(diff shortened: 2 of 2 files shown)


components/site-search/element.js
```diff
@@ -361,7 +361,11 @@ export class MDNSiteSearch extends L10nMixin(LitElement) {
                         <p class="site-search-results__description">
                           ${result.highlight.body &&
                           result.highlight.body.length > 0
-                            ? result.highlight.body.map((b) => unsafeHTML(b))
+                            ? result.highlight.body.flatMap((b, i) =>
+                                i === 0
+                                  ? [unsafeHTML(b)]
+                                  : [html` … `, unsafeHTML(b)],
+                              )
```
caugner: I think it would be cleaner to use Lit's [`join` directive](https://lit.dev/docs/templates/directives/#join):

```suggestion
                            ? join(result.highlight.body.map((b) => unsafeHTML(b)), html` … `),
                              )
```

components/site-search/element.css
```diff
@@ -209,6 +209,10 @@
     color: inherit;
     background-color: var(--color-highlight-background);
   }
+
+  .divider {
+    color: var(--color-text-secondary);
+  }
```
caugner, line 215: This doesn't seem to be used. Leftover from an earlier version?

```suggestion
```

caugner (changes requested):

components/site-search/element.css
argl: Not a leftover but a half-done styling clue. See latest version.

caugner (approved):

components/site-search/element.js
```diff
@@ -361,7 +362,10 @@ export class MDNSiteSearch extends L10nMixin(LitElement) {
                         <p class="site-search-results__description">
                           ${result.highlight.body &&
                           result.highlight.body.length > 0
-                            ? result.highlight.body.map((b) => unsafeHTML(b))
+                            ? join(
+                                result.highlight.body.map((b) => unsafeHTML(b)),
+                                html`<span class="divider"> … </span>`,
```
caugner, line 367: Note: An alternative would be to render this as an actual list, and add the divider with CSS. But I'm not sure how much better this really would be from the point of view of a screen reader.

argl: Thought about that, but bending lists into the current design seems moot, and the joining with n-1 dividers seems cumbersome with lists. I think the current approach is as sound as it can be, without rethinking the results design.

Issue #507, opened 2025-08-05
Announce the current theme on the theme switcher


pepelsbey:
The icon changes with the selected theme, but the button says “Theme” no matter what.

https://github.com/mdn/fred/blob/ad3302872d4974585e03b14bcb7defcb43e91962/components/color-theme/element.js#L73

globau:
Thanks Vadim, are you able to talk to the use-case for this change?

pepelsbey:
It's an accessibility issue: visually we can tell what theme is currently selected (the icon changes), but this info is not conveyed on the button. Basically, we've built a select that doesn't say what option is selected. It's not the case in the language switcher right next to it.

globau:
ty; while this issue exists, it doesn't seem to be significant given there are other signals when the theme is changed - putting into the backlog.

PR #876, opened 2025-10-07
fix(color-theme): show current theme in dropdown button

brysonbw:
#### Changes
- show current theme on the color theme toggle button

#### Before

image

#### After

https://github.com/user-attachments/assets/d9c926c7-b16c-4eee-96a6-58c05d52ea54

#507


caugner:
@pepelsbey Is this solution what you had in mind?

pepelsbey:
Thank you your help! Unfortunately, we might need to scale back the implementation a little. In the language switcher case, the language name is sufficient to represent what this dropdown is for. I'm afraid in the theme switcher's case, "Light", "Dark", or "OS default" might not look clear enough. Let's keep the "Theme" label and the appropriate icon next to it. In the original issue I filed, [I pointed out](https://github.com/mdn/fred/issues/507#issuecomment-3161411996) that the current value is not available for screen readers, and this is the problem we need to address.

brysonbw:
For sure, all smooth

caugner:
Converting to draft awaiting @pepelsbey's feedback to be addressed.

Issue #873, opened 2025-10-06
With JavaScript disabled, the BCD should ideally work


danielhjacobs:
### What information was incorrect, unhelpful, or incomplete?

Prior to the yari redesign to utilize React, the BCD table actually worked with JS disabled.

### What did you expect to see?

Ideally the BCD table would work.

globau:
We should note that JS is required.

danielhjacobs:
Ok, opened PR, for the BCD not working with JS disabled issue, see https://github.com/mdn/yari/issues/3653

danielhjacobs:
It would be nice for the BCD to work without JS like it did years ago.

danielhjacobs:
I opened https://github.com/mdn/fred/pull/890 to fix this issue.

danielhjacobs:
As https://github.com/mdn/yari/issues/3653 was closed as a duplicate of this issue, I changed the wording on the PR that would have closed this issue to stop it from doing so, as that issue requested the BCD explicitly be made not to use JS, and I also think if possible that would be ideal.

PR #890, opened 2025-10-08
fix(compat-table): show message when JavaScript is disabled

danielhjacobs:
### Description

Addresses #873

### Motivation

So if JS is disabled there is a message

JS enabled (same as current):

https://github.com/user-attachments/assets/cb557f5d-9abd-4d0a-baad-a7dde207db79

JS disabled:

image


components/compat-table-lazy/element.js
```diff
@@ -96,7 +97,11 @@ export class MDNCompatTableLazy extends L10nMixin(LitElement) {
 
   render() {
     return this._dataTask.render({
-      initial: () => html`<p>${this.l10n("compat-loading")`Loading…`}</p>`,
+      initial: () =>
+        html`<p>${this.l10n("compat-loading")`Loading…`}</p>
```
danielhjacobs: The formatting here has a functional impact, but the commit is reverting my changes to it. This needs to be exactly like this:
```js
        html`<p>${this.l10n("compat-loading")`Loading…`}</p>${
          unsafeHTML(
            `<noscript><style>p:first-of-type { display: none !important; }</style><div>${this.l10n("compat-js-required")`Browser compatibility tables only load in the browser with JavaScript enabled.`}</div></noscript>`,
          )}`,
```
But the commit is moving `${` to the same line as `unsafeHTML(`. That causes the Loading... message to not disappear when the table loads in.

danielhjacobs: Had no choice but to use `// prettier-ignore`, made it less ugly by also putting the `unsafeHTML` in a variable.

Note: Had to use `unsafeHTML` because lit complains about any HTML in a `<noscript>` tag, likely since it doesn't render with JS enabled.

components/compat-table-lazy/element.js
```diff
@@ -95,8 +96,13 @@ export class MDNCompatTableLazy extends L10nMixin(LitElement) {
   });
 
   render() {
+    const noScript = unsafeHTML(
+      `<noscript><style>p:first-of-type:not(noscript p) { display: none !important; }</style><p>${this.l10n("compat-js-required")`Browser compatibility tables only load in the browser with JavaScript enabled.`}</p></noscript>`,
+    );
     return this._dataTask.render({
-      initial: () => html`<p>${this.l10n("compat-loading")`Loading…`}</p>`,
+      initial: () =>
+        // prettier-ignore
```
caugner: What is the prettier formatting issue here?

components/compat-table-lazy/element.js
```diff
@@ -95,8 +96,13 @@ export class MDNCompatTableLazy extends L10nMixin(LitElement) {
   });
 
   render() {
+    const noScript = unsafeHTML(
+      `<noscript><style>p:first-of-type:not(noscript p) { display: none !important; }</style><p>${this.l10n("compat-js-required")`Browser compatibility tables only load in the browser with JavaScript enabled.`}</p></noscript>`,
```
caugner: Why do we need that CSS?

As for the message, I would suggest making it more concise and actionable:

```suggestion
      `<noscript><style>p:first-of-type:not(noscript p) { display: none !important; }</style><p>${this.l10n("compat-js-required")`Enable JavaScript to view this browser compatibility table.`}</p></noscript>`,
```

l10n/en-US.ftl
```diff
@@ -61,6 +61,7 @@ obs-mdn = The { -brand-name-obs } provides effective security insights, guided b
 
 
 compat-loading = Loading…
+compat-js-required = Browser compatibility tables only load in the browser with JavaScript enabled.
```
caugner:
```suggestion
compat-js-required = Enable JavaScript to view this browser compatibility table.
```

components/compat-table-lazy/element.js
danielhjacobs: The formatting here has a functional impact, but the commit is reverting my changes to it.

This needs to stay on a single line, up to here at least:

```html`<p>${this.l10n("compat-loading")`Loading…`}</p>${```

If it doesn't, the message does not get replaced when the compatibility table loads, and instead the message "Loading..." continues showing with the compatibility table beneath it.

danielhjacobs: See https://github.com/mdn/fred/pull/890#discussion_r2420180104

components/compat-table-lazy/element.js
danielhjacobs: The CSS hides the message "Loading...". Otherwise, that message would show with the noscript message beneath it. There's no such thing as `<yesscript>`, after all.

components/compat-table-lazy/element.js
danielhjacobs: I think it's to do with Lit's update logic. If it sees an element, then a text node, then an element, it treats that differently than just two elements. Do you have a suggestion on how to change that? I think there are a few ways, but they're all a lot more complex than just ignoring prettier for a line.

components/compat-table-lazy/element.js
danielhjacobs: https://stackoverflow.com/a/431554/30680220

danielhjacobs: Message updated.

danielhjacobs:
Added comments regarding unaddressed review notes to document why I did not address them.

components/compat-table-lazy/element.js
caugner: I think we could just render `<noscript><p>...</p></noscript>` in `initial`. Technically, we're not loading at this stage yet.

danielhjacobs: Ok, but for what it's worth. I tried adding this before the fetch in _dataTask's task, and what I discovered is this component does not enter the pending state as fetch is running for even a moment:

`await new Promise((resolve) => setTimeout(resolve, 10000));`

With that code immediately above `const response = await fetch(`, for a full ten seconds, whatever is in `initial` shows, so if that's the `<noscript>`, nothing will show with JS enabled.

danielhjacobs:
Addressed the review comments, removing the CSS and the prettier ignore directive.

As noted in one of the review threads, I don't think this component ever actually enters the pending state, seemingly going straight from initial to complete. I guess maybe that's fine, but it does mean the Loading... message seems to no longer appear as the page first loads.

caugner (approved):

LeoMcA (approved):
Nice, thank you!

Issue #485, opened 2025-07-31
Module orange underline for Core module extends to Extensions


globau:

brysonbw:
Just to confirm - this refers to the sidebar? If so, we don't want an orange underline for `Extension modules` title or sub list items on hover?

<img width="452" height="102" alt="Image" src="https://github.com/user-attachments/assets/305d3843-3939-4fb7-bfcb-0359ad500c61" />

mirunacurtean:
@brysonbw This was initially an issue reported by QA, so I can provide more context on this. 
It refers to the Curriculum homepage [Modules](https://developer.mozilla.org/en-US/curriculum/#modules) sections underline. 

<img width="1630" height="657" alt="Image" src="https://github.com/user-attachments/assets/71489293-5c9c-4876-a9a7-103f253a2502" />

brysonbw:
Ok, smooth. Thanks

PR #898, opened 2025-10-12
fix(curriculum): fix core module orange underline extending to extens…

brysonbw:
#### Changes
- Fix core module orange underline extending to extensions tab

#### Before

https://github.com/user-attachments/assets/d17958e4-784c-4901-ae87-09aa1771f04f

#### After

https://github.com/user-attachments/assets/7ee387e6-ba16-43e4-b8c8-c04fa2bd79eb

- #485


caugner (changes requested):
Thanks for looking into this. Now that the review deployment works on fred PRs, I finally took a look: https://fred-pr898.review.mdn.allizom.net/en-US/curriculum/#modules

Unfortunately, it doesn't look right for the "Extensions" tab yet:
image

Maybe the [pre-Fred version of this page in the Web Archive](https://web.archive.org/web/20250801113253/https://developer.mozilla.org/en-US/curriculum/#modules) gives indications of what we're missing?

PR #905, opened 2025-10-13
feat(telemetry): port Google Analytics

### LINKED ISSUE #877: GA on Fred
Let's add GA on the new frontend of MDN. It will help us make data driven decisions for the new features and content. 

- Include GA tracking snippet in all the MDN pages 
- Capture page views and sessions and path funnel for the user 
- Track site errors like 404 events 
- Capture referrers and clicks to external sites 
- Create a GA dashboard segmented by geography, device and content category 

Note: Ensure the tracking complies with MDN privacy policy and does not collect personal identifiers.

caugner:
### Description

Ports Google Analytics measurement from Yari.

### Motivation

Supports us to make data driven decisions for new features and content.

Fixes https://github.com/mdn/fred/issues/877.

### DIFF
--- a/hooks/ga-init.js
+++ b/hooks/ga-init.js
@@ -7,14 +7,9 @@
 
 import { GA_ENABLED, GA_MEASUREMENT_ID } from "../components/env/index.js";
 import { dntEnabled } from "../utils/dnt-helper.js";
+import { userIsOptedOut } from "../utils/telemetry-opt-out.js";
 
-const FIRST_PARTY_DATA_OPT_OUT_COOKIE_NAME = "moz-1st-party-data-opt-out";
-
-const userIsOptedOut = document.cookie
-  .split("; ")
-  .includes(`${FIRST_PARTY_DATA_OPT_OUT_COOKIE_NAME}=true`);
-
-if (GA_ENABLED && GA_MEASUREMENT_ID && !userIsOptedOut && !dntEnabled()) {
+if (GA_ENABLED && GA_MEASUREMENT_ID && !userIsOptedOut() && !dntEnabled()) {
   // @ts-expect-error - added by GA
   globalThis.dataLayer = globalThis.dataLayer || [];
 
--- a/hooks/ga-init.js
+++ b/hooks/ga-init.js
@@ -1,0 +1,36 @@
+/**
+ * Google Analytics, enabled if:
+ * 1. FRED_GA_ENABLED and FRED_GA_MEASUREMENT_ID are set
+ * 2. User has not opted out via moz-1st-party-data-opt-out cookie
+ * 3. User has not enabled Do Not Track (DNT)
+ */
+
+import { GA_ENABLED, GA_MEASUREMENT_ID } from "../components/env/index.js";
+import { dntEnabled } from "../utils/dnt-helper.js";
+
+const FIRST_PARTY_DATA_OPT_OUT_COOKIE_NAME = "moz-1st-party-data-opt-out";
+
+const userIsOptedOut = document.cookie
+  .split("; ")
+  .includes(`${FIRST_PARTY_DATA_OPT_OUT_COOKIE_NAME}=true`);
+
+if (GA_ENABLED && GA_MEASUREMENT_ID && !userIsOptedOut && !dntEnabled()) {
+  // @ts-expect-error - added by GA
+  globalThis.dataLayer = globalThis.dataLayer || [];
+
+  /** @param {...any} args */
+  function gtag(...args) {
+    // @ts-expect-error - added by GA
+    globalThis.dataLayer.push(args);
+  }
+
+  gtag("js", new Date());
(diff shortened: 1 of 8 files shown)


utils/dnt-helper.js
LeoMcA, line 1: I presume this is (based on) the general dnt-helper used across Mozilla sites for time immemorial to filter out those IE versions where DNT was on by default? If there's an upstream source like bedrock or moz-central it might be good to leave it in a comment to aid (or discourage!) future changes

LeoMcA (commented):
Nothing jumps out at me, but I'd like to have a proper poke at it tomorrow morning before giving explicit approval

utils/dnt-helper.js
caugner: Yes, it's a copy of https://github.com/mdn/yari/blob/main/ssr/mozilla.dnthelper.min.js.

There is https://www.npmjs.com/package/@mozmeao/dnt-helper (see https://unpkg.com/@mozmeao/dnt-helper@1.0.0/index.js), but I'm not a fan of pulling in a dependency for a single function that won't change.

hooks/ga-init.js
```diff
@@ -0,0 +1,31 @@
+/**
+ * Google Analytics, enabled if:
+ * 1. FRED_GA_ENABLED and FRED_GA_MEASUREMENT_ID are set
+ * 2. User has not opted out via moz-1st-party-data-opt-out cookie
+ * 3. User has not enabled Do Not Track (DNT)
```
LeoMcA, line 5: In the future it might be nice to expand this to include the GPC header/api too

utils/dnt-helper.js
LeoMcA: > I'm not a fan of pulling in a dependency for a single function that won't change

Agreed, I [did the same](https://github.com/mdn/fred/blob/main/hooks/dialog-closedby.js) with a simple polyfill in fred.

utils/dnt-helper.js
```diff
@@ -0,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/**
+ * Checks if Do Not Track (DNT) is enabled.
+ * @param {string} [dnt] - DNT value to check (for testing)
+ * @param {string} [ua] - User agent string (for testing)
+ * @returns {boolean} True if DNT is enabled, false otherwise
+ */
```
LeoMcA, line 10: Copying the comprehensive jsdoc from https://github.com/mozmeao/dnt-helper/blob/main/src/mozilla-dnt-helper.js and adding a link to it:

```suggestion
/**
 * Returns true or false based on whether doNotTrack is enabled. It also takes into account the
 * anomalies, such as !bugzilla 887703, which effect versions of Fx 31 and lower. It also handles
 * IE versions on Windows 7, 8 and 8.1, where the DNT implementation does not honor the spec.
 * Based on https://github.com/mozmeao/dnt-helper/blob/main/src/mozilla-dnt-helper.js
 * @see https://bugzilla.mozilla.org/show_bug.cgi?id=1217896 for more details
 * @params {string} [dnt] - An optional mock doNotTrack string to ease unit testing.
 * @params {string} [ua] - An optional mock userAgent string to ease unit testing.
 * @returns {boolean} true if enabled else false
 */
```

LeoMcA (approved):
Good stuff, consider all comments nits

utils/dnt-helper.js
caugner: FWIW, the dnt-helper originated in https://bugzilla.mozilla.org/show_bug.cgi?id=1217896, and https://github.com/mozilla/bedrock/blob/3c0e227e9f9d95a26e77d6ee08a697db72d74f42/media/js/base/dnt-helper.js is the original version that landed in bedrock.

caugner: Note: Added more comments in 63839e25a64b9795a616b99d3c3b5486efb9620a.

utils/dnt-helper.js
```diff
@@ -0,0 +1,48 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
```
caugner: Given that all of Fred code is MPL, we probably don't need this.


```suggestion
```

utils/telemetry-opt-out.js
```diff
@@ -0,0 +1,12 @@
+const FIRST_PARTY_DATA_OPT_OUT_COOKIE_NAME = "moz-1st-party-data-opt-out";
+
+/**
+ * Checks if user has opted out of first-party data collection.
+ * @param {string} [cookie] - Cookie value to check (for testing)
```
caugner:
```suggestion
 * @param {string} [cookie] - An optional mock cookie string to ease unit testing.
```

PR #922, opened 2025-10-15
fix(live-sample-result): use code hash as subdomain

### LINKED ISSUE #724: LiveSampleLink macro links point to HTTP 403
STR:

1. Go to https://developer.mozilla.org/en-US/docs/Web/API/Window/innerWidth#result
2. Click on _view the results of the demo code in a separate page_.

Expected: Interstitial asking user to confirm.

Actual: HTTP 403 error page.

(Similar to https://github.com/mdn/fred/issues/697, but not the same.)

caugner:
### Description

- Adds `permalink` option to the `play-runner` component.
- Uses this option in the `live-sample-result` component.

### Motivation

Ensures the runner URL can be opened in a new tab, without yielding a HTTP 403.

### Additional details

Compare

- https://fred-pr922.review.mdn.allizom.net/en-US/docs/Web/API/Window/innerWidth#result
- https://fred.review.mdn.allizom.net/en-US/docs/Web/API/Window/innerWidth#result

Fixes #724.

### DIFF
--- a/components/play-runner/element.js
+++ b/components/play-runner/element.js
@@ -109,16 +109,17 @@ export class MDNPlayRunner extends LitElement {
       );
       const prefix = (srcPrefix || "").replace(/\/$/, "");
       signal.throwIfAborted();
-      this._subdomain = permalink ? hash : this._uuid;
+      const subdomain = permalink ? hash : this._uuid;
       const url = new URL(
         `${prefix}/runner.html`,
         PLAYGROUND_LOCAL
           ? location.origin.replace(PORT.toString(), PLAYGROUND_PORT.toString())
-          : `${location.protocol}//${this._subdomain}.${PLAYGROUND_BASE_HOST}`,
+          : `${location.protocol}//${subdomain}.${PLAYGROUND_BASE_HOST}`,
       );
       // pass the uuid for postMessage isolation
-      url.searchParams.set("uuid", this._subdomain);
+      url.searchParams.set("uuid", subdomain);
       url.searchParams.set("state", state);
+      this._subdomain = subdomain;
       this._src = url.href;
       this.dispatchEvent(
         new CustomEvent("mdn-play-runner-src", {


caugner:
@LeoMcA Do you know what might cause the error `private names aren't valid in this context`?

LeoMcA:
@caugner yes, it appears to come from [this example](https://fred-pr922.review.mdn.allizom.net/en-US/docs/MDN/Kitchensink#rect_code) and bad escaping, you'll see the JS example contains lines like `if (rect1.x > rect2.x + rect2.w &#x26;&#x26;` with `&` escaped to `&#x26` - it's weird that this only occurs in the test in this PR though, considering it's also visible [on prod](https://developer.mozilla.org/en-US/docs/MDN/Kitchensink#rect_code) - perhaps something about our test environment wasn't allowing these examples to run prior to this change

components/play-runner/element.js
```diff
@@ -48,7 +49,12 @@ export class MDNPlayRunner extends LitElement {
     this.allow = undefined;
     /** @type {string | undefined} */
     this.sandbox = undefined;
-    this._subdomain = crypto.randomUUID();
+    /** @type {boolean} */
+    this.permalink = false;
+    /** @type {string} */
+    this._hash = crypto.randomUUID();
```
LeoMcA: This isn't the hash, so the naming's a bit confused, perhaps:
```suggestion
    this._uuid = crypto.randomUUID();
```
or simply keep:
```suggestion
    this._subdomain = crypto.randomUUID();
```
and conditionally overwrite it later

components/play-runner/element.js
```diff
@@ -102,7 +109,7 @@ export class MDNPlayRunner extends LitElement {
       );
       const prefix = (srcPrefix || "").replace(/\/$/, "");
       signal.throwIfAborted();
-      // We're using a random subdomain for origin isolation.
+      this._subdomain = permalink ? hash : this._hash;
```
LeoMcA: Though I *do* like the replacement of this comment with actual code, so perhaps:

```suggestion
      const subdomain = permalink ? hash : this._uuid;
```

caugner: So we need to remember the current subdomain in order to only handle messages from that host.

Furthermore, if we were to toggle `permanent` on and then off (unlikely but possible), then the current implementation would use the same subdomain before and after.

I had first implemented the random yet stable subdomain as the opt-in behaviour, assuming this is only needed for CSS choice examples possibly to support the transition, but then inverted the property to keep this scoped to fix the issue.

caugner:
> @caugner yes, it appears to come from [this example](https://fred-pr922.review.mdn.allizom.net/en-US/docs/MDN/Kitchensink#rect_code) and bad escaping, you'll see the JS example contains lines like `if (rect1.x > rect2.x + rect2.w &#x26;&#x26;` with `&` escaped to `&#x26` - it's weird that this only occurs in the test in this PR though, considering it's also visible [on prod](https://developer.mozilla.org/en-US/docs/MDN/Kitchensink#rect_code) - perhaps something about our test environment wasn't allowing these examples to run prior to this change

@LeoMcA Thanks, good catch, I opened https://github.com/mdn/content/pull/41559.

components/play-runner/element.js
LeoMcA: > So we need to remember the current subdomain in order to only handle messages from that host.

So we do! I had ctrl-clicked on `this._subdomain` expecting it to give me the references, but it gave me the definitions.

Issue #923, opened 2025-10-16
Can not run `yarn start` command on Windows


yin1999:
### What information was incorrect, unhelpful, or incomplete?

With `@mdn/fred@v1.6.2` installed on Windows, I can never successfully run `yarn start` command (in `mdn/content` folder), and I got the TypeError: `The argument 'filename' must be a file URL object, file URL string, or absolute path string. Received 'file:///home/runner/work/fred/fred/node_modules/fdir/dist/index.mjs'`:

<img width="1436" height="419" alt="Image" src="https://github.com/user-attachments/assets/a8cdf678-1c9f-4a49-8bd5-06316949c5b6" />

When I downgrade the version to `v1.6.1`, I could successfully run `yarn start` command:

<img width="755" height="429" alt="Image" src="https://github.com/user-attachments/assets/adcb7dbb-4548-4610-84df-1c92bb8e7bf0" />

### What did you expect to see?

Make fred compatible with Windows operating system

globau:
Confirming.

caugner:
This sounds like a regression caused by https://github.com/mdn/fred/pull/895.

LeoMcA:
Results of my initial debugging: yes https://github.com/mdn/fred/pull/895 causes the issue, but this issue is particularly pernicious because it only appears when the npm package is *built* on Linux and *run* on Windows: the code in the PR works fine when built and run on Windows. Looking into a fix now.

PR #925, opened 2025-10-16
ci(npm-publish-sim): run tests across linux, mac and windows

LeoMcA:
### Description

Run the kitchensink tests across windows and linux when doing an npm publish simulation.

### Motivation

Avoid bugs like https://github.com/mdn/fred/issues/923, by automatically testing changes to fred within content on Windows.

Running kitchensink tests because my powershell scripting wasn't up to task for keeping the current curl-based tests, and this simplifies the workflow: and opens up the possibility of easily testing things like the writers tools themselves in future.

Currently fails, as it should: we have a bug. Will pass when https://github.com/mdn/fred/pull/926 is merged.


.github/workflows/npm-publish-simulation.yml
```diff
@@ -38,8 +38,23 @@ jobs:
...
+        with:
+          name: tarball
+          path: mdn/fred/${{ steps.build.outputs.tarball }}
+          compression-level: 0 # file already compressed
+
+  test:
+    needs: build
+    strategy:
+      matrix:
+        os: [ubuntu-latest, windows-latest]
```
caugner: Nit: Slight preference for the following.

```suggestion
        os:
          - ubuntu-latest
          - windows-latest
```

PS: Why not also test on macOS?

.github/workflows/npm-publish-simulation.yml
```diff
@@ -54,41 +69,49 @@ jobs:
...
+      - name: (mdn/content) Download tarball
+        uses: actions/download-artifact@v5
+        with:
+          name: tarball
+
       - name: (mdn/content) Install tarball
         working-directory: mdn/content
-        env:
-          TARBALL: ${{ steps.build.outputs.tarball }}
+        shell: bash
```
caugner, line 82: Alternatively, set globally?

If I understand https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#jobsjob_idstepsshell correctly, then `sh` is the default for Linux. Why `bash`? Assuming `bash` is available on Windows, shouldn't `sh` as well?

```yml
defaults:
  run:
    shell: bash
```

wdio.conf.js
```diff
@@ -18,13 +20,13 @@ export const config = {
...
-  baseUrl: "http://localhost:3000/",
+  baseUrl: `http://localhost:${FRED_PORT}/`,
   async before(_, __, browser) {
     console.log("waiting for servers to start");
     await browser.waitUntil(
       async () => {
         try {
-          await browser.url("http://localhost:3000");
+          await browser.url(`http://localhost:${FRED_PORT}`);
           await browser.url("http://localhost:8083");
```
caugner, line 30: Nit: While we're here, also add an env variable for the Playground port?

.github/workflows/npm-publish-simulation.yml
LeoMcA: The docs suggest `sh` isn't available on Windows, `bash` is the only `sh`-like shell available there. I only use it in this command because this is the only one which has `$variable` substitution: all other steps I want to run in powershell in windows to catch any possible compatibility errors there.

.github/workflows/npm-publish-simulation.yml
LeoMcA: Yes, I suppose we should test on mac too: my thought was that there shouldn't be many differences between linux/mac, but I'm sure we'll be able to find some obscure edge case bug soon!

wdio.conf.js
LeoMcA: `8083` is rari, and it shouldn't change (unlike fred which has a different default in content/the npm package): we can add one in the future if we need it.

PR #926, opened 2025-10-16
fix(windows): remove fdir from ssr bundle

LeoMcA:
### Description

Split SSR runtime utils from SSR buildtime utils.

### Motivation

We were bundling fdir into the SSR code, which isn't required: and causing an error on Windows *when the npm package was built on Linux*.

### Additional details

Tested locally building the package in WSL with `npm pack` and installing it in content in Windows. (Also established this was a fair test, as building `main` in WSL caused the same error when installing in content in Windows).

Also experimenting with CI to catch errors like this (build in Linux, run on Windows) in the future: https://github.com/mdn/fred/pull/925 - still a WIP.

Fixes https://github.com/mdn/fred/issues/923

### DIFF
--- a/utils/name-transformation.js
+++ b/utils/name-transformation.js
@@ -1,7 +1,16 @@
 const ACRONYMS = new Set(["MDN", "IX"]);
 
-/** @param {string} name */
-export function toPascalCase(name) {
+/**
+ * Converts a kebab-case string to PascalCase,
+ * useful for converting a web component tag name to its
+ * associated class name.
+ *
+ * e.g. `mdn-custom-element` -> `MDNCustomElement`
+ *
+ * @param {string} name kebab-case string
+ * @returns {string} PascalCase string
+ */
+export function kebabToPascalCase(name) {
   return name
     .split("-")
     .map((word) => {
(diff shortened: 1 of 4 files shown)


utils/name-transformation.js
```diff
@@ -0,0 +1,22 @@
...
+    .join("");
+}
+
+/** @param {string} name */
+export function toCamelCase(name) {
+  return name
+    .replaceAll(/([a-z0-9])([A-Z])/g, "$1-$2")
+    .replaceAll(/([A-Z])([A-Z][a-z])/g, "$1-$2")
+    .toLowerCase();
+}
```
caugner: Nit: Can we add examples in JSDoc? It is not quite clear to me what these are doing, despite their name.

LeoMcA: That would probably be because `toCamelCase` is named totally wrong: whoops! I've updated the names and added more to the jsdoc - good catch.

LeoMcA:
> Does this mean fdir can be moved to devDependencies?

no, because we use `fdir` (through the `crawl()`) function in `build/ssr.js` which we expose through the `fred-ssr` binary

LeoMcA:
> no windows test instance here to actually run it

https://github.com/mdn/fred/pull/925 should prove that it works after I merge this, then merge main back in! :crossed_fingers:

hamishwillee:
@LeoMcA Thanks so much for this. I'll be testing on the next update of mdn/content, probably Monday.

Issue #924, opened 2025-10-16
Ensure workflows have permissions defined


caugner:
Ensure that all workflows across the MDN org have `permissions:` defined, either for every job or at the workflow-level.

See: [MDN workflows without permissions:](https://github.com/search?q=org%3Amdn+NOT+is%3Aarchived+path%3A.github%2Fworkflows%2F**.yml+NOT+%2Fpermissions%3A%2F&type=code)

https://github.com/pulls?q=org%3Amdn+is%3Apr+"assign+explicit+permissions"+author%3Acaugner+created%3A<2025-10-16:

- https://github.com/mdn/browser-compat-data/pull/25793
- https://github.com/mdn/content/pull/38017

caugner:
> See: [MDN workflows without permissions:](https://github.com/search?q=org%3Amdn+NOT+is%3Aarchived+path%3A.github%2Fworkflows%2F**.yml+NOT+%2Fpermissions%3A%2F&type=code)

✅ There are no longer any workflows without permissions.

PR #929, opened 2025-10-16
ci(workflows): assign explicit permissions

caugner:
### Description

Adds explicit `permissions:` configuration to workflow files that don't already have permissions defined either at the workflow level or for all jobs.

- If the workflow doesn't use `secrets.GITHUB_TOKEN` or `github.token`, sets `permissions: {}` to restrict all permissions.
- If the workflow uses the GitHub token, adds `permissions:` with required permissions.

### Motivation

Security best practice to explicitly declare `GITHUB_TOKEN` permissions instead of relying on default permissions, following the principle of least privilege by ensuring workflows only have the permissions they actually need.

### Additional details

See: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

Part of https://github.com/mdn/fred/issues/924.

### DIFF
--- a/.github/workflows/main-review-companion.yml
+++ b/.github/workflows/main-review-companion.yml
@@ -9,7 +9,6 @@ on:
 
   workflow_dispatch:
 
-# Same as: _deploy.yml
 permissions:
   # Authenticate with GCP.
   id-token: write


.github/workflows/main-review-companion.yml
```diff
@@ -9,6 +9,13 @@ on:
 
   workflow_dispatch:
 
+# Same as: _deploy.yml
+permissions:
+  # Post comment in pull request.
+  pull-requests: write
```
caugner:
```suggestion
```

.github/workflows/main-review-companion.yml
```diff
@@ -9,6 +9,11 @@ on:
 
   workflow_dispatch:
 
+# Same as: _deploy.yml
```
caugner:
```suggestion
```

LeoMcA (approved):

PR #936, opened 2025-10-17
fix(button): replace random `aria-labelledby` ids with `aria-label` + `labelId`

### LINKED ISSUE #913: Use of `randomIdString()` impacts deployment time
We use `randomIdString()` in several places, which causes files to be different in every build. During deployment all files have changed and need to be replaced in our bucket, increasing deployment time to 55 minutes.

We should find an alternative.

caugner:
### Description

Update button labeling to use `aria-label` for string labels and a fixed internal ID for slotted labels in each custom element's shadow root.

### Motivation

Make rendered button markup deterministic while preserving accessible labels and avoiding collisions between buttons in the same tree.

### Additional details

Keep random IDs as a fallback for template labels without an explicit ID. Dropdown ID handling remains unchanged.

`Button` in light DOM (server components) was the only `randomIdString()` caller whose output reached the build: `mdn-dropdown` assigns its ID in `updated()`, which Lit SSR never runs, and `interactive-example` opts out of SSR via `static ssr = false`.

The only remaining build-to-build difference is in `/play/` pages, where attribute bindings carry lit-html's template marker (`lit$…$`), which is seeded with `Math.random()` at module load. See the follow-up issue draft in the comments.

Add 12 SSR regression cases covering repeated string labels, distinct template labels, and slotted labels in separate shadow roots, for buttons and links in both regular and icon-only modes. The cases check label references, ID uniqueness within a tree, and deterministic output where IDs are fixed or unnecessary.

Related to #913.


components/dropdown/element.js
```diff
@@ -70,7 +73,7 @@ export class MDNDropdown extends LitElement {
   _setAriaAttributes() {
     let id = this._dropdownSlotElements.find((element) => element.id)?.id;
     if (!id) {
-      id = randomIdString("uid_");
+      id = createElementId("dropdown", this.name);
```
LeoMcA, line 76: This seems unnecessary, why not construct the string directly?

components/utils/index.js
```diff
@@ -1,9 +1,31 @@
 /**
- * Used to generate a random element id by combining a prefix with a random string.
+ * Simple FNV-1a hash implementation for generating deterministic IDs.
+ * @param {string} str - String to hash
+ * @returns {string} - Hexadecimal hash string
+ */
+function hashString(str) {
```
LeoMcA, line 6: Not sure what hashing gives us here: if the input isn't unique, the output won't be either - this seems unnecessary

components/utils/index.js
```diff
@@ -1,9 +1,31 @@
...
+ * Falls back to random generation if no content is provided (for backwards compatibility).
  *
- * @param {string} prefix
+ * @param {string} prefix - Prefix for the ID
+ * @param {string} content - Content to hash for ID generation
  * @returns {string}
  */
-export function randomIdString(prefix = "id-") {
-  return Math.random().toString(36).replace("0.", prefix);
+export function createElementId(prefix, content) {
```
LeoMcA, line 25: This function itself seems unnecessary: where we can set a static, unique ID in call sites, let's do that instead of routing it through here.

components/dropdown/element.js
```diff
@@ -22,12 +23,14 @@ export class MDNDropdown extends LitElement {
   static styles = styles;
 
   static properties = {
+    name: { type: String, required: true },
```
LeoMcA, line 26: `required: true` doesn't do anything

components/button/pure.js
```diff
@@ -28,7 +28,10 @@ export default function Button({
   variant = "primary",
   action,
 }) {
-  const labelId = randomIdString("label-");
+  const labelId = createElementId(
+    "label",
+    `${typeof label === "string" ? label : "btn"}-${href || ""}`,
```
LeoMcA, line 33: This is likely to cause collisions: we can realistically have two buttons with the same href on the same page.

We only need this ID to set `aria-labelledby` for icon-only buttons. Perhaps we set `aria-label` for icon-only buttons instead, and enforce that `label` is a string with types when `iconOnly` is true?

LeoMcA (changes requested):
I think this fixes everything in the wrong place:

We're sacrificing the guarantee of ID-uniqueness in a number of places, which would result in invalid HTML, and a confused accessibility tree, for an issue which IIRC we already have a fix for: just upload everything on each build, no rsync/diff - that's the slow bit.

Regardless, having deterministic output is a good goal, but I don't think this is at all the correct approach: the new function does very little that manually adding `id`s would do, and the addition of `name` in dropdowns isn't enforced, nor is necessary - we can just add `ids` to the slotted elements directly in the DOM.

I'd suggest we keep the random ID fallback where we can't be sure that manual IDs have been set, and enforce this deterministic requirement in CI so we catch future fallbacks to the random ID and fix.

caugner:
Follow-up issue draft for the remaining Playground non-determinism:

> **Title:** `lit$…$` markers make Playground HTML non-deterministic across builds
>
> After #936, the only remaining build-to-build difference in the SSR output is in `/play/` pages, where attribute bindings on `mdn-play-runner` and `mdn-modal` carry lit-html's template marker (e.g. `lit$391985138$0` vs `lit$070035930$0`). lit-html seeds that marker with `Math.random()` at module load, so every SSR process produces different attribute values.
>
> **Expected:** identical HTML on repeated builds of unchanged content, so the deployment sync skips these files.
>
> **Possible approaches:** find out why the marker leaks into the rendered output on these pages (it should be consumed during template preparation), or strip it in post-processing. Independently, add a CI job that runs the SSR step twice and fails on any diff, to catch future regressions.
>
> Related to #913.

caugner:
FYI @LeoMcA The scope of this PR has changed following your review.

PR #942, opened 2025-10-20
chore(menu): improve Learn items

chrisdavidmills:
### Description

This PR is part of a next-step plan for MDN learn, to make the curriculum a bit more prominent and help it integrate better with Learn.

This PR in particular improves the top-level nav menu items in the Learn menu. Currently, I feel that they don't really represent the content that we most want to point people to:

- The "Common questions" articles are not articles that we want to point people to as a top priority, and they are not part of the Frontend developer course, so I want to include the core module entry page instead.
- Both Learn HTML items basically point to the same place, so I decided to just include one link to the module to learn HTML.
- Both Learn CSS items basically point to the same place, so I decided to include two links to the most important modules to learn CSS.
- The Learn JavaScript link choices were quite frankly bizarre. I decided it is much better to just include one link to the module to learn JavaScript.


Rumyra:
@caugner I remember there was a reason we used the same article title in the link text as we have in the url - was it for l10n?

caugner:
> @caugner I remember there was a reason we used the same article title in the link text as we have in the url - was it for l10n?

No, I don't think the link text needs to be equal to the article title.

caugner (approved):

caugner:
(Awaiting content decision before merging.)

chrisdavidmills:
> (Awaiting content decision before merging.)

What decision is that? I'm just checking to see if you are waiting on me for something, as it is not clear.

Rumyra:
@chrisdavidmills I need to approve - I just wanted to check the above with @caugner first

chrisdavidmills:
> @chrisdavidmills I need to approve - I just wanted to check the above with @caugner first

Ah, OK, cool.

Can you merge it?

Issue #801, opened 2025-09-18
Search dialog input has unexpected autocompletions


caugner:
<img width="848" height="491" alt="Image" src="https://github.com/user-attachments/assets/4eabcc28-5b1e-4840-8b5c-324c79229c29" />

PR #993, opened 2025-10-26
fix(search-modal): disable autocompletions on search input

brysonbw:
### Changes
- Add `autocomplete="off"` to search input

### Before

https://github.com/user-attachments/assets/a8993a29-a8be-444f-8768-c6327b3048d8

### After

https://github.com/user-attachments/assets/156dde7a-53a2-40e8-b76d-5a6294b49280

- #801

### DIFF
--- a/components/search-modal/element.js
+++ b/components/search-modal/element.js
@@ -245,6 +245,7 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
             type="search"
             name="q"
             .value=${this._query}
+            autocomplete="off"
             autofocus
             @input=${this._input}
             placeholder=${this.l10n`Search`}


brysonbw:
@caugner Is this what you were envisioning?

caugner:
> @caugner Is this what you were envisioning?

I was hoping that setting something like [`autocomplete="off"`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/autocomplete#off) on the input field would be sufficient to resolve the issue.

brysonbw:
I guess I looked too much into it lol. All smooth, I set `autocomplete="off"` in changes, so I'll remove all other changes

components/search-modal/element.js
```diff
@@ -244,6 +244,7 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
           <input
             type="search"
             name="q"
+            autocomplete="off"
             .value=${this._query}
```
caugner: Nit: Let's move it next to `autofocus`.
```suggestion
            .value=${this._query}
            autocomplete="off"
```

argl (approved):

Issue #947, opened 2025-10-21
(design) Keyboard keys are not represented with CSS


tristantheb:
### What information was incorrect, unhelpful, or incomplete?

The design surrounding the keyboard keys in the documentation has disappeared in recent versions, making the text difficult for users to digest. We talk about keyboard keys, but they are not distinguishable from the rest of the text.

<img width="759" height="113" alt="Image" src="https://github.com/user-attachments/assets/7b9e92b6-a992-471f-9d3b-cb61941e44a0" />

### What did you expect to see?

This helps differentiate between the keys on the keyboard, as pages that use a lot of representations of these keys can become difficult to digest.

<img width="758" height="170" alt="Image" src="https://github.com/user-attachments/assets/2fe34897-5ad0-4b03-a0bb-a63fde21d2cb" />

### Do you have any supporting links, references, or citations?

Page used for screens: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key

DIY design of example: 

<img width="327" height="161" alt="Image" src="https://github.com/user-attachments/assets/1976f3c4-153f-4579-b493-4e3eb7714628" />

### Do you have anything more you want to share?

This issue is related to another design disappearance for the “good” and ‘bad’ examples when code blocks have a “brush: plain.”

globau:
@tristantheb thank you for the report.

@mdn/content-team - is this an intentional change with the new UI, or an accidental regression?

Rumyra:
An accidental regression - we can update the styles to be like the previous for this.

caugner:
Yari [had this `kbd` style](https://github.com/mdn/yari/blob/4e5a00ed9a72f5534d9e818298f76a0c5c2e6970/client/src/document/index.scss#L878-L886):

```css
kbd {
  border: 2px solid var(--border-secondary);
  border-radius: var(--elem-radius);
  box-shadow: var(--border-secondary);
  box-shadow: inset 0 -1px 0 0 var(--border-secondary);
  color: var(--text-secondary);
  font-size: 0.825rem;
  padding: 0.25rem;
}
```

PR #1003, opened 2025-10-27
fix(content-section): add `<kbd>` style

caugner:
### Description

Styles `<kbd>` elements in content.

Like yari, but:
- Regular font size, not smaller.
- Thinner border, for consistency.

### Motivation

Parity with yari.

### Yari

image

### Fred (before)

image

### Fred (after)

image

Fixes https://github.com/mdn/fred/issues/947.

### DIFF
--- a/components/content-section/server.css
+++ b/components/content-section/server.css
@@ -117,20 +117,22 @@
     }
   }
 
-  code,
-  kbd {
+  code {
     padding: 0.125em 0.25em;
-    border-radius: 0.25em;
+    border-radius: var(--radius-normal);
   }
 
   kbd {
+    padding: 0.0625em 0.125em;
+
     font-family: var(--font-family-code);
 
     color: var(--color-text-secondary);
 
-    background-color: var(--color-background-page);
+    background-color: var(--color-background-primary);
     border: 1px solid var(--color-border-primary);
-    box-shadow: inset 0 -1px 0 0 var(--color-border-primary);
+    border-bottom-width: 2px;
+    border-radius: var(--radius-normal);
   }
 
   img {


caugner:
@mdn/content-team I reduced the border width, to avoid adjacent `<kbd>`s to touch. Does that look right?

Note:
- In Yari, list items (`<li>`) had additional `0.5rem` top/bottom margin (see [here](https://github.com/mdn/yari/blob/4e5a00ed9a72f5534d9e818298f76a0c5c2e6970/client/src/document/index.scss#L108-L114)).
- In Fred, list items do not have additional margin.)

components/content-section/server.css
```diff
@@ -117,11 +117,22 @@
...
 
+  kbd {
+    font-family: var(--font-family-code);
+
+    color: var(--color-text-secondary);
+
+    background-color: var(--color-background-page);
+    border: 1px solid var(--color-border-primary);
+    box-shadow: inset 0 -1px 0 0 var(--color-border-primary);
+  }
```
pepelsbey: I would suggest some changes:

1. Use a background, not just a border. I need to file a separate issue about `--color-background-page` used like `.green { color: red }` 🥲
2. Make vertical padding smaller: there’s no need for it to be the same as for code, as it doesn’t have underlines. It helps with the vertical spacing in lists.
3. Use border-width instead of box-shadow for better visual vertical alignment
4. Use the border-radius token.

Image

image

```suggestion
  code {
    padding: 0.125em 0.25em;
    border-radius: var(--radius-normal);
  }

  kbd {
    padding: 0.0625em 0.125em;

    font-family: var(--font-family-code);

    color: var(--color-text-secondary);

    background-color: var(--color-background-primary);
    border: 1px solid var(--color-border-primary);
    border-bottom-width: 2px;
    border-radius: var(--radius-normal);
  }
```

caugner: Thanks!

> 1. Use a background, not just a border. I need to file a separate issue about `--color-background-page` used like `.green { color: red }` 🥲

Good catch, that was a left-over from (temporarily) adding `kbd` to the `code, pre` rule above.

> 2\. Make vertical padding smaller: there’s no need for it to be the same as for code, as it doesn’t have underlines. It helps with the vertical spacing in lists.

Good idea, I hadn't thought about the underline.


> 3\. Use border-width instead of box-shadow for better visual vertical alignment

I had originally done this, but I believe `border-width` moves the content slightly up. Probably not noticeable though.

> 4\. Use the border-radius token.

TIL, will apply all these changes tomorrow.

pepelsbey (approved):
Perfect! Thank you :)

PR #1023, opened 2025-10-29
chore(menu): update Layout Cookbook slugs

dipikabh:
### Description

As part of the CSS reorganization, we've moved `css/layout_book` to `css/how_to/layout_cookbook` via https://github.com/mdn/content/pull/41735.

Are there any other slug instances in fred that need to be updated?


dipikabh:
Thank you!

Issue #1027, opened 2025-10-30
Homepage: Contributor spotlight section overflows


caugner:
STR:

1. Open https://developer.mozilla.org/en-US/#:~:text=Contributor%20Spotlight on mobile.

<img width="414" height="172" alt="Image" src="https://github.com/user-attachments/assets/3b218e93-42f0-42fb-a01a-a40c4e8e2d5e" />

globau:
Fixed by bf722088ee43f746bece188fc92c63066f420255

PR #1045, opened 2025-10-31
fix(homepage-contributor-spotlight): fix content overflow

brysonbw:
#### Changes

- Edit `homepage-contributor-spotlight__name` class styles - prevent name content overflow on mobile screens

### Before

https://github.com/user-attachments/assets/2de6bb60-cf99-4ab0-8164-c5e533becdfe

### After

https://github.com/user-attachments/assets/08627481-457d-4a9e-ba01-ce86c29d882d

- #1027

### DIFF
--- a/components/homepage-contributor-spotlight/server.css
+++ b/components/homepage-contributor-spotlight/server.css
@@ -21,8 +21,7 @@
 }
 
 .homepage-contributor-spotlight__name {
-  width: max-content;
-  max-width: 100%;
+  width: fit-content;
 
   padding: 0.125em 0.25em;
 


components/homepage-contributor-spotlight/server.css
```diff
@@ -21,7 +21,7 @@
 }
 
 .homepage-contributor-spotlight__name {
-  width: max-content;
+  max-width: 100%;
```
caugner: I would think we should keep `width: max-content`, so it takes as much space as it needs, but without overflowing.

```suggestion
  max-width: 100%;
  width: max-content;
```

LeoMcA: I believe `width: max-content` is what's causing it to still overflow on certain screen sizes:

Image

I think what we actually need is just `width: fit-content`, no `max-width`

LeoMcA (approved):

LeoMcA:
Thank you for fixing this @brysonbw! Apologies for taking so long to review it, @caugner recently put some work in to get the [review companion](https://github.com/mdn/fred/pull/1045#issuecomment-3822453769) to work on PRs from forks, which makes reviewing much more convenient for us - so we ought to be much faster at it from now on!

brysonbw:
> Thank you for fixing this @brysonbw! Apologies for taking so long to review it, @caugner recently put some work in to get the [review companion](https://github.com/mdn/fred/pull/1045#issuecomment-3822453769) to work on PRs from forks, which makes reviewing much more convenient for us - so we ought to be much faster at it from now on!

Np and ok, smooth

PR #1072, opened 2025-11-11
feat(ServerComponent): add method to render simplified html

LeoMcA:
### Description

- Adds a `renderSimple` method to the server component class, which can be overwritten in a component to define a simplified HTML version of the component
- Does this for the Doc component to render simply the page contents, no sidebars etc.
- Doesn't render the outer layout either 

### Motivation

- Main motivation is to provide a better API than the hacky one used in https://github.com/mdn/mcp/pull/10
  - With the `renderSimple` export in `entry.ssr.js` I don't have to set up async local storage, mock fluent, etc.
  - Will also easily allow us to define simple forms of other components, e.g. Baseline or BCD, rather than maintaining a custom implementation divorced from the "real" component
- This could also prove useful for other purposes we might want to explore in the future: an alternative offline option, embedding in devtools, etc.

### Testing

For a preview of the simple form of a component, run:

```
FRED_SIMPLE_HTML=true npm start
```

And then navigate directly to a doc page (other page types will render nothing), e.g. http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isSafeInteger

#### Screenshot

Screen Shot 2025-11-11 at 17 17 58

### DIFF
--- a/components/server/async-local-storage.js
+++ b/components/server/async-local-storage.js
@@ -6,6 +6,8 @@ import { AsyncLocalStorage } from "node:async_hooks";
  * Generally only used within the `ServerComponent` class itself,
  * or very special server components (such as the `OuterLayout`).
  *
+ * e.g. We add the rspack compilation stats for use in `OuterLayout`.
+ *
  * Populated in `entry.ssr.js`.
  *
  * @type {AsyncLocalStorage<import("./types.js").FredLocalContents>}


components/doc/server.js
```diff
@@ -5,8 +5,16 @@ import { ServerComponent } from "../server/index.js";
...
+  render(context) {
     context.pageTitle = context.doc.pageTitle;
     return PageLayout.render(context, ReferenceLayout.render(context));
   }
+
+  /**
+   * @param {import("@fred").Context<import("@rari").DocPage>} context
+   */
+  renderSimple(context) {
+    return ReferenceLayout.render(context);
```
caugner, line 18: Did you mean to call `renderSimple` on it?

```suggestion
    return ReferenceLayout.renderSimple(context);
```

And why not create a `Plain` or `PlainLayout` component?

```suggestion
    return PlainLayout.render(context);
```

components/reference-layout/server.js
```diff
@@ -44,4 +44,18 @@ export class ReferenceLayout extends ServerComponent {
...
+  renderSimple(context) {
+    const { doc } = context;
+    const sections =
+      doc.body?.map((section) => ContentSection.render(context, section)) || [];
+
+    return html`
+      <h1>${doc.title}</h1>
+      ${BaselineIndicator.render(context)} ${sections}
+    `;
+  }
```
caugner, line 60: Shouldn't this be a separate layout, e.g. `PagelessLayout`?

build/server-worker.js
```diff
@@ -28,10 +28,12 @@ if (!indexModulePath) {
 try {
   /** @type {import("../entry.ssr.js")} */
   const indexModule = await import(pathToFileURL(indexModulePath).href);
-  const html = await indexModule?.render(reqPath, context, {
-    client: compilationStats.find((x) => x.name === "client") || {},
-    legacy: compilationStats.find((x) => x.name === "legacy") || {},
-  });
+  const html = process.env.FRED_SIMPLE_HTML
+    ? `<!doctype html><meta charset="UTF-8">${await indexModule?.renderSimple(reqPath, context)}`
```
caugner: Should this be called `FRED_PLAIN_HTML` maybe?

components/outer-layout/server.js
```diff
@@ -23,7 +23,7 @@ export class OuterLayout extends ServerComponent {
    */
   render(context, markup) {
     const asyncStore = asyncLocalStorage.getStore();
-    if (!asyncStore) {
+    if (!asyncStore || "renderSimple" in asyncStore) {
```
caugner: Why would `renderSimple` end up in asyncStore?

caugner (commented):
Could we avoid `renderSimple()` if an environment variable just caused this effect?

components/doc/server.js
LeoMcA: > Did you mean to call renderSimple on it?

There isn't a static `renderSimple` method on server components, only the instance method (which the static `render` method calls): I did this for consistency/simplicity: some components don't need a `renderSimple` method because their `render` method is simple enough - it would be a little weird to have to swap/change the use of `render` to `renderSimple` in parent components if you then implemented one in the child component. This way simple components can be nested within non-simple within simple without each having to be aware of what the other is doing.

> And why not create a Plain or PlainLayout component?

Let me think about this one a little bit: my original intent was to have the same component hierarchy amongst complex and simple rendering - so no purely simple components, they're always a different "view" on a complex component. I think in future there might be the potential for options on the simple view, i.e. (give me CSS, give me images, give me sidebars, etc.): if that's the case, again I think it would be simpler to have the same components across all those views.

components/outer-layout/server.js
LeoMcA: It wouldn't in the outer layout, since we don't call it from the `renderSimple` root function: this is simply to make typescript happy.

components/reference-layout/server.js
LeoMcA: Same response as above, here especially options on the simple layout (e.g. sidebars/no sidebars) I think would make using the same components preferable.

LeoMcA (commented):
> Could we avoid renderSimple() if an environment variable just caused this effect?

Hmm, let me think about this: I think though it would make the component code more complex, and I think there's a nice simplicity to just having two render methods on components which need them.

components/server/index.js
```diff
@@ -73,4 +80,12 @@ export class ServerComponent {
   render(..._args) {
     throw new Error("Must be implemented by subclass");
   }
+
+  /**
+   * @param {...any} args
+   * @returns {any}
+   */
```
caugner, line 87: nit: add description of this method

components/doc/server.js
```diff
@@ -5,8 +5,16 @@ import { ServerComponent } from "../server/index.js";
...
-   */ render(context) {
+   */
+  render(context) {
     context.pageTitle = context.doc.pageTitle;
     return PageLayout.render(context, ReferenceLayout.render(context));
   }
+
+  /**
+   * @param {import("@fred").Context<import("@rari").DocPage>} context
+   */
```
caugner, line 16: nit: explain what this method omits (even if it's comparably trivial from looking at render/renderSimple side-by-side

components/reference-layout/server.js
```diff
@@ -44,4 +44,18 @@ export class ReferenceLayout extends ServerComponent {
       </div>
     `;
   }
+
+  /**
+   * @param {import("@fred").Context<import("@rari").DocPage>} context
+   */
```
caugner, line 50: nit: explain what this simplified/minimal rendering omits

components/doc/server.js
caugner: Sounds right.

components/outer-layout/server.js
caugner: I realize that I don't fully understand what `asyncLocalStorage` holds.

I looked at https://github.com/mdn/fred/blob/22628366e1b9ecc0d8f36811492c5a6f08e62db4/components/server/async-local-storage.js and it doesn't have any information, it just creates an instance of `AsyncLocalStorage` without any semantic notion of what it contains.

It is not clear to me why `"renderSimple" in asyncStore` implies that `asyncLocalStorage is missing`. It would probably be better to throw a separate error in this case, because if `asyncStore` is truthy, it's apparently not missing.

LeoMcA: Yeah it's a little bit esoteric: the top level `render` methods (in `entry.ssr.js`) populate it. I'll add a jsdoc comment or similar to explain what's going on, and point back to there.

Agreed on the error, that was just me being a little lazy.

LeoMcA: Added a comment, and realised the name of `AsyncLocalStorageContents` may have seemed a little "official" - like it was coming from node itself - rather than something we had defined so renamed it, in [`198df40` (#1072)](https://github.com/mdn/fred/pull/1072/commits/198df40d7eb247b258bfa40566bbab7201a08e09)

components/server/async-local-storage.js
```diff
@@ -1,4 +1,13 @@
 import { AsyncLocalStorage } from "node:async_hooks";
 
-/** @type {AsyncLocalStorage<import("./types.js").AsyncLocalStorageContents>} */
+/**
+ * Store for internal context passed around components.
+ *
+ * Generally only used within the `ServerComponent` class itself,
+ * or very special server components (such as the `OuterLayout`).
```
caugner, line 7: nit: Can you give an example?

PR #1081, opened 2025-11-14
Observatory allow port paths url style

argl:
### Description

This enables port/path support in observatory, if the host is given as a comnplete URL

### Motivation

This was a bug that prevented the port/path support in the observatory API to work, if the input was given as a full URL (traditionally, only the bare host name was allowed, but this has bee incrementally opened for convenience).


components/observatory-form/element.js
```diff
@@ -49,7 +49,14 @@ export class MDNObservatoryForm extends LitElement {
...
       const url = new URL(host);
-      this._hostname = url.hostname.trim() || host;
+      let fullHost = url.hostname.trim() || host;
+      if (url.port) {
+        fullHost += ":" + url.port;
+      }
+      if (url.pathname && url.pathname !== "/") {
+        fullHost += url.pathname;
+      }
+      this._hostname = fullHost;
```
caugner, line 59: Nit: Might be good to rename this property to reflect that this is no longer a hostname.

PR #1138, opened 2025-12-07
chore(footer): add socials + external links in new tab

muhammad-paksi:
Add target="_blank" rel="noopener noreferrer"

### Description

Updates footer social links and external navigation links to open in a new tab by adding target="_blank" along with rel="noopener noreferrer" for security and privacy.

### Motivation

External links (such as social media and outbound resources) should open in a new tab to avoid disrupting the current navigation context.

### Additional details

Applies only to external links; internal navigation behavior remains unchanged.

### DIFF
--- a/components/footer/server.js
+++ b/components/footer/server.js
@@ -175,7 +175,7 @@ export class Footer extends ServerComponent {
                   <a
                     href=${item.href}
                     target="_blank"
-                    rel="noopener noreferrer"
+                    rel="noopener"
                     aria-label=${item.ariaLabel}
                     data-icon=${item.icon}
                   ></a>


components/footer/server.js
```diff
@@ -174,6 +174,8 @@ export class Footer extends ServerComponent {
                 <li>
                   <a
                     href=${item.href}
+                    target="_blank"
+                    rel="noopener noreferrer"
```
caugner:
```suggestion
                    rel="noopener"
```

components/footer/server.js
```diff
@@ -194,8 +196,11 @@ export class Footer extends ServerComponent {
                           <a
                             href=${link.href}
                             class=${link.external ? "external" : ""}
-                            >${link.text}</a
+                            target=${link.external ? "_blank" : undefined}
+                            rel=${link.external ? "noopener noreferrer" : undefined}
```
caugner:
```suggestion
                            rel=${link.external ? "noopener" : undefined}
```

caugner (changes requested):
LGTM, but let's keep sending the referrer (it is only the origin).

PR #1197, opened 2026-01-13
ci(workflows): add update-menu workflow

caugner:
### Description

Adds new `update-menu` workflow that runs the `update-missing-docs` script, and opens a PR if there are any changes.

### Motivation

Ensure that we link to localized pages when they exist.

### Additional details

Later we can also apply redirects to the actual menu items.

### DIFF
--- a/.github/workflows/update-menu.yml
+++ b/.github/workflows/update-menu.yml
@@ -6,10 +6,6 @@ on:
   schedule:
     - cron: "30 4 * * 1-5"
 
-  push:
-    branches:
-      - update-menu-workflow
-
 permissions: {} # No GITHUB_TOKEN permissions, because we use MDN_BOT_PAT instead.
 
 jobs:


.github/workflows/update-menu.yml
```diff
@@ -0,0 +1,58 @@
...
+on:
+  workflow_dispatch:
+
+  schedule:
+    - cron: "30 4 * * 1-5"
+
+  push:
+    branches:
+      - update-menu-workflow
+
```
LeoMcA: Remove this before merging?

.github/workflows/update-menu.yml
```diff
@@ -0,0 +1,58 @@
...
+          commit-message: "chore(menu): update missing docs"
+          author: mdn-bot <108879845+mdn-bot@users.noreply.github.com>
+          committer: mdn-bot <108879845+mdn-bot@users.noreply.github.com>
+          signoff: false
+          branch: update-menu
+          delete-branch: true
+          add-paths: components/menu/missing-docs.json
+          title: "chore(menu): update missing docs"
+          body: (This PR was generated by [this run](https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}) of the [`${{ github.workflow }}`](https://github.com/${{ github.repository }}/blob/${{ github.ref_name }}/.github/workflows/update-menu.yml) workflow.)
+          draft: false
```
LeoMcA, line 54: Could see some test runs of the workflow, but no tests of an actual PR being opened: we should test that somehow - perhaps a temporary modification to the workflow to point at an old main ref with an outdated missing-docs.json?

caugner: https://github.com/mdn/fred/actions/runs/20958471813 opened https://github.com/mdn/fred/pull/1200

caugner: PS: I couldn't figure out why capturing the output didn't work, so I removed that bit.

See: https://github.com/mdn/fred/actions/runs/20952453141/job/60209767207#step:5:7 and https://github.com/mdn/fred/actions/runs/20957060478/job/60224209712#step:5:26

PR #1154, opened 2025-12-10
fix(layout): increase gap between sidebar and content

LeoMcA:
A fix for https://github.com/mdn/fred/issues/690#issuecomment-3606927535, increasing the gap between content and the sidebar content by ~1rem to 3rem

Also fixes https://github.com/mdn/fred/issues/593: left sidebar no longer overlaps with header anchor hashes

PR #1221, opened 2026-01-17
fix(left-sidebar): text alignment of current page

mayank99:
### Description

This PR adds a negative margin to the `<em>` element which houses the current link (`<a aria-current="page">`) inside `left-sidebar`. The value of this negative margin is calculated as the sum of padding and border width.

It seems like this codebase mixes logical properties with physical properties everywhere, which is confusing and probably not intentional. For this PR, I've converted the `border-left` declaration within the same ruleset into its logical equivalent (`border-inline-start`) ~~and also grouped it together with the `padding-inline` and `margin-inline-start` declarations for better readability / local reasoniing~~.

### Motivation

This change was necessary for optical left alignment of all items within the same list.

| Before | After |
| --- | --- |
| Screenshot showing the beginning of Core modules. The active item in the list is emphasized with border and additional padding, producing additional space before the text starts | Screenshot of the exact same UI but with the text inside the active item neatly aligned with the non-active items |

### Additional details

N/A

N/A

### DIFF
--- a/components/curriculum/sidebar.css
+++ b/components/curriculum/sidebar.css
@@ -102,9 +102,12 @@
 
     padding-inline: 0.5rem;
 
+    /* Negative margin for text alignment. */
+    margin-inline-start: calc(-1 * (0.5rem + 2px));
+
     font-style: normal;
 
     background-color: var(--background-toc-active);
-    border-left: 2px solid var(--category-color);
+    border-inline-start: 2px solid var(--category-color);
   }
 }
--- a/components/left-sidebar/server.css
+++ b/components/left-sidebar/server.css
@@ -111,10 +111,13 @@
 
     padding-inline: 0.5rem;
 
+    /* Negative margin for text alignment. */
+    margin-inline-start: calc(-1 * (0.5rem + 2px));
+
     font-style: normal;
 
     background-color: var(--color-area-background);
-    border-left: 2px solid var(--color-area-highlight-border);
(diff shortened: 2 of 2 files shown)


LeoMcA (changes requested):
This causes an issue when the highlighted item is at the top level, where the accent is cut off:

Image

I'm touching the sidebar padding/margin in https://github.com/mdn/fred/pull/1154 so it might be worth revisiting after that is merged.

Please also run prettier.

mayank99:
I can move this to draft while waiting for https://github.com/mdn/fred/pull/1154.

mayank99:
> This causes an issue when the highlighted item is at the top level, where the accent is cut off:
>
> I'm touching the sidebar padding/margin in https://github.com/mdn/fred/pull/1154 so it might be worth revisiting after that is merged.

I think this problem is now resolved with the recent sidebar changes.

LeoMcA (changes requested):
Unfortunately the current approach doesn't work with our `<details>` items, e.g. on `/en-US/docs/Web/CSS/Reference/Properties`:

Image

LeoMcA:
I've opened https://github.com/mdn/fred/pull/1407 to fix this, along with the `<details>` case, so I'll close this - but thank you for your work!

mayank99:
Ah that's perfect. I haven't had time to get back to this, apologies.

PR #1227, opened 2026-01-19
fix(outer-layout): avoid `summary_large_image` in `twitter:card`

caugner:
### Description

Updates the `twitter:card` meta value from `summary_large_image` to `summary`.

### Motivation

Avoid the image from being shown unnecessarily large in link previews.

### Additional details

| Before | After |
|--------|--------|
| image | image | 

Fixes https://github.com/mdn/fred/issues/960.


caugner:
I tested how it looks on X, and unfortunately the image doesn't look good:

image


So we would probably want to change it at the same time.

components/outer-layout/server.js
```diff
@@ -173,15 +173,14 @@ export class OuterLayout extends ServerComponent {
...
-        "https://developer.mozilla.org/mdn-social-share.d893525a4fb5fb1f67a2.png",
+      "og:image": "https://developer.mozilla.org/mdn-social-image.46ac2375.png",
       "og:image:type": "image/png",
-      "og:image:height": "1080",
-      "og:image:width": "1920",
+      "og:image:height": "1024",
+      "og:image:width": "1024",
       "og:image:alt":
         "The MDN Web Docs logo, featuring a blue accent color, displayed on a solid black background.",
       "og:site_name": "MDN Web Docs",
```
LeoMcA: Might we want to remove references to "MDN Web Docs" here? Perhaps a question for @mdn/content-team

Rumyra:
> So we would probably want to change it at the same time.

@caugner Yes I agree here - let's make it look ok now and not have to do a follow up

components/outer-layout/server.js
Rumyra: @LeoMcA are you suggesting we just have "MDN" here (I think we should keep sitename)

LeoMcA: Yes, sorry: I'm not suggesting we remove the site name nor description, just change references of "MDN Web Docs" to "MDN" in these places, as we appear to have dropped the former from our branding

components/outer-layout/server.js
```diff
@@ -173,15 +173,14 @@ export class OuterLayout extends ServerComponent {
...
-      "og:image":
-        "https://developer.mozilla.org/mdn-social-share.d893525a4fb5fb1f67a2.png",
+      "og:image": "https://developer.mozilla.org/mdn-social-image.46ac2375.png",
       "og:image:type": "image/png",
-      "og:image:height": "1080",
-      "og:image:width": "1920",
+      "og:image:height": "1024",
+      "og:image:width": "1024",
       "og:image:alt":
         "The MDN Web Docs logo, featuring a blue accent color, displayed on a solid black background.",
```
caugner: Shall we improve the alt text? (cc @LeoMcA)

```suggestion
        "MDN logo on a near-black background, featuring a light-blue, stylized 'M' made of two alternating slashes and vertical bars, white 'mdn' lettering, and a light-blue underscore.",
```

LeoMcA: Is mentioning the colours necessary? I'd keep it short and simple:

```suggestion
        "The MDN logo",
```

components/outer-layout/server.js
Rumyra: Cool

components/outer-layout/server.js
caugner: According to https://ogp.me/#structured, it should be "a description of what is in the image (not a caption)".

LeoMcA: IMO "the MDN logo" is a description of the image: that's what the image is. AIUI irrelevant detail makes it harder for users of alt text e.g. on a swag store with an MDN logo t-shirt this would be a great description of something you're potentially purchasing and wearing, but on a social website the exact makeup of the MDN logo seems to me to be unimportant extra information.

caugner: Applied your suggestion. 🙂

PR #1230, opened 2026-01-20
feat(components): add social-image

LeoMcA:
https://github.com/mdn/fred/pull/1227 requires a new, square, social image.

I thought it would be fun to implement this as a component, and help maintainability in the future: e.g. it takes colours direct from our colour theme, and if we wanted to generate custom thumbnails for each page using HTML/CSS this would give us a starting point.

### DIFF
--- a/components/social-image/README.md
+++ b/components/social-image/README.md
@@ -6,8 +6,11 @@
 2. Screenshot the 1024px node from devtools
 3. Copy to `fred/public/twitter-card-summary.png`
 4. Compress the image:
+
    a. Run pngquant to reduce the colours in the image, try `pngquant --speed 1 twitter-card-summary.png` first, and see if it's indistinguishable from the original
+
    b. Run optipng to further compress the image without any quality loss: `optipng -o7 -zm1-9 twitter-card-summary-fs8.png`
+
 5. Generate a hash for the image: `shasum twitter-card-summary-fs8.png | cut -c 1-8`
 6. Rename the image, including the hash: `mv twitter-card-summary-fs8.png twitter-card-summary.46ac2375.png`
 7. Cleanup the uncompressed image: `rm twitter-card-summary.png`


components/social-image/README.md
```diff
@@ -0,0 +1,13 @@
+# Social Image
+
+## Generating an image from the component
+
+1. Navigate to the sandbox http://localhost:3000/sandbox#SocialImage
+2. Screenshot the 1024px node from devtools
+3. Copy to `fred/public/twitter-card-summary.png`
+4. Compress the image:
+   a. Run pngquant to reduce the colours in the image, try `pngquant --speed 1 twitter-card-summary.png` first, and see if it's indisigunisable from the original
```
LeoMcA:
```suggestion
   a. Run pngquant to reduce the colours in the image, try `pngquant --speed 1 twitter-card-summary.png` first, and see if it's indistinguishable from the original
```

components/social-image/server.js
```diff
@@ -0,0 +1,13 @@
+import { html } from "@lit-labs/ssr";
+
+import { Logo } from "../logo/server.js";
+import { ServerComponent } from "../server/index.js";
+
+export class SocialImage extends ServerComponent {
```
caugner, line 6: Nit: To avoid confusion, how about `MDNSocialLogo`?

```suggestion
export class MDNSocialLogo extends ServerComponent {
```

components/social-image/README.md
```diff
@@ -0,0 +1,13 @@
...
+   b. Run optipng to further compress the image without any quality loss: `optipng -o7 -zm1-9 twitter-card-summary-fs8.png`
```
caugner: This doesn't render as a sublist in Markdown. Needs 4 spaces of indentation I think.


```suggestion
    a. Run pngquant to reduce the colours in the image, try `pngquant --speed 1 twitter-card-summary.png` first, and see if it's indisigunisable from the original
    b. Run optipng to further compress the image without any quality loss: `optipng -o7 -zm1-9 twitter-card-summary-fs8.png`
```

public/twitter-card-summary.46ac2375.png
caugner, line 1: The hash isn't mentioned in the README.md above. Is this a leftover, or intentional?

LeoMcA: Steps 5 and 6 mention the hash, no?

caugner: True, not sure how I could have missed that. 🙈

components/social-image/server.js
LeoMcA: Eeeeeh, given how some of the internal naming works I think prefixing with MDN would get *more* confusing: linting would require us to put it in a folder called `mdn-social-image`, and if it needed a web component that would have to be called `mdn-mdn-social-image`, etc: I think if we need the social-image name for something else in the future which renders *other* sites social images, or something, we can rename this then.

PR #1234, opened 2026-01-21
fix(about-tabs): apply styles correctly on hover

### LINKED ISSUE #744: About page tabs don't have hover effect
STR:

1. Go to https://developer.mozilla.org/en-US/about
2. Hover over the tabs (What we offer, Our journey, etc.)

Expected: Hover effect.

Actual: No effect on hover.

brysonbw:
### Changes
- Ensure tab styles are applied on hover and active states

### Before

image

### After

https://github.com/user-attachments/assets/d6edb29f-866e-42ab-bd27-70fdddabf0ea

Fixes #744

### DIFF
--- a/components/about-tabs/element.css
+++ b/components/about-tabs/element.css
@@ -31,14 +31,14 @@
 
   font-weight: normal;
 
-  color: var(--about-tablist-color);
+  color: var(--about-tablist-color) !important;
 
   text-decoration: none;
 }
 
 ::slotted([slot="tab"]:hover),
 ::slotted([slot="tab"].active) {
-  color: var(--about-tablist-active-color);
+  color: var(--about-tablist-active-color) !important;
   border-bottom: 2px solid var(--about-tablist-active-border);
 }
 


components/about-tabs/element.css
```diff
@@ -31,13 +31,14 @@
 
   font-weight: normal;
 
-  color: var(--about-tablist-color);
+  color: var(--about-tablist-color) !important;
```
caugner, line 34: Does Obseratory also use `!important`? If not, we want to avoid `!important` here as well.

brysonbw: It doesn't and I agree. Yet, without it that color isn't applied and `color: var(--color-link-normal);` takes precedence. If intended then all smooth - will remove

caugner: I took a look, and it seems that we only have two options here, because the global style takes precedence over the style on the slot:

1. Assign the `color` in the global style, by creating a `global.css` with `mdn-about-tabs [slot="tab"] { color: var(--about-tablist-color); }`.
2. Keep the `!important` delimiter.

Given that option 1 is not very elegant, I think we should actually go with option 2, so I reverted your second commit, reinstating the `!important`.

caugner (approved):

PR #1240, opened 2026-01-21
chore(deps): migrate from insane to isomorphic-dompurify

caugner:
### Description

Migrate from `insane` to `isomorphic-dompurify`.

### Motivation

Avoid `insane`, which is unmaintained and has an unpatched vulnerability.

### Additional details

Add 33 sanitizer regression cases covering HTML rendering, hook isolation,
element name/tag matching, translator-supplied attributes, and URL
validation, including relative links.

The SSR bundle externalizes `isomorphic-dompurify`, because jsdom reads
package-relative assets that cannot be bundled. It resolves from
`node_modules` at runtime, so the package stays in `dependencies`.

The client `index` entrypoint grows from 59.8 KB to 68.9 KB gzipped
(170.5 KB to 195.5 KB raw), since `dompurify` is larger than `insane`.

Resolves GHSA-w455-mfq9-hf74.


caugner:
Closing in favor of build-time validation of `.ftl` markup, tracked in #1918. See the issue for the bundle-size motivation.

PR #1241, opened 2026-01-21
fix(server): handle HEAD requests + send Content-Length header

caugner:
### Description

Fixes the `server.js`:

- Handle HEAD requests properly.
- Send `Content-Length` header to clients.

### Motivation

Mainly avoid that `curl -I` requests crash the dev server.

### Additional details

Note: This only affects the local dev server.

#### Before

```console
% curl -I http://localhost:3000/en-US/
curl: (52) Empty reply from server
```

And crashes the server:

```
SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at ProxyServer.proxyRes (file:///path/to/mdn/fred/server.js:253:29)
    at processTicksAndRejections (node:internal/process/task_queues:103:5)
```

#### After

```console
% curl -I http://localhost:3000/en-US/
HTTP/1.1 200 OK
X-Powered-By: Express
Content-Type: text/html
Content-Length: 151484
Date: Wed, 21 Jan 2026 20:08:19 GMT
Connection: keep-alive
Keep-Alive: timeout=5
```

### DIFF
--- a/server.js
+++ b/server.js
@@ -93,7 +93,7 @@ async function serverRenderMiddleware(req, res, page) {
       "Content-Type": "text/html",
       "Content-Length": Buffer.byteLength(html),
     });
-    res.end(html);
+    res.end(res.locals.wasHead ? undefined : html);
   } catch (error) {
     console.error("SSR render error:", error);
     res.writeHead(500).end();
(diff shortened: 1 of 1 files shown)


LeoMcA:
I'm not entirely sure this is worth the added complexity, why are you doing HEAD requests against the dev server?

caugner:
> I'm not entirely sure this is worth the added complexity, why are you doing HEAD requests against the dev server?

I regularly use `curl -I` for debugging.

Edit: And IIRC we make HEAD requests from HTTP Observatory.

Issue #1255, opened 2026-01-26
Migrate away from "@dependabot squash and merge"


caugner:
See: https://github.blog/changelog/2025-10-06-upcoming-changes-to-github-dependabot-pull-request-comment-commands/

> Beginning on January 27th (…), these comment commands will no longer be supported.

caugner:
In mdn/rumba, I noticed [this error](https://github.com/mdn/rumba/actions/runs/21723423948/job/62659110116?pr=743#step:3:8) following the migration:

> GraphQL: Resource not accessible by personal access token (enablePullRequestAutoMerge)

The explanation was that the PAT only had `pull-requests: write` permission, but not `content: write`.

PR #1257, opened 2026-01-26
ci(auto-merge): migrate away from "@dependabot squash and merge"

caugner:
### Description

Migrates from Dependabot merge comment to [GitHub auto-merge](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/incorporating-changes-from-a-pull-request/automatically-merging-a-pull-request).

### Motivation

The "squash and merge" command is no longer supported from 2026-01-27.

Part of https://github.com/mdn/fred/issues/1255.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -26,5 +26,5 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.AUTOMERGE_TOKEN }}
         run: |
-          gh pr review ${{ github.event.pull_request.html_url }} --approve 
-          gh pr comment ${{ github.event.pull_request.html_url }} --body "@dependabot squash and merge"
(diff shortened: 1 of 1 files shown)


LeoMcA:
> Am I missing something

Yes, I am: this merges when all *required* workflows pass - so it solves the issue in fred, but can't where we have optional "required" workflows like content.

caugner:
> Am I missing something, or does this solve our issues with having to manually merge patch dependabot updates @caugner ?

Yes, it solves the issue, and is effectively equivalent to `@dependabot squash and merge`, because I marked all relevant workflows as required.

This only doesn't work in repositories where have "required" workflows that run conditionally only when certain files were changed. (In those cases, we would have to change the conditional workflows to always run, moving the conditional check into an initial job that runs before the actual check/lint/test job.)

PR #1265, opened 2026-01-30
fix(layout): consolidate 2 sidebar layout across content types

LeoMcA:
Split from https://github.com/mdn/fred/pull/1154 to extract the layout refactor/consolidation work without changing the gap between sidebar content and layout content so that can be discussed/approved by content.

Introduces a unified approach to how we do our 2 sidebar layout across docs, curriculum and generic content.

Commits are atomic, and some have lengthy explanations in their commit message: I'd recommend reviewing one by one.

Fixes these issues:
- fix for an issue where just above our mobile breakpoint the top placement would still be shown, breaking the toggleable sidebar overlay
- partial reversion of https://github.com/mdn/fred/commit/dff52a5c02563499e93e83b3e356fcb3e0cd9ee6, fixing this issue instead by:
  - allowing the left sidebar to show items in the minimum page margin on the left
  - showing the outline on each sidebar inset
- fixes the right sidebar scrollbar appearing offset from the page scrollbar
- fixes docs, curriculum and generic content using the 2 sidebar layout in subtly different ways requiring updates in multiple places to increase the padding between sidebar and content everywhere

If we like this approach, we should probably migrate all the other layouts using the variables to a class-based system, but it felt like I'd got to a coherent and consistent place with this so stopped here.

### DIFF
--- a/components/generic-layout/server.css
+++ b/components/generic-layout/server.css
@@ -2,11 +2,6 @@
 
 /* Generic Layout */
 
-/* Sidebar */
-.generic-layout__sidebar {
-  padding-block: 2rem;
-}
-
 /* Content */
 .generic-layout__content {
   padding-block: 2rem;
--- a/components/generic-sidebar/server.css
+++ b/components/generic-sidebar/server.css
@@ -2,7 +2,7 @@
   --color-accent: var(--color-blue-50);
   --color-background-accent: var(--color-background-blue);
   --color-border-items: light-dark(var(--color-gray-80), var(--color-gray-20));
-
+  padding-block: 2rem;
   margin-bottom: 2rem;
 
   .generic-sidebar__header {


components/layout/2-sidebars.css
```diff
@@ -0,0 +1,141 @@
+.layout__2-sidebars,
+.layout__2-sidebars-inline {
```
LeoMcA, line 2: from @caugner: [nit] maybe:

```suggestion
.layout__2-sidebars--inline {
```

components/layout/2-sidebars.css
```diff
@@ -0,0 +1,141 @@
+.layout__2-sidebars,
```
LeoMcA, line 1: from @caugner: [nit] maybe add comment explaining the different classes, and where they are used

components/generic-layout/server.css
```diff
@@ -1,69 +1,18 @@
...
-  padding: 2rem 0;
+  padding-block: 2rem;
 }
 
 /* TOC */
 .generic-layout__toc {
-  grid-column: toc;
-  padding: 2rem 0;
+  padding-block: 2rem;
 }
```
LeoMcA: from @caugner: [nit] should these blocks be merged now?

components/page-layout/server.css
```diff
@@ -29,7 +29,7 @@
     color-scheme: dark;
   }
 
-  @media (--screen-small-and-narrower) {
+  @media (--screen-small-and-narrower) or (--screen-layout-no-sidebar) {
```
LeoMcA, line 32: from @caugner: does this mean we show placements less often?

LeoMcA (commented):
Copied review comments from https://github.com/mdn/fred/pull/1154#pullrequestreview-3727114037

mirunacurtean:
Tested for various layouts on Chrome/Firefox on Windows 11 using an [example](https://test.developer.allizom.org/en-US/docs/Web/JavaScript/Reference/Errors/Negative_repetition_count) with sidebar and TOC: 
**Laptop 1440px** 
image

**Laptop 1024px**
image
image

Tablet 768
image
image

Mobile 425px 
image
image


Tested on iOS on Safari/ Frirefox across generic content, curriculum and tools: 
image
image

Generally looks good, there is only one change that is not an improvement @LeoMcA in the Observatory docs sidebar- the heading is higher. 
Prod 
image

Test.allizom
image

**Test links:** 
https://test.developer.allizom.org/en-US/docs/Web/CSS/Reference/Properties#syntax 
https://test.developer.allizom.org/en-US/docs/Web/CSS/Guides/Backgrounds_and_borders/Box-shadow_generator
https://test.developer.allizom.org/en-US/docs/Web/CSS/Reference/At-rules
https://test.developer.allizom.org/en-US/docs/Web/CSS/Reference/Properties/-moz-force-broken-image-icon 
https://test.developer.allizom.org/en-US/docs/Web/CSS/How_to/Layout_cookbook/Column_layouts
https://test.developer.allizom.org/en-US/docs/Web/JavaScript/Guide
https://test.developer.allizom.org/en-US/docs/Web/JavaScript/Reference/Errors/Negative_repetition_count
https://test.developer.allizom.org/en-US/observatory/docs/tests_and_scoring 
https://test.developer.allizom.org/en-US/curriculum/core/css-layout/#5.4_modern_layout
https://test.developer.allizom.org/en-US/docs/Web/API 
https://test.developer.allizom.org/en-US/docs/Web/API/Window/error_event

components/page-layout/server.css
LeoMcA: Not really: it moved the breakpoint horizontally by about 25px to avoid this buggy layout:

Screen Shot 2026-02-03 at 17 54 02

Issue #1028, opened 2025-10-30
Incomplete Playground measurements


caugner:
We're missing at least the following measurements:

```
playground: share-markdown
playground: share-permalink
playground: reset-click
playground: revert-click
playground: share-click
playground: flag-click
playground: queue $action
```

The following are also missing, but might no longer be relevant:

```
playground: banner-link
playground: banner-login
playground: banner-signup
```

PR #1266, opened 2026-01-30
feat(playground): add share data url button

LeoMcA:
@caugner as we briefly discussed in our 1:1

since working on devtools, I've noticed it's quite a common/nice pattern to share testcases (on slack, bugzilla, etc.) with `data:text/html` urls. authoring them is a bit of pain though: I asked the team if there was any better way than manually typing into the url bar... and they told me that's what they were all doing!

I've added in the simplest possible way for now: obviously references to static assets are going to break, we could remove non-syntactic line breaks to make the urls a bit shorter/prettier, etc. but we can make those improvements if the metrics show any usage/we get direct feedback about it.

I did make the effort to only urlencode non-space whitespace as otherwise the urls are very un-readable, and what's nice about the practice is you can read what the page you're about to load is before pasting it in. Firefox/Chromium/Epiphany seem to handle unencoded utf8 characters just fine when pasted in (I'll check in Safari once my mac is charged).

For an example where newlines are important:

```html
hello
world
❤️
```

```css
body {
  white-space: pre;
}

```

```js
console.log(`hello
world!`);

```

The generated url is readable, and functional!

``data:text/html;charset=utf-8,<!doctype html><style>body {%0A  white-space: pre;%0A}%0A</style>hello%0Aworld%0A❤️<script>console.log(`hello%0Aworld!`);%0A</script>``

### DIFF
--- a/wdio.conf.js
+++ b/wdio.conf.js
@@ -17,6 +17,7 @@ export const config = {
     [
       "firefox-profile",
       {
+        // Give the test framework permissions to read the clipboard:
         "dom.events.testing.asyncClipboard": true,
         "dom.events.testing.asyncClipboard.readText": true,
       },


components/playground/element.js
```diff
@@ -118,6 +118,22 @@ ${"```"}`,
     }
   }
 
+  async _copyDataUrl() {
+    const controller = this._controller.value;
+    if (controller) {
+      const { css, html, js } = controller.code;
+      let code = `<!doctype html>`;
+      if (css) code += `<style>${css}</style>`;
+      if (html) code += html;
```
caugner, line 128: Nit: What if `html` contains the doctype, or a full HTML structure? (I guess in this case, chances are `css` is empty.)

components/playground/element.js
```diff
@@ -372,10 +388,22 @@ ${"```"}`,
       <mdn-modal ${ref(this._shareModal)} class="share">
         <section>
           <h2>${this.l10n`Share Markdown`}</h2>
-          <mdn-button variant="secondary" @click=${this._copyMarkdown}
+          <mdn-button
+            variant="secondary"
+            @click=${this._copyMarkdown}
+            data-glean-id="playground: copy-markdown"
```
caugner:
```suggestion
            data-glean-id="playground: share-markdown"
```

See: https://github.com/mdn/fred/issues/1028#issue-3571645538

components/playground/element.js
```diff
@@ -389,6 +417,7 @@ ${"```"}`,
                       <mdn-button
                         variant="secondary"
                         @click=${this._copyPermalink}
+                        data-glean-id="playground: copy-permalink"
```
caugner:
```suggestion
                        data-glean-id="playground: share-permalink"
```

See: https://github.com/mdn/fred/issues/1028#issue-3571645538

components/playground/element.js
```diff
@@ -372,10 +388,22 @@ ${"```"}`,
...
+            data-glean-id="playground: copy-markdown"
             >${this.l10n`Copy markdown to clipboard`}</mdn-button
           >
         </section>
+        <section>
+          <h2>${this.l10n`Share Data URL`}</h2>
+          <mdn-button
+            variant="secondary"
+            @click=${this._copyDataUrl}
+            data-glean-id="playground: copy-data-url"
```
caugner: For consistency, see: https://github.com/mdn/fred/issues/1028#issue-3571645538

```suggestion
            data-glean-id="playground: share-data-url"
```

components/playground/element.js
```diff
@@ -118,6 +118,22 @@ ${"```"}`,
...
+  async _copyDataUrl() {
+    const controller = this._controller.value;
+    if (controller) {
+      const { css, html, js } = controller.code;
+      let code = `<!doctype html>`;
+      if (css) code += `<style>${css}</style>`;
+      if (html) code += html;
+      if (js) code += `<script>${js}</script>`;
+      // encode non-space whitespace
+      code = code.replaceAll(/[^\S ]/g, (ch) => encodeURIComponent(ch));
```
caugner: Encode sequences together, rather than each character individual:

```suggestion
      code = code.replaceAll(/[^\S ]+/g, (ch) => encodeURIComponent(ch));
```

components/playground/element.js
LeoMcA: We have the same issue in the playground/runner currently: a user could add a full HTML structure to the HTML input.

I don't think it's really a problem, HTML parsers are able to work around it, but if we did want to fix it we should do it cohesively as a separate issue.

test/specs/playground.e2e.js
```diff
@@ -0,0 +1,64 @@
...
+      // will cause a webdriver error about permissions, but we've set them in firefox through about:config
+      await expect(browser).toHaveClipboardText(
+        expect.stringContaining("@%0A@ @%2520 @"),
+      );
+      await expect(browser).toHaveClipboardText(
+        expect.stringContaining(
+          "<style>body {%0A  font-size: 5em;%0A}</style>",
+        ),
+      );
+    });
```
caugner, line 62: [nit] I would expect this to be a unit test, rather than an e2e test.

wdio.conf.js
```diff
@@ -13,6 +13,15 @@ export const config = {
       },
     },
   ],
+  services: [
+    [
+      "firefox-profile",
+      {
+        "dom.events.testing.asyncClipboard": true,
+        "dom.events.testing.asyncClipboard.readText": true,
```
caugner, line 22: [nit] Are these here to stay,? Is there a page or bug with more info?

test/specs/playground.e2e.js
```diff
@@ -0,0 +1,64 @@
+import { $, browser, expect } from "@wdio/globals";
+
+import DocPage from "../pageobjects/doc.page.js";
+
+describe("Playground", () => {
+  describe("copy data url", () => {
+    beforeEach(async () => {
+      await DocPage.open("en-US/play");
```
caugner, line 8: [nit] Why is there no leading slash here?

```suggestion
      await DocPage.open("/en-US/play");
```

components/playground/element.js
caugner: Here's a reason to set a more complete HTML structure.

Copy the following in the Playground:

```js
document.body.style.backgroundColor = 'DeviceMotionEventAcceleration' in window ? 'green' : 'red';
```

It is green in the Playground runner, but when copying and opening the data URL, the page is white, because there is no body.

test/specs/playground.e2e.js
LeoMcA: True, though we don't currently have the setup for running unit tests - it's not complex, but I'd prefer to separate concerns from this PR: I'll add the setup, migrate this test, and document in a follow up PR.

wdio.conf.js
LeoMcA: Yes, I'll add an explanatory comment

test/specs/playground.e2e.js
LeoMcA: `baseUrl` in the config has a trailing slash, and this is appended: another fix for a follow up PR!

PR #1286, opened 2026-02-06
fix(content-section): Make table headings bold

### LINKED ISSUE #1247: content-section table header has normal font-weight instead of bold

In the content-section table, the header is hard to distinguish from the body because they have the same font-weight—is this intended?

Image

Image

https://github.com/mdn/fred/blob/a7dcfb5a88c60e531a13ab818197087e09f623fc/components/content-section/server.css#L159-L162

```diff
 th { 
-  font-weight: var(--font-weight-normal);
+  font-weight: var(--font-weight-bold);
   text-align: left;
 }
```

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#:~:text=method-,Non%2Dmutating%20alternative

shaedrich:
### Description
Changes the font-weight in content-section table headings from normal to bold.

### Motivation
Table headings should be visually distinguishable from content rows.

### Additional details
| Before | After |
| ------ | ----- |
| Image | Image |

Fixes #1247

### DIFF
--- a/components/content-section/server.css
+++ b/components/content-section/server.css
@@ -157,7 +157,7 @@
     }
 
     th {
-      font-weight: var(--font-weight-normal);
+      font-weight: var(--font-weight-bold);
       text-align: left;
     }
 


caugner (approved):
LGTM, this looks like [how it was before in Yari](https://web.archive.org/web/20250802180719/https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array#copying_methods_and_mutating_methods):

image

Issue #1311, opened 2026-02-17
Nested sidebar items show visually misleading hierarchy


caugner:
Reported by @joewalker via Slack:

> it looks like dom is nested under dns.

<img width="1107" height="437" alt="Image" src="https://github.com/user-attachments/assets/fc844445-2eb2-452d-aed6-a8205ad86706" />

See: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/dom/openOrClosedShadowRoot

The same applies to the CSS sidebar, see: https://developer.mozilla.org/en-US/docs/web/20260102003136/https:_developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/align-content

LeoMcA:
This isn't especially simple to fix, and requires @mdn/content-team making some decisions about how we view hierarchy in the sidebar.

I've prototyped a few approaches to this, and opened PRs so we have preview builds of most of our sidebars. I'd encourage you to click around those, rather than just looking at the screenshots I've assembled below.

These are intended to showcase the different ways we can represent the hierarchy: they're not necessarily perfect, and we can make minor adjustments to padding, certain well-defined edgecases, etc.

First the screenshots, then let me explain each column:

| `main`                                                                                                                             | [`dedent-details`](https://fred-pr1322.review.mdn.allizom.net/en-US/)                                                              | [`dedent-summary`](https://fred-pr1323.review.mdn.allizom.net/en-US/)                                                              | [`chevron-right`](https://fred-pr1324.review.mdn.allizom.net/en-US/)                                                               | [`indent-all`](https://fred-pr1326.review.mdn.allizom.net/en-US/)                                                                  |
| ---------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------- |
[...]

PR #1312, opened 2026-02-17
fix(left-sidebar): unindent nested details

caugner:
### Description

Updates the `left-sidebar` layout, unindenting nested details:

- If the details element has nested details, unindent (only) the summary element.
- If the details element has no nested details, unindent the whole details element.

### Motivation

1. Avoid confusion.
2. Save horizontal space (possibly less line breaks in nested sidebars).

### Additional details

| Sidebar | Before | After |
|--------|--------|--------|
| CSS | image | image |
| WebExt | image | image | 

Fixes https://github.com/mdn/fred/issues/1311.

### DIFF
--- a/components/left-sidebar/server.css
+++ b/components/left-sidebar/server.css
@@ -78,6 +78,23 @@
 
     ol {
       padding-left: 1.5rem;
+
+      /* Unindent nested details so their chevron aligns with
+         the parent list level instead of being indented. */
+      > li:has(> details) > details {
+        /* Leaf details (no nested collapsibles):
+           pull back the entire block. */
+        &:not(:has(ol > li > details)) {
+          margin-left: -1.5rem;
+        }
+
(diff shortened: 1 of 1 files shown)


caugner:
At first, I thought the new behavior for CSS properties is confusing, but it's actually consistent with the current behavior at the top level, so I think it's fine.

LeoMcA:
I think this is overly complex, and piles on the edge-cases - a path we already know to be dangerous with the sidebars.

I'd first take a step back and re-evaluate how we consider the sidebar hierarchy, which I've attempted to do in https://github.com/mdn/fred/issues/1311#issuecomment-3946512145. We can then revisit and fix some remaining edge-cases with the not-yet-baseline-green `:has` selector - it shouldn't be the basis of our whole hierarchical display.

caugner:
Closing in favor of https://github.com/mdn/fred/pull/1329, which looks better.

PR #1329, opened 2026-02-25
fix(sidebar): align siblings' text, control indent directly

LeoMcA:
Best effort attempt at fixing https://github.com/mdn/fred/issues/1311 given the universal feedback on https://github.com/mdn/fred/issues/1311#issuecomment-3946512145 to go with the `indent-all` option.

This takes that approach - which ensures hierarchical consistency across each level (including the root) - and makes it better:
- Adds a variable controlling the indent level, which automatically resizes the chevron to match
- Saves some space around the chevron icon by cropping it slightly - ensuring the left edge of the chevron itself, not its containing box, is aligned with its parent's text
- Indents lists with no `<details>` element in it slightly less: this ends up looking visually more consistent, and saves space

With all this, despite the increase in base padding to make everything align on the root level, we actually increase the space available for text compared to `main`.

| Before                                                                                                                              | After                                                                                                                                              |
| ----------------------------------------------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------- |
| 1-main | 1-1311-indent-control |
| 2-main | 2-1311-indent-control |
| 3-main | 3-1311-indent-control |
| 4-main | 4-1311-indent-control |

### DIFF
--- a/components/left-sidebar/server.css
+++ b/components/left-sidebar/server.css
@@ -71,11 +71,11 @@
 
         flex-shrink: 0;
 
-        width: calc(var(--calc-width-height));
-        height: calc(var(--calc-width-height));
+        width: var(--calc-width-height);
+        height: var(--calc-width-height);
 
         margin-right: var(--calc-margin-right);
-        margin-left: calc(var(--calc-margin-left));
+        margin-left: var(--calc-margin-left);
 
         content: "";
 


components/left-sidebar/server.css
```diff
@@ -29,18 +48,34 @@
...
+        );
+
         display: inline-block;
 
         flex-shrink: 0;
 
-        width: 1rem;
-        height: 1rem;
+        width: calc(var(--calc-width-height));
+        height: calc(var(--calc-width-height));
```
caugner: Why wrap `var()` in `calc()`?

```suggestion
        width: var(--calc-width-height);
        height: var(--calc-width-height);
```

components/left-sidebar/server.css
```diff
@@ -29,18 +48,34 @@
...
 
-        margin: 0.25rem;
+        margin-right: var(--calc-margin-right);
+        margin-left: calc(var(--calc-margin-left));
```
caugner: dito
```suggestion
        margin-left: var(--calc-margin-left);
```

components/left-sidebar/server.css
LeoMcA: Ah good catch, thank you: no reason - just an artifact from trying different approaches here.

PR #1365, opened 2026-03-12
chore(tests): run unit tests, update script

LeoMcA:
Would recommend reviewing commit-by-commit with whitespace changes disabled.

Adds unit tests to our scripts/tests.js file, and revamps that file a little bit to support that. Reduces the number of scripts in package.json (I don't like it when they get out of hand and overly confusing!)

This introduces a single `npm test` entrypoint which can be used for running:
- linters
- unit tests
- e2e tests against:
  - currently fred or rari servers
  - a fred dev server
  - a fred preview server
  - fred/rari servers within a content repo


.github/workflows/npm-publish-simulation.yml
```diff
@@ -118,8 +118,7 @@ jobs:
         working-directory: mdn/fred
         env:
           FRED_PORT: 5042
-          CONTENT_REPO_ROOT: ../content
           # Increase GitHub API limit.
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          node scripts/npm-test.js
+          node scripts/tests.js --content ../content
```
caugner: [nit] Would still call it just `scripts/test.js`.

scripts/tests.js
```diff
@@ -4,19 +4,41 @@ import yargs from "yargs";
...
   .parse();
 
 concurrently(
   [
-    {
-      command: `npm run wdio`,
-      name: "wdio",
-      prefixColor: "green",
-    },
+    ...(argv.e2e
```
caugner: [nit] For readability, instead of the `...(condition ? [job] : [])` pattern, could we collect the jobs before with `if (condition) { jobs.push(job); }` instead?

scripts/tests.js
```diff
@@ -33,58 +33,69 @@ const argv = await yargs(hideBin(process.argv))
...
+        successCondition: "first",
+      },
+    ).result,
+  );
+}
+
+try {
+  await Promise.all(runs);
+} catch {
+  process.exitCode = 1;
```
caugner, line 125: [nit] Don't the individual commands return codes other than 1?

scripts/tests.js
caugner: [nit] Need to remove "preview" here

scripts/tests.js
```diff
@@ -52,7 +52,7 @@ if (argv.e2e) {
     concurrently(
       [
         {
-          command: `npm run wdio`,
+          command: `npx wdio run ./wdio.conf.js`,
```
caugner: [nit] Is the `./` needed?

It doesn't seem so from `npx wdio --help`:

```
wdio [command]

The `wdio` command allows you run and manage your WebdriverIO test suite.
If no command is provided it calls the `run` command by default, so:

$ wdio wdio.conf.js

is the same as:
$ wdio run wdio.conf.js
```

.github/workflows/npm-publish-simulation.yml
```diff
@@ -121,4 +121,4 @@ jobs:
           # Increase GitHub API limit.
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          node scripts/tests.js --e2e --content ../content --unit false
+          node scripts/tests.js --e2e --content ../content --unit false --lint false
```
caugner: [nit] According to https://yargs.js.org/docs/#api-reference-booleankey, we should be able to set `--no-unit --no-lint` for readability

package.json
```diff
@@ -16,7 +16,7 @@
     "preview": "NODE_ENV=production node ./server.js",
     "ssr": "node build/ssr.js",
     "lit-analyzer": "npx @jackolope/lit-analyzer --rules.no-incompatible-type-binding off",
-    "test": "lefthook run --force pre-push",
+    "test": "node scripts/tests.js",
```
caugner, line 19: [nit] Do we need to update the pre-push hook?

.github/workflows/test.yml
```diff
@@ -88,4 +88,4 @@ jobs:
         env:
           CONTENT_ROOT: ../content/files
           GITHUB_TOKEN: ${{ github.token }}
-        run: npm test -- --e2e --fred preview --rari --lint false
+        run: npm test -- --unit --e2e --fred preview --rari --lint false
```
caugner: [nit] The advantage of running the unit tests in a separate step is that you see right away which caused the issue. But if the e2e tests aren't flaky, it's less of an issue.

caugner (approved):

scripts/tests.js
caugner: Already fixed in later commit.

scripts/tests.js
LeoMcA: Not sure it matters, if any of them return a failing exit code, concurrently will throw a error, and we then make the overall script return a failing exit code in order to fail CI etc.

package.json
LeoMcA: I suppose it depends, do we want to run unit tests automatically on pre-push? Perhaps we should also let the tests.js script own the linters we run, rather than delegating that to the lefthook config file? I've leave this for a later PR.

scripts/tests.js
LeoMcA: Fixed in [`bb8559f` (this PR)](https://github.com/mdn/fred/pull/1365/commits/bb8559f82d46771d4d175e69e4aa380d6a8f8623)

scripts/tests.js
LeoMcA: Fixed in [`d472fbf` (this PR)](https://github.com/mdn/fred/pull/1365/commits/d472fbffcde6da71f748fe36f091f954cd1d1384)

.github/workflows/npm-publish-simulation.yml
LeoMcA: Fixed in [`3cfd8f9` (this PR)](https://github.com/mdn/fred/pull/1365/commits/3cfd8f90681bd7659039455d43fdf63d17d64209)

.github/workflows/test.yml
LeoMcA: Fixed in [`8291f9c` (this PR)](https://github.com/mdn/fred/pull/1365/commits/8291f9c880afdccb6d2cb7047e93497931517930)

PR #1367, opened 2026-03-12
fix(user-menu): migrate l10n to new format, remove manual arg

LeoMcA:
@caugner fixes the issue we noticed in our 1:1, also cleans up the manual locale assignment we don't need to do with the mixin.

Before:

image

After:

image

When running against our pseudo-locale I identified another bug where things race and we don't actually load the localised strings here - I'll be opening a separate PR to fix that.


caugner (approved):
Nice!

PR #1378, opened 2026-03-13
chore(l10n): migrate CLI to yargs

LeoMcA:
Migrates the l10n cli to use yargs, and moves the current base command into an `extract` command.

This will make it easier to add future commands in the (near) future (e.g. `npm l10n merge`)


l10n/cli.js
```diff
@@ -1,20 +1,33 @@
...
-  }
-}
+await yargs(hideBin(process.argv))
+  .command({
+    command: "extract",
+    describe:
+      "Generates template.ftl by combining en-US.ftl and inline l10n tagged strings",
+    builder: (yargs) =>
+      yargs
+        .option("lint", {
```
caugner, line 15: Could this be a separate `lint` command?

Otherwise, to avoid confusion, we could call the option `check` (like `prettier --check` and `cargo fmt -- --check`).

l10n/cli.js
```diff
@@ -1,20 +1,33 @@
...
+import { generateQaaLocale } from "./parser/transform.js";
 
-  const generatePseudo = process.argv.includes("--gen-pseudo");
-  if (generatePseudo) {
-    await generateQaaLocale();
-  }
-}
+await yargs(hideBin(process.argv))
+  .command({
+    command: "extract",
```
caugner, line 10: Could this be called `sync` ~~or `fix`~~?

PR #1387, opened 2026-03-16
fix(sidebar): make current page indentation not confusing

mkljczk:
### Description

This aligns the position of the emphasized sidebar item text with other elements at the same nesting levels

### Motivation

The current positioning makes the hierarchy harder to read at a glance and made `mix-blend-mode` appear as though it might have been nested under `min-*` by mistake. I double-checked that impression with my partner to confirm the ambiguity wasn’t just on my side.

<table>
<tr>
 <td>Before
 <td>After
<tr>
 <td>
Screenshot_2026-03-16_18-52-39
Screenshot_2026-03-16_19-00-42
 <td>
Screenshot_2026-03-16_18-52-54
Screenshot_2026-03-16_19-00-54

</table>

This didn’t look confusing before #1329. I hope my PR doesn’t introduce any regressions like this. Tried to browse around MDN and didn’t notice any issues.

### DIFF
--- a/components/left-sidebar/server.css
+++ b/components/left-sidebar/server.css
@@ -150,6 +150,10 @@
     border-left: 2px solid var(--color-area-highlight-border);
   }
 
+  li li em {
+    margin-left: calc(-0.5rem - 2px);
+  }
+
   .icon {
     margin-left: 0.5ch;
   }


LeoMcA:
Thank you for your report and work here, but I've opened https://github.com/mdn/fred/pull/1407 to handle cases with a dropdown, so I'll close this.

PR #1392, opened 2026-03-17
chore(eslint): enable `@typescript-eslint/recommended` config

caugner:
### Description

Enables the `@typescript-eslint/recommended` config.

### Motivation

- Enables additional TypeScript-related rules.

### Additional details

It looks like it caught a bug with the [community issue list](https://developer.mozilla.org/en-US/community#help_us_fix_open_issues) (missing labels).


components/issues-table/element.js
```diff
@@ -91,7 +91,7 @@ export class MDNIssuesTable extends L10nMixin(LitElement) {
                         typeof label === "object" && label !== null
                           ? label.name
                           : label;
-                      this.LABELS.includes(labelName)
+                      return this.LABELS.includes(labelName)
```
caugner, line 94: This was actually broken on prod!

PR #1403, opened 2026-03-20
feat(menu): support entry localization

### LINKED ISSUE #1402: Make menu localizable
The menu is composed from [components/menu/tabs.json](https://github.com/mdn/fred/blob/main/components/menu/tabs.json), which includes the English descriptions.

The JSON file is maintained via the [update-tabs](https://github.com/mdn/fred/blob/main/components/menu/update-tabs.js) and [update-missing-docs](https://github.com/mdn/fred/blob/main/components/menu/update-missing-docs.js) scripts, so we inlining them in the component is not really an option.

caugner:
### Description

Makes the menu entries localizable.

### Motivation

Ensure other locales don't have to live with the English menu.

Fixes https://github.com/mdn/fred/issues/1402.


LeoMcA (commented):
Only done an initial skim so far, but nice idea. I imagine you've already thought of this, but this approach of exposing strings at "l10n extract" runtime would allow us to localise strings coming from an upstream source as well - nice!

Let's add some simple tests, with some simple fixtures, in the same vein as the current ones.

components/menu/l10n.js
```diff
@@ -0,0 +1,39 @@
...
+    /** @type {[string, string][]} */
+    const entries = [];
+
+    if (typeof tab.buttonL10nId === "string") {
+      entries.push([tab.buttonL10nId, /** @type {string} */ (tab.buttonText)]);
+    } else {
+      entries.push([tab.buttonL10nId.long, tab.buttonText.long], [tab.buttonL10nId.short, tab.buttonText.short]);
+    }
+
+    if (!("panelTitle" in tab)) return entries;
```
LeoMcA, line 26: nit: I'd prefer an `if("panelTitle" in tab) {` block wrapping the below lines, than a negated condition - makes the code read easier IMO

components/menu/l10n.js
```diff
@@ -0,0 +1,39 @@
...
+
+    entries.push([tab.panelTitle.l10nId, tab.panelTitle.text]);
+
+    for (const group of tab.panelGroups) {
+      if (group.titleL10nId) {
+        entries.push([group.titleL10nId, /** @type {string} */ (group.title)]);
+      }
+      for (const item of group.items) {
+        if (!("l10nId" in item)) continue;
```
LeoMcA, line 35: nit: similarly, a non-negated if would read better imo

components/menu/types.d.ts
```diff
@@ -1,5 +1,26 @@
 import { ServerRenderedTemplate } from "@lit-labs/ssr";
 
+export interface RawItem {
```
LeoMcA: check: so now we can remove the optional bits from the original types?

l10n/parser/extractor.js
```diff
@@ -48,6 +48,33 @@ export async function extract(options = {}) {
...
+ * @returns {Promise<Map<string, string>>}
+ */
+async function scrapeL10nModules(componentsDir) {
+  /** @type {Map<string, string>} */
+  const map = new Map();
+  for await (const file of glob("**/l10n.js", { cwd: componentsDir })) {
+    const { strings } = await import(
+      pathToFileURL(path.join(componentsDir, file)).href
+    );
+    if (!(strings instanceof Map)) continue;
```
LeoMcA, line 65: nit: avoid the negated if, wrap the section below

l10n/parser/extractor.js
```diff
@@ -48,6 +48,33 @@ export async function extract(options = {}) {
...
+    for (const [id, value] of strings) {
+      if (map.has(id) && map.get(id) !== value) {
+        throw new Error(
+          `L10n extractor: \`${id}\` is a duplicate id with different text`,
```
LeoMcA, line 69: this needs a test

l10n/parser/extractor.js
```diff
@@ -56,6 +83,16 @@ export async function createFluentResource(manualEntryPath, scrapeGlob) {
   const manualEntries = await getManualEntries(manualEntryPath);
   const tags = scrapeL10nTags(scrapeGlob);
 
+  const componentsDir = path.join(__dirname, "..", "..", "components");
+  for (const [id, value] of await scrapeL10nModules(componentsDir)) {
+    if (tags.has(id) && tags.get(id) !== value) {
+      throw new Error(
+        `L10n extractor: \`${id}\` is a duplicate id with different text`,
```
LeoMcA, line 90: this needs a test

l10n/parser/extractor.js
```diff
@@ -56,6 +83,16 @@ export async function createFluentResource(manualEntryPath, scrapeGlob) {
   const manualEntries = await getManualEntries(manualEntryPath);
   const tags = scrapeL10nTags(scrapeGlob);
 
+  const componentsDir = path.join(__dirname, "..", "..", "components");
```
LeoMcA, line 86: We already reference the components dir elsewhere in this file, let's extract them all to a constant.

l10n/template.ftl
```diff
@@ -318,3 +318,124 @@ user-menu-feedback = Feedback
...
+menu-javascript-panel-title = JavaScript: Scripting language
+menu-javascript-js-reference = JS reference
+menu-javascript-js-reference-standard-built-in-objects = Standard built-in objects
+menu-javascript-js-reference-expressions-operators = Expressions & operators
+menu-javascript-js-reference-statements-declarations = Statements & declarations
+menu-javascript-js-reference-functions = Functions
+menu-javascript-js-reference-see-all = See all…
+menu-javascript-js-reference-see-all-label = See all JavaScript references
+menu-javascript-js-guides = JS guides
+menu-javascript-js-guides-control-flow-error-handing = Control flow & error handing
```
LeoMcA, line 378: Typo in tabs.json: "handing" should be "handling"

l10n/template.ftl
```diff
@@ -318,3 +318,124 @@ user-menu-feedback = Feedback
...
+menu-webapis-panel-title = Web APIs: Programming interfaces
+menu-webapis-web-api-reference = Web API reference
+menu-webapis-web-api-reference-file-system-api = File system API
+menu-webapis-web-api-reference-fetch-api = Fetch API
+menu-webapis-web-api-reference-geolocation-api = Geolocation API
+menu-webapis-web-api-reference-html-dom-api = HTML DOM API
+menu-webapis-web-api-reference-push-api = Push API
+menu-webapis-web-api-reference-service-worker-api = Service worker API
+menu-webapis-web-api-reference-see-all = See all…
+menu-webapis-web-api-reference-see-all-label = See all Web API guides
```
LeoMcA, line 394: Error in tabs.json, should be references rather than guides?

components/menu/l10n.js
```diff
@@ -0,0 +1,45 @@
...
+
+    if (typeof tab.buttonText === "string") {
+      entries.push([/** @type {string} */ (tab.buttonL10nId), tab.buttonText]);
+    } else {
+      const buttonL10nId = /** @type {{ long: string; short: string }} */ (
+        tab.buttonL10nId
+      );
+      entries.push(
+        [buttonL10nId.long, tab.buttonText.long],
+        [buttonL10nId.short, tab.buttonText.short],
```
LeoMcA, line 22: I think we shouldn't be afraid to use typeof checks to avoid unnecessary (and, potentially in future, erroneous) casts:


```suggestion
    if (
      typeof tab.buttonText === "string" &&
      typeof tab.buttonL10nId === "string"
    ) {
      entries.push([tab.buttonL10nId, tab.buttonText]);
    } else if (
      typeof tab.buttonText !== "string" &&
      typeof tab.buttonL10nId !== "string"
    ) {
      entries.push(
        [tab.buttonL10nId.long, tab.buttonText.long],
        [tab.buttonL10nId.short, tab.buttonText.short],
```

components/menu/l10n.js
```diff
@@ -0,0 +1,45 @@
...
+      );
+    }
+
+    if (!("panelTitle" in tab)) return entries;
+
+    entries.push([tab.panelTitle.l10nId, tab.panelTitle.text]);
+
+    for (const group of tab.panelGroups) {
+      if (group.titleL10nId) {
+        entries.push([group.titleL10nId, /** @type {string} */ (group.title)]);
```
LeoMcA, line 32:
```suggestion
      if (group.titleL10nId && group.title) {
        entries.push([group.titleL10nId, group.title]);
```

components/menu/l10n.js
```diff
@@ -0,0 +1,45 @@
...
+      }
+      for (const item of group.items) {
+        if (!("l10nId" in item)) continue;
+        entries.push([item.l10nId, item.text]);
+        if (item.labelL10nId) {
+          entries.push([item.labelL10nId, /** @type {string} */ (item.label)]);
```
LeoMcA, line 38:
```suggestion
        if (item.labelL10nId && item.label) {
          entries.push([item.labelL10nId, item.label]);
```

components/menu/server.js
```diff
@@ -69,27 +69,29 @@ export class Menu extends ServerComponent {
                   class="menu__tab-link"
                   href=${tab.href}
                   data-glean-id=${`menu_click_link: top-level -> ${tab.href}`}
-                  >${tab.buttonText}</a
+                  >${context.l10n(tab.buttonL10nId)}</a
```
LeoMcA: The plain `context.l10n` function doesn't work like this any more, perhaps that's an oversight from me.

You have to use `context.l10n.raw({ id: ...` if you're not providing the en-US text inline.

components/menu/tabs.js
```diff
@@ -0,0 +1,64 @@
...
+const RAW_TABS = Object.freeze(tabsDataRaw);
+
+/**
+ * Converts a string to a kebab-case slug for use in l10n IDs.
+ * @param {string} str
+ */
+const slugify = (str) =>
+  str
+    .toLowerCase()
+    .replaceAll(/[^\p{L}\p{N}]+/gu, "-")
```
LeoMcA, line 13: I'm not sure if fluent ids support non-ascii characters, perhaps simplify to `[^a-z0-9]`?

l10n/template.ftl
```diff
@@ -318,3 +318,124 @@ user-menu-feedback = Feedback
...
+menu-tools = Tools
+menu-tools-panel-title = Discover our tools
+menu-tools-0-playground = Playground
+menu-tools-0-http-observatory = HTTP Observatory
+menu-tools-1-border-image-generator = Border-image generator
+menu-tools-1-border-radius-generator = Border-radius generator
+menu-tools-1-box-shadow-generator = Box-shadow generator
+menu-tools-1-color-format-converter = Color format converter
+menu-tools-1-color-mixer = Color mixer
+menu-tools-1-shape-generator = Shape generator
```
LeoMcA, line 441: I don't think the number is necessary here, and just means we'll have unstable ids if we rearrange things

l10n/parser/extractor.js
```diff
@@ -48,6 +48,33 @@ export async function extract(options = {}) {
   }
 }
 
+/**
+ * Imports all `l10n.js` modules under the components directory and merges
+ * their exported `strings` Maps into a single Map.
+ *
+ * @param {string} componentsDir
```
LeoMcA, line 55: Don't think it's necessary to pass `componentsDir` as a param

l10n/template.ftl
```diff
@@ -318,3 +318,124 @@ user-menu-feedback = Feedback
 user-menu-user = User
 writer-open-editor-open-in-editor = Open in editor
 writer-toolbar-view-on-mdn = View on MDN
+menu-html = HTML
```
LeoMcA, line 328: As these strings don't exist in `en-US.ftl`, we don't actually have anything to render:

Image

I'd suggest that the extraction logic you've added should add them to `en-US.ftl`, then allow the usual `en-US.ftl` -> `template.ftl` logic to copy them here.

LeoMcA (changes requested):

l10n/template.ftl
caugner: The challenge is how to sync these. If we add them to `en-US`, how could we remove them automatically from there as well?

caugner: Leo's idea: Put them both in a separate `dynamic.ftl` file, and to `template.ftl, and make sure to load the dynamic Fluent file in the client in all locales as a fallback.

PR #1410, opened 2026-03-21
fix(search-modal): prevent focus ring from being clipped by scroll container

### LINKED ISSUE #992: Search box focus outline is getting clipped

Steps to reproduce:
1. Bring up the search box
2. type `border`
3. Click `TAB` key
4. You'll see the first suggestion getting an outline.
5. Notice the outline is clipped at the top and the right border.
6. Click `TAB` key one more time.
7. Notice the outline is clipped at the right border.
8. Click tab key till you reach the last suggestion.
9. Notice the outline is clipped at the bottom and the right border.

Refer to the following screen recording:

https://github.com/user-attachments/assets/24ff6bf8-0de4-44a0-bb19-d8bc2fa65341

The focus indicating outline is not clipped:
 
https://github.com/user-attachments/assets/d7afba48-1274-4aee-b0c2-279515a31c73

The above result was achieved using an ad hoc fix:
```css
a {
  ...
  width: calc(100% - 4px);
  margin: 2px;
}
```

_No response_

The issue is reproducible in the latest Linux Mint os on the latest Firefox and Chrome browsers.

low-perry:
### Description

Added an explicit `:focus-visible` rule to `components/search-modal/element.css`:

​```CSS
a:focus-visible {
  outline: 2px solid var(--color-link-normal);
  outline-offset: -2px;
}
​```

### Motivation

When navigating search results with the keyboard (Tab key), the
focus ring around a result link was visually clipped, cut off at the edges
instead of forming a clean rectangle around the item.

The root cause is a CSS clipping rule: any element with `overflow` set to a
value other than `visible` (here `overflow: auto` on the `<ul>` results list)
becomes a clipping container. The default browser focus outline renders
*outside* an element's border box, and because the result `<a>` links fill
100% of the scroll container's width, that outward outline sits exactly on
the clipping boundary and gets cut.

### Additional details

I tested the fix on Windows 11 and Ubuntu 22.04, using Firefox, Chrome, and Edge.

Fixes #992


components/search-modal/element.css
```diff
@@ -165,6 +165,11 @@ a {
   &:visited {
     color: var(--color-link-visited);
   }
+
+  &:focus-visible {
+    outline: 2px solid var(--color-link-normal);
```
caugner:
```suggestion
```

Instead, we might want to make the `border-color` of `li[data-selected]` conditionally on `:not(:focus-visible)`.

caugner (changes requested):
The outline looks good, except for the color (should not change) and the left (thicker than top/right/bottom).

[Before](https://developer.mozilla.org/en-US/):

Image

[After](https://fred-pr1410.review.mdn.allizom.net/en-US/):

Image

components/search-modal/element.css
low-perry: I tried conditioning the selected row border on `:not(:focus-visible)`, but in this component the focused element is the inner `a` rather than the `li`, so that did not produce a good result, especially in Firefox on Windows and Ubuntu.

caugner: Hm, in that case would `:not(:has(:focus-visible))` work?

low-perry: There’s a chance I misunderstood your suggestion the first time around. I’ve now updated `li[data-selected]` to use :`has(:focus-visible)` so the selected row drops its border color when the inner link is `focus-visible`, while keeping the link’s focus outline in place. 
I also tried removing the outline width, but without it the focus treatment looks a bit off in every browser I tested, so I kept the outline with the offset for now.

components/search-modal/element.css
```diff
@@ -165,6 +169,11 @@ a {
   &:visited {
     color: var(--color-link-visited);
   }
+
+  &:focus-visible {
+    outline: 2px solid;
+    outline-offset: -1px;
```
LeoMcA:
```suggestion
    outline-offset: -2px;
```

Otherwise it gets clipped on the right hand side.

components/search-modal/element.css
LeoMcA, line 152: I think cleanest solution is going to be to move this and other border definitions into the `a {}` block, and then remove the border (and add 2px of padding in its place) when `:focus-visible` is active.

LeoMcA (changes requested):

PR #1456, opened 2026-04-07
fix(search-modal): overflow documentElement when dialog is open

### LINKED ISSUE #725: Search dialog obscures content but still scrolls it

When the central BIG search field is clicked and open, there is a display change to the field and screen background when the viewport width is about 1007px or less.

- The screen background turns black 
- The search field aligns top
- The search field height is very small
-  There screen scrolls vertically but there is nothing to see (it's all black)

I expect to see the same display of a blurred background, and the search field to remain centered and large as it does for viewport width greater than 1007px.  When I scroll vertically, I see the blurred content scrolling.

This is the BIG search to click
Image

Opens as such
Image

Flips to this on 1007px or less
Image

Thank you!

caugner:
### Description

Update the `search-modal`, making the `<html>` element overflow when the dialog is open.

### Motivation

Avoid that scrolling on the backdrop scrolls the page content in the background.

Fixes https://github.com/mdn/fred/issues/725.

### DIFF
--- a/components/search-modal/global.css
+++ b/components/search-modal/global.css
@@ -1,0 +1,4 @@
+:root.search-modal-open {
+  overflow: hidden;
+  scrollbar-gutter: stable;
+}
--- a/components/search-modal/global.css
+++ b/components/search-modal/global.css
@@ -1,4 +1,7 @@
-:root.search-modal-open {
-  overflow: hidden;
+:root {
   scrollbar-gutter: stable;
+
+  &.search-modal-open {
+    overflow: hidden;
+  }
 }
--- a/components/search-modal/element.js
+++ b/components/search-modal/element.js
@@ -209,6 +209,14 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
     this.shadowRoot?.querySelector("dialog")?.close();
   }
 
+  /** @param {ToggleEvent} event */
+  _toggle({ newState }) {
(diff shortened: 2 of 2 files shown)


components/search-modal/global.css
```diff
@@ -0,0 +1,4 @@
+:root.search-modal-open {
+  overflow: hidden;
+  scrollbar-gutter: stable;
```
LeoMcA, line 2: I think we need to elevate this to the always-on global styles (preferably in another folder), so that we don't get into a situation like:

- modal is closed, page isn't overflowing, implicit `scrollbar-gutter: auto;` behaviour means there is no scroll gutter
- model is opened, `scrollbar-gutter: stable;` behaviour means a scroll gutter is added - oops the page jumps around

LeoMcA (approved):
Nice! One comment:

components/search-modal/global.css
caugner: I don't think we have any page that doesn't scroll, but how about:

```css
:root {
  scrollbar-gutter: stable;

  &.search-modal-open {
    overflow: hidden;
  }
}

```

We include the `<mdn-search-modal>` unconditionally in `Navigation`, so this should apply to apply pages that can show the search modal?!

caugner: See 1a04d0a2.

PR #1458, opened 2026-04-08
fix(blog-index): correct grid layout

pepelsbey:
### Description

Fixes blog layout bug in Safari. Chrome and Firefox are fine.

### Motivation

| Before | After |
| - | - |
| desktop-before | desktop-after |
| mobile-before | mobile-after |

### DIFF
--- a/components/blog-index/server.css
+++ b/components/blog-index/server.css
@@ -68,7 +68,7 @@
   display: grid;
 
   grid-template-rows: subgrid;
-  grid-row: span 6;
+  grid-row: span 5;
 
   padding: 2rem;
   margin-top: 2rem;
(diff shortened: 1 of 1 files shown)


components/blog-index/server.css
```diff
@@ -68,7 +68,7 @@
   display: grid;
 
   grid-template-rows: subgrid;
-  grid-row: span 6;
+  grid-row: span 5;
```
pepelsbey, line 71: There are only 5 elements in the card

components/blog-index/server.css
```diff
@@ -78,13 +78,6 @@
 }
 
 .blog-post-preview__header {
-  display: grid;
-
-  flex-direction: column;
-
-  grid-template-rows: subgrid;
-  grid-row: span 3;
-
```
pepelsbey, line 87: It’s a weird mix of grid/flex that doesn’t do much.

caugner (approved):

PR #1464, opened 2026-04-10
chore(media): reorder width-first in two-part queries

pepelsbey:
### Description

Reorders two-part queries width-first.

### Motivation

It’s difficult to read number-first queries: “number less than variable” sounds like a riddle.

```diff
- 426px < width
+ width > 426px
```

Having width in the middle still makes sense in three-part queries: “variable is in between numbers”.

```
992px < width <= 1200px
```

### DIFF
--- a/components/media/index.css
+++ b/components/media/index.css
@@ -2,23 +2,23 @@
 
 @custom-media --screen-xsmall (width <= 426px);
 
-@custom-media --screen-small-and-wider (426px < width);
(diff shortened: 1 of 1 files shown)


LeoMcA (approved):
Personally I view the current code like a number line. (`width` is to the right, therefore larger) And the consistency with the other queries feels nice. But I'll live :)

caugner:
> Personally I view the current code like a number line.

Agree, I too found the previous writing clearer.

PR #1475, opened 2026-04-13
fix(a11y): add `aria-label` where `title` is used

### LINKED ISSUE #959: Accessible tooltips
This is a follow up from https://github.com/mdn/content/issues/987#issuecomment-768313842

So far, I've noticed the usage of `title` attribute in some places on MDN web docs.
Those are not accessible for non-hover devices or screenreaders.

However, its usage hint that there is some value in it.
Thus, I'd like us to explore the use cases and constraints of those tooltips.

An accessible and inclusive markup is suggested in https://inclusive-components.design/tooltips-toggletips/ which could be use as a starting point.

I'm not sure, whether an implementation would end up in Yari, Content or Minimalist. (Likely a combination of all of them - design in minimalist, hints in Content which then would be transformed by Yari).

caugner:
### Description

Add `aria-label` where `title` is used.

Also adds an ESLint rule to ensure this going forward.

### Motivation

`title` is not accessible to screen-readers.

Fixes https://github.com/mdn/fred/issues/959.


build/eslint-fred.js
```diff
@@ -116,6 +116,57 @@ export default {
         };
       },
     },
+    "title-needs-aria-label": {
+      meta: {
+        type: "problem",
+        docs: {
+          description:
+            "Enforce that HTML elements with a title attribute also have aria-label set to the same value.",
```
LeoMcA: nit: might be nice to leave a comment to the issue/a webpage which advises taking this approach

build/eslint-fred.js
```diff
@@ -116,6 +116,57 @@ export default {
...
+            // Expressions are replaced with "__EXPR__" so no stray > chars.
+            const tagPattern = /<([a-z][a-z0-9-]*)(\s[^>]*)?\/?>/gis;
+            let match;
+
+            while ((match = tagPattern.exec(reconstructed)) !== null) {
+              const [fullTag, tagName] = match;
+
+              // Skip elements where title has standardised HTML semantics
+              // (abbr expansion) or where aria-label is not applicable (link).
+              if (tagName === "abbr" || tagName === "link") continue;
```
LeoMcA: nit: Perhaps elevate to an `EXCEPTIONS` constant, in case we identify tags we don't want to apply this to in future?

build/eslint-fred.js
```diff
@@ -116,6 +116,57 @@ export default {
...
+              .map((q) => q.value.cooked ?? q.value.raw)
+              .join('"__EXPR__"');
+
+            // Match opening HTML tags; attributes may span multiple lines.
+            // Expressions are replaced with "__EXPR__" so no stray > chars.
+            const tagPattern = /<([a-z][a-z0-9-]*)(\s[^>]*)?\/?>/gis;
+            let match;
+
+            while ((match = tagPattern.exec(reconstructed)) !== null) {
+              const [fullTag, tagName] = match;
```
LeoMcA: This is a bit esoteric, can we use `String.matchAll(tagPattern)` instead?

build/eslint-fred.js
```diff
@@ -116,6 +116,57 @@ export default {
...
+
+              // Skip elements where title has standardised HTML semantics
+              // (abbr expansion) or where aria-label is not applicable (link).
+              if (tagName === "abbr" || tagName === "link") continue;
+
+              if (!/(?:^|\s)title\s*=/.test(fullTag)) continue;
+              if (/(?:^|\s)aria-label\s*=/.test(fullTag)) continue;
+
```
LeoMcA, line 174: a generic regex to find attributes might be nicer, we should also check if `aria-labelledby` is set, and not warn in that instance

LeoMcA (approved):

build/eslint-fred.js
caugner: 46df307296bbcf46f7b640bc854acdc1a38d94a5 + b2595a1e7cf52d9dca1243f6f8a8879df083611a

build/eslint-fred.js
caugner: fe0874bd80e5d324f71f2af6b198dbcab83a4594

build/eslint-fred.js
caugner: 20c30e0a39e6bb35fc4d54b5ea4f6ee424fc74f1

build/eslint-fred.js
caugner: fc745d81ebc308fe984701ba496dbb70ec2733fc + b11f2389a9ce21ae9c8618ed697fd21dc7dde61d (+ a55c142999504d3f08bd41360e2b3615785117f8)

build/eslint-fred.js
caugner: Also updated the rule message in 855316dc to include the bit from the spec about user agents not exposing the attribute in an accessible manner.

components/compat-table/element.js
```diff
@@ -418,6 +427,7 @@ export class MDNCompatTable extends L10nMixin(LitElement) {
 
         const hasHistory = notes.length > 0;
         const isExpanded = hasHistory && this._showTimelineId == timelineId;
+        const toggleTitle = ifDefined(hasHistory && "Toggle history");
```
LeoMcA, line 430: Bug from before, but I think this'll evaludate to title=false in DOM - need:


```suggestion
        const toggleTitle = ifDefined(hasHistory ? "Toggle history" : undefined);
```

components/baseline-indicator/server.js
```diff
@@ -167,26 +167,28 @@ export class BaselineIndicator extends ServerComponent {
...
-              </span>`,
-          )}
+          ${ENGINES.map(({ name, browsers }) => {
+            const title = engineTitle(browsers);
+
+            return html`<span
+              key=${name}
+              class="engine"
+              title=${title}
+              aria-label=${title}
```
LeoMcA, line 177: `<span>` is `role="generic"` by default, and `aria-label` is [prohibited/ignored on `generic` elements](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/generic_role#:~:text=Because%20the%20generic%20role%20is%20nameless%2C%20the%20aria%2Dlabelledby%20and%20aria%2Dlabel%20attributes%20are%20prohibited%2E).

So perhaps we'll need a list of GENERIC_TAGS, which we ignore, unless role is set to a non-ignored value.

And we ignore all roles in this list: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-label#associated_roles

components/compat-table-lazy/element.js
```diff
@@ -60,15 +60,18 @@ export class MDNCompatTableLazy extends L10nMixin(LitElement) {
...
       href="#"
       @click=${onClick}
       target="_blank"
       rel="noopener noreferrer"
-      title=${this.l10n(
-        "compat-link-report-missing-title",
-      )`Report missing compatibility data`}
+      title=${title}
+      aria-label=${title}
       >${this.l10n("compat-link-report-missing")`Report this issue`}</a
```
LeoMcA, line 75: This linter might be too much of a blunt instrument, there's a number of examples like this where we have a label shown in UI, but then an even more descriptive one in `aria-label`.

This fails https://www.w3.org/WAI/WCAG21/Understanding/label-in-name: AIUI for users using visible labels, but speech input, they'll say something like "click report this issue", but this element will be represented within the accessibility tree as "Report missing compatibility data" and they won't be able to.

> [For this reason, when a visible label already exists, aria-label should be avoided or used carefully, and aria-labelledby should be used as a supplement with care.](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name#:~:text=For%20this%20reason%2C%20when%20a%20visible%20label%20already%20exists%2C%20aria%2Dlabel%20should%20be%20avoided%20or%20used%20carefully%2C%20and%20aria%2Dlabelledby%20should%20be%20used%20as%20a%20supplement%20with%20care)

So perhaps our original premise: use `aria-label` everywhere we use `title` is a false one? Let's discuss

components/compat-table/element.js
caugner: I thought `undefined` renders as `"undefined"`, but IIRC I've seen `nothing` do the trick:

```suggestion
        const toggleTitle = hasHistory ? "Toggle history" : nothing;
```

PR #1488, opened 2026-04-16
style(eslint): disallow title attribute

caugner:
### Description

Adds new ESLint rule `a11y-no-title-attribute` that disallows the HTML `title` attribute in general (except for iframes where it is encouraged).

### Motivation

Avoid accessibility issues.

### Additional details

For now, the rule is only a warning to raise awareness.

Derived from https://github.com/mdn/fred/pull/1475.

### DIFF
--- a/build/eslint-fred.js
+++ b/build/eslint-fred.js
@@ -116,6 +116,65 @@ export default {
         };
       },
     },
+    "a11y-no-title-attribute": {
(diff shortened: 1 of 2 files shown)


caugner:
After discussion, closing in favor of https://github.com/mdn/fred/pull/1475.

PR #1500, opened 2026-04-21
feat(baseline-indicator): add renderSimplified method, avoid rendering repetitive links

LeoMcA:
Currently most pages rendered to simplified HTML (and then markdown) have a lot of noise at the top, with Baseline learn more/feedback and BCD table links (and the BCD table isn't currently rendered, so the link is useless).

This adds a `renderSimplified` method, returning a simpler output. Extracts some common logic between the two render methods.

Tested locally with `FRED_SIMPLE_HTML=true npm start`:

http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array
```
Baseline Widely available *
This feature is well established and works across many devices and browser versions. It’s been available across browsers since July 2015.
* Some parts of this feature may have varying levels of support.
```

http://localhost:3000/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/length
```
Baseline Widely available
This feature is well established and works across many devices and browser versions. It’s been available across browsers since July 2015.
```

http://localhost:3000/en-US/docs/Web/API/CSS_Custom_Highlight_API
```
Baseline 2025 Newly available
Since June 2025, this feature works across the latest devices and browser versions. This feature might not work in older devices or browsers. 
```

http://localhost:3000/en-US/docs/Web/API/Speculation_Rules_API
```
Limited availability
This feature is not Baseline because it does not work in some of the most widely-used browsers. 
```

### DIFF
--- a/components/baseline-indicator/server.js
+++ b/components/baseline-indicator/server.js
@@ -51,13 +51,11 @@ export class BaselineIndicator extends ServerComponent {
   static inlineScript = inlineScript;
 
   /**
-   * @param {import("@rari").Baseline} status
+   * @param {string | null | undefined} date
    */
-  parseDate(status) {
-    const lowDateRange = status.baseline_low_date?.match(/^([^0-9])/)?.[0];
-    return status.baseline_low_date
-      ? new Date(status.baseline_low_date.slice(lowDateRange ? 1 : 0))
-      : undefined;
+  parseDate(date) {
+    const lowDateRange = date?.match(/^([^0-9])/)?.[0];
+    return date ? new Date(date.slice(lowDateRange ? 1 : 0)) : undefined;
   }
 
   /**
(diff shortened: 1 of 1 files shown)


components/baseline-indicator/server.js
```diff
@@ -51,7 +51,45 @@ export class BaselineIndicator extends ServerComponent {
   static inlineScript = inlineScript;
 
   /**
-   *
+   * @param {import("@rari").Baseline} status
+   */
+  parseDate(status) {
```
caugner: Shouldn't this function receive `date`?

components/baseline-indicator/server.js
```diff
@@ -71,11 +109,7 @@ export class BaselineIndicator extends ServerComponent {
       LOCALIZED_BCD_IDS[context.locale] || LOCALIZED_BCD_IDS[DEFAULT_LOCALE]
     }`;
 
-    const low_date_range = status.baseline_low_date?.match(/^([^0-9])/)?.[0];
-    const low_date = status.baseline_low_date
-      ? new Date(status.baseline_low_date.slice(low_date_range ? 1 : 0))
-      : undefined;
-
+    const lowDate = this.parseDate(status);
```
caugner:
```suggestion
    const lowDate = this.parseDate(status.baseline_low_date);
```

caugner (commented):
LGTM, except for the `parseDate()` signature.

caugner (approved):

PR #1504, opened 2026-04-21
feat(specifications): render a list in simplified mode

LeoMcA:
It would've taken a whole lot of repetition (or overhead plumbing) to DRY `render` and `renderSimplified` here, so I've added a new `simplifiedMode` property which can be used instead.

This also gives us the flexibility to turn `simplifiedMode` into an enum in the future to specify different simplified targets, imagine something like:

```
renderSimplified() {
  switch(this.simplifiedMode) {
    case "mcp":
        return ...
    case "standalone-zip":
        return ...
    case "carrier-pigeon":
        return ...
```


components/specifications-list/server.js
```diff
@@ -0,0 +1,73 @@
...
+    const links = urlsByTitle
+      .entries()
+      .flatMap(([title, urls]) =>
+        urls.map((url) => this.renderLink(url, title)),
+      );
+
+    return this.simplifiedMode
+      ? html`<ul>
+          ${links.map((link) => html`<li>${link}</li>`)}
+        </ul>`
```
caugner: nit: Could we make this an `if (this.simplifiedMode) { return html`…`; }` block instead? Even though it's obvious, I had to look twice to see which branch is which.

caugner (approved):
LGTM, just one nit.

PS: If we document simplifiedMode somewhere, we could mention this (alternative) strategy.

PR #1532, opened 2026-05-04
fix(content-section): add styling for good/bad plain examples

### LINKED ISSUE #948: (design) Good and Bad examples are not stylized

Example blocks using `plain` text mode do not apply “green/red” formatting when using associated keywords.

Image

That syntax highlighting for code sample blocks is once again functional for blocks in text mode.

Example of MDN Macro Syntax detection:

Image

Example of valid color (without icon, but normaly there is a cross-circle and a check-circle icons for the version from code)

Image

What we can see in blocks with a code name:

Image

Page used for example: https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Scripting/Variables#an_aside_on_variable_naming_rules

The possible cause is that `.example-bad:is(.code-example pre)` is not detected on this type of block because there is no `.code-example` encapsulating the `pre` element.

caugner:
### Description

Add styling for good/bad plain examples.

### Motivation

Fix regression in Fred where good and bad plain examples cannot be distinguished.

Fixes https://github.com/mdn/fred/issues/948.

### DIFF
--- a/components/code-example/common.css
+++ b/components/code-example/common.css
@@ -56,8 +56,8 @@
 
         flex-shrink: 0;
 
-        width: 1.24rem;
-        height: 1.24rem;
+        width: var(--font-size-medium);
+        height: var(--font-size-medium);
 
         margin: 1.255rem 1rem;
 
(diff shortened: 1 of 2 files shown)


LeoMcA (approved):
I think it would be nice to DRY the css across the two components, but this is good enough for now!

PR #1534, opened 2026-05-05
fix(content-section): add min-width to headers of properties table

### LINKED ISSUE #969: First column in Technical summary table for HTML elements is squished
See: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/option#technical_summary

__Description__
There's an unnecessary amount of constrained space for the `<th>` within the table that comes after the first description of an HTML element. It's table that has information such as "Content categories" and "Permitted content". Current state screen shot is below.

Content within the table headers of the table are forced to be squished

__Proposed solution__
Here are a couple screen shots of things I tried.

1. Giving the `<th>` a `min-width: 20%`:

Content within table headers of the table after 20% min width allow more content per line and less wrapping

2. Giving the `<th>` a `min-width: 20%` as well as the same padding as the `<td>` (`padding: .5rem 0 .5rem .75rem) except for the right side since the `<td>` already has padding on the left:

Content within table headers of the table after 20% min width allow more content per line and less wrapping

TGianella:
### Description

Add `min-width: 25%` to `th` of `.properties` tables (used in "Technical summary" section of HTML elements documentation). I tried `20%` but most headers would still wrap.

### Motivation

The headers column was squished and text wrapped for no reason.

### Additional details

I noticed while working on this that the properties table lack a top margin so there is no spacing with the "Technical summary" header (see attached image). I could fix in this PR or open another PR.

image

Fixes #969

### DIFF
--- a/components/content-section/server.css
+++ b/components/content-section/server.css
@@ -174,6 +174,7 @@
       }
 
       th {
+        min-width: 25%;
         font-weight: var(--font-weight-bold);
         background-color: transparent;
       }


LeoMcA (approved):
Nice, thank you!

> I noticed while working on this that the properties table lack a top margin so there is no spacing with the "Technical summary" header (see attached image). I could fix in this PR or open another PR.

Please open another PR: I've filed an issue you can attach to: https://github.com/mdn/fred/issues/1535

PR #1536, opened 2026-05-05
fix(writer-mode): improve handling of _open route

### LINKED ISSUE #906: Improve _open editor route handling
https://github.com/mdn/fred/blob/eab0c156d5c5138d19381464cbf759cf3606b911/server.js#L185-L197

We don't `res.end` or send a status, leaving the connection dangling.
We don't catch errors, which will kill the server.
We don't log any feedback like "attempting to open ... with ...".

TGianella:
### Description

* Return a status code (`204 no content`) as to not leave the connection dangling.
* Add a log in console when trying to open the editor.
* Spawn manually the process to open the editor so we can tap on errors and handle them without crashing the server.

### Motivation

The route handling is incomplete at the moment, the connection is never closed, errors are not handled, and nothing is logged while this can be tricky to debug or investigate.

Closes #906

### DIFF
--- a/server.js
+++ b/server.js
@@ -6,7 +6,7 @@ import { Worker } from "node:worker_threads";
 import cookieParser from "cookie-parser";
 import express from "express";
 import { createProxyMiddleware } from "http-proxy-middleware";
-import openEditor from "open-editor";
+import { getEditorInfo } from "open-editor";
 
 import { FRED_BUILD_ROOT } from "./build/env.js";
 import {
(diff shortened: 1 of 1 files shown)


TGianella:
@LeoMcA This PR lacks the error handling part of #906 because it is the trickiest. `openEditor` spawns a detached async process so the error cannot be caught in a try/catch block, but it also doesn't provide any hook to listen for errors. The best way to catch errors would be to use `getEditorInfo` and then spawn the process manually so we can attach a listener for errors, something like this:
```js
const { binary, arguments: args } = getEditorInfo([absolutePath]);
const child = spawn(binary, args, { detached: true, stdio: 'ignore' });

child.on('error', (error) => {
  console.error('Failed to open editor:', error);
  return;
});
```
It seems a bit involved though for something that is only used during development so I seek confirmation before implementing that.

What do you think?

LeoMcA:
@TGianella the test case I'd like to satisfy is using the route with a command like `EDITOR=foobar FRED_WRITER_MODE=true npm run start` doesn't cause the whole server to crash - your proposal looks like a good, simple, way of handling that!

LeoMcA (approved):
Nice, thank you!

PR #1548, opened 2026-05-07
fix(search): scroll only results

### LINKED ISSUE #1474: Wrong section scrolls in search results
Follow up to https://github.com/mdn/fred/pull/1455

Image

Only the results list should scroll, rather than the whole modal (including the input area).

TGianella:
### Description

Migrate the search modal dialog from grid to flexbox layout and add a wrapper div around the search input and close button to keep the same UI in the new layout system.

### Motivation

In the current grid layout, `min-content` on the search results row was causing the grid track to be as high as the whole `ul` containing the search results, so the `ul` itself couldn't be a scroll container because its content wasn't overflowing its height. Hence the next parent (`dialog`) with a capped width became the scroll container and the whole dialog scrolled (with the input field).

As @LeoMcA pointed out, the easy fix of using 1fr doesn't work on WebKit (because they have a different implementation of computing what 1fr means when there is no explicit height). 

I could not find a solution which worked while keeping the grid layout so I propose to migrate to flexbox to resolve the issue on all browsers.

> [!WARNING]
> I tested on Firefox/Safari/Chrome both small and large breakpoints with empty search results, few results (not overflowing) and overflowing search results. With my limited knowledge about the search modal specs, I did not find any regression except that the close button on the dev server falls back to the text because the icon is not loaded, and in the new layout the text doesn't break and grows to take as much space as needed to render on one line:
>| Before | After |
>|--------|-------|
>| image | image |
>
> This is a minor issue that only appears if the icon doesn't load.

Closes #1474


LeoMcA (changes requested):
Unfortunately this causes weird behaviour in webkit:

Image

TGianella:
Thanks for the review @LeoMcA and sorry for not checking on WebKit and wasting your time.

To the best of my ability I could not find a solution that would fix the bug in all supported browsers while keeping the current grid layout and HTML structure.

I pushed a new commit migrating the search modal to flexbox and updated the description of the PR accordingly, if you do not agree with the solution you can close this PR directly because I won't have a better solution. :slightly_smiling_face:

LeoMcA (approved):
Sorry for the long delay reviewing this, looks great - no need to use grid if flex suffices (and fixes a bug while we're at it!). Thank you!

PR #1550, opened 2026-05-07
fix(content-section): add margin before table

### LINKED ISSUE #1535: No padding between subheadings and tables
In https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/option#technical_summary the "Technical summary", "Specifications" and "Browser compatibility" headings all have insufficient padding between them and the tables which follow.

I imagine this was caused by https://github.com/mdn/fred/pull/1426

caugner:
### Description

Adds a margin before tables.

### Motivation

Prevent tables from sticking to the preceding paragraph.

### Additional details

| [Before](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/color#formal_definition) | [After](https://fred-pr1550.review.mdn.allizom.net/en-US/docs/Web/CSS/Reference/Properties/color#formal_definition) |
|--------|--------|
| image | image | 

Fixes https://github.com/mdn/fred/issues/1535.


components/content-section/server.css
```diff
@@ -182,6 +182,7 @@
 
   table {
     width: 100%;
+    margin: 1rem 0;
```
LeoMcA: Let's set this on `.table-container` so margins can collapse with other tags, otherwise it looks a bit weird: https://github.com/mdn/fred/issues/1535#issuecomment-4397698509

caugner: e19cbbf

LeoMcA (approved):

PR #1183, opened 2026-01-07
chore(l10n): update french strings

tristantheb:
### Description

* Fixed Gremlins alerts for em-short to em-long
* Fixed gendering to use H+F in lines
* Fixed string order (word position change in french)
* Added new strings

### Motivation

Further promote inclusivity, as was already done in a previous interface translation (< 2021). And apply the French teams' policy on gender inclusivity, as applied in Pontoon by the French team.

### Additional details

> [!NOTE]
> Need an engineer's opinion/observation

Strange behavior was observed with the translation file, before and after modifications, for browser compatibility, when a browser has a flag.

When a flag is present, as in `@media/inverted-colors`, it is displayed correctly in English, but in French, part of the translation is triggered that should not be.

> From version ⁨114⁩⁩⁨⁨ ⁩users⁩ ⁨ ⁩must explicitly set the ⁨layout.css.inverted-colors.enabled⁩⁨ ⁩ ⁨preference⁩⁨⁨ ⁩to ⁨true⁩⁩⁨.⁩ ⁨To change preferences in ⁨Firefox⁩, visit ⁨about:config⁩.⁩

> ⁨De la version ⁨114⁩⁩⁨⁨ **⁩jusqu'à ⁨{$versionLast}⁩ des utilisateur·ice·s**⁩ ⁨ ⁩doit explicitement définir ⁨la préférence⁩ ⁨ ⁩⁨layout.css.inverted-colors.enabled⁩ ⁨⁨ ⁩à ⁨true⁩⁩⁨.⁩ ⁨Pour changer vos préférences sur le navigateur ⁨Firefox⁩, visitez ⁨about:config⁩.⁩

If you can see what I didn't see, I'm interested 😅 I double-checked the conditions to make sure I hadn't forgotten a character when I first copied it two months ago, but everything is correct...

### Related issues and pull requests

_none_

PR #1562, opened 2026-05-12
fix(l10n): correct compat-support-flags rendering across locales

### LINKED ISSUE #1220: The translation of the BCD table presents problems with Fluent variables

During translation, it has been observed that some elements are visible even though the en-US page does not display them, and sometimes elements disappear completely or are not displayed correctly.

| French | English |
| --- | --- |
| Image | Image |

([HTMLSelectElement : méthode showPicker()](https://developer.mozilla.org/fr/docs/Web/API/HTMLSelectElement/showPicker) | [HTMLSelectElement: showPicker() method](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/showPicker))

For another example, see #1183 with `@media/inverted-colors`

### Steps to reproduce

1. Open the page in the examples below for FR and EN.
2. Go to **Browser compatibility** ("Compatibilité des navigateurs").
3. Open a flagged compatibility or experimental compatibility cell.
4. Some elements hidden by Fluent in en-US are displayed in FR.

Normaly, the french content need to display "Utilisateur·ice·s ⁩doit explicitement définir ⁨la préférence ⁨⁨` ` à ⁨`Enabled⁩⁩⁨`.⁩ ⁨⁩" (for the example with `HTMLSelectElement.showPicker()`) like it's displayed in english.

[HTMLSelectElement : méthode showPicker()](https://developer.mozilla.org/fr/docs/Web/API/HTMLSelectElement/showPicker) (fr)
[HTMLSelectElement: showPicker() method](https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/showPicker) (en)

https://developer.mozilla.org/fr/docs/Web/CSS/Reference/At-rules/@media/inverted-colors
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@media/inverted-colors

_none_

caugner:
### Description

Fixes two bugs in the BCD flag footnote rendering: a CLDR plural-rule mismatch in `compat-support-flags`, and HTML special chars in flag names breaking the sanitizer.

### Motivation

1. `has_added`/`has_last`/`has_value`/`has_pref_url` are boolean 0/1 args, but the FTL selectors used CLDR plural variants (`[one]` / `*[other]`). French (and others) treat 0 as `one`, so the wrong branch fired — producing extra text and a literal `{ $versionLast }`. Switch to exact-integer variants (`[1]` / `*[0]`).
2. Flag names like `<select> showPicker() method` got parsed by `insane` after Fluent substitution, eating content up to the next close tag (empty `<code>` element). HTML-escape string args before `formatPattern`.

Fixes https://github.com/mdn/fred/issues/1220.

Also resolves the residual variable-rendering quirk noted in https://github.com/mdn/fred/issues/1183.

### DIFF
--- a/l10n/fluent.js
+++ b/l10n/fluent.js
@@ -165,12 +165,45 @@ export class Fluent {
 
     /** @type {Error[]} */
     const errors = [];
-    const formatted = bundle?.formatPattern(message, args, errors);
+    const formatted = bundle?.formatPattern(
+      message,
+      /** @type {Record<string, import("@fluent/bundle").FluentVariable>} */ (
+        escapeArgs(args)
+      ),
+      errors,
+    );
     if (errors.length > 0) {
       console.error(errors);
     }
     return formatted;
   }
+}
+
+/**
+ * HTML-escape string-valued args so that variable substitution can't inject
+ * tags that confuse the downstream sanitizer. Only string values are touched
+ * so that Number/Date selectors keep working.
+ *
+ * @param {Record<string, unknown>} args
+ * @returns {Record<string, unknown>}
+ */
+function escapeArgs(args) {
+  /** @type {Record<string, unknown>} */
+  const out = {};
+  for (const [k, v] of Object.entries(args)) {
+    out[k] = typeof v === "string" ? escapeHtml(v) : v;
+  }
+  return out;
+}
+
+/**
+ * @param {string} s
+ */
+function escapeHtml(s) {
+  return s
+    .replaceAll("&", "&amp;")
+    .replaceAll("<", "&lt;")
+    .replaceAll(">", "&gt;");
 }
 
 /** @type {Map<string, Fluent>} */
--- a/l10n/fluent.js
+++ b/l10n/fluent.js
@@ -57,7 +57,7 @@ export class Fluent {
   /**
    * @param {string} id
    * @param {string} [attr]
-   * @param {Record<string, any>} [args]
+   * @param {Record<string, FluentVariable | undefined>} [args]
    * @param {Record<string, import("../types/fluent.js").Element>} [elements]
    * @returns {string | ReturnType<typeof unsafeHTML> | undefined}
    */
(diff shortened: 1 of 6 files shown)


l10n/fluent.js
```diff
@@ -165,14 +165,47 @@ export class Fluent {
 
     /** @type {Error[]} */
     const errors = [];
-    const formatted = bundle?.formatPattern(message, args, errors);
+    const formatted = bundle?.formatPattern(
+      message,
+      /** @type {Record<string, import("@fluent/bundle").FluentVariable>} */ (
```
LeoMcA: Let's fix the types a little bit, remove the cast here, and set `args` up in the `getMessage` function to `@param {Record<string, import("@fluent/bundle").FluentVariable>} [args]`

l10n/fluent.js
```diff
@@ -165,14 +165,47 @@ export class Fluent {
...
+/**
+ * HTML-escape string-valued args so that variable substitution can't inject
+ * tags that confuse the downstream sanitizer. Only string values are touched
+ * so that Number/Date selectors keep working.
+ *
+ * @param {Record<string, unknown>} args
+ * @returns {Record<string, unknown>}
+ */
+function escapeArgs(args) {
+  /** @type {Record<string, unknown>} */
```
LeoMcA: I prefer implicit returns types where possible, but perhaps that's personal preference. Let's use `FluentVariable` rather than `unknown` - the changes I recommend in `getMessage` will ensure the args are always of that type:

```suggestion
 * @param {Record<string, import("@fluent/bundle").FluentVariable>} args
 */
function escapeArgs(args) {
  /** @type {Record<string, import("@fluent/bundle").FluentVariable>} */
```

LeoMcA (approved):
Looks good, let's just tighten up the types a little bit.

Would be best to `@import` `FluentVariable` so we don't inline import it multiple times:

l10n/fluent.js
caugner: b50606a4

l10n/fluent.js
caugner: b50606a4

LeoMcA (approved):
The types could be tightened a little more, and I took the liberty of doing + pushing rather than trying to explain through comments - hope that's ok! :)

I'll approve, but let you merge if my changes look good

caugner:
> The types could be tightened a little more, and I took the liberty of doing + pushing rather than trying to explain through comments - hope that's ok! :)
> 
> I'll approve, but let you merge if my changes look good

Nice, looking good, thanks!

I'll merge in a second.

PR #1563, opened 2026-05-12
fix(content-section): display index in multiple columns

### LINKED ISSUE #779: Display indexes in 3 columns
In yari, indexes like the ones on https://developer.mozilla.org/en-US/docs/Web/CSS/Reference#index were shown in 3 columns, rather than 1 long one.

caugner:
### Description

Updates the `content-section` style for index lists (`.index`) to use multiple columns, according to space.

### Motivation

Restore the layout from Yari times, reducing the length of the page.

### Additional details

Rather than fixing the column count, we use a minimum column width.

Fixes https://github.com/mdn/fred/issues/779.

### DIFF
--- a/components/content-section/server.css
+++ b/components/content-section/server.css
@@ -380,6 +380,22 @@
     }
   }
 
+  /* Alphabetical/grouped indexes (rari `CSS_Ref`, `ListGroups`, `APIListAlpha`, `css_ref_list`) */
+  .index {
+    columns: 18rem;
+    column-gap: 2rem;
+
+    h3 {
+      margin-top: 0;
+      break-after: avoid;
+    }
+
+    ul {
+      margin-top: 0.5rem;
+      break-inside: avoid;
+    }
+  }
+
   /* Article footer */
   &.article-footer {
     margin: 2rem 0;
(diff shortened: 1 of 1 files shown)


components/content-section/server.css
```diff
@@ -380,6 +380,22 @@
...
+
+    h3 {
+      margin-top: 0;
+      break-after: avoid;
+    }
+
+    ul {
+      margin-top: 0.5rem;
+      break-inside: avoid;
+    }
```
LeoMcA:
```suggestion
    ul {
      columns: 18rem;
      column-gap: 2rem;
    }

    li {
      break-inside: avoid;
    }
```

LeoMcA (changes requested):
Nice job making the columns dynamic, but my memory from yari times is that we had columns per-heading: it's not super useful having to scroll up to the very top of the page halfway through the alphabet, nor does it present the information in a usefully more dense way.

With my suggested change applied, it does that:

Image

caugner:
> we had columns per-heading

Much better, indeed:

- https://fred-pr1563.review.mdn.allizom.net/en-US/docs/Web/API#specifications
- https://fred-pr1563.review.mdn.allizom.net/en-US/docs/Web/CSS/Reference#index

LeoMcA (approved):

PR #1600, opened 2026-05-22
ci(npm-publish-simulation): don't clean cache

LeoMcA:
Two changes aimed at speeding up this test on Windows:

- On Windows specifically, move the npm cache to the faster D drive (which appears to be local SSD storage, rather than some kind of NAS)
- Across all platforms, no longer clear the npm cache when installing the built fred package, and instead rely on a git hash embedded in fred's package.json to ensure we haven't got the wrong version installed.

Test run showing that a non-matching hash (set to "foobar" in the package) fails: https://github.com/mdn/fred/actions/runs/26455922611/job/77889989841?pr=1600

Comparing CI run times is hard, because they're incredibly variable: indeed, it seems like the runs are a few seconds slower on macos! But ubuntu seems to be just as fast, and windows is clearly faster: going from 5-7 min runs down to consistent sub-4 minutes.


.github/workflows/npm-publish-simulation.yml
```diff
@@ -81,14 +86,18 @@ jobs:
...
           npm install file:$TARBALL
           npm install
         env:
           # Increase GitHub API limit.
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
+      - name: (mdn/content) Verify correct fred installed
+        working-directory: mdn/content/node_modules/@mdn/fred
+        shell: bash
+        run: test "$(npm pkg get _gitSha --json)" = '"${{ github.sha }}"'
```
caugner, line 99: Given that we install Fred from the tarball, why should we ever get the wrong Fred version?

LeoMcA: I've observed weird behaviour locally where re-running `npm pack` after changes and re-installing that tarball would still use an old, cached, version. Perhaps this was a yarn limitation, but I think it makes sense to test/verify that we're not getting an older tarball from the cache.

caugner (approved):

PR #1649, opened 2026-06-12
build(deps-dev): update eslint-unicorn and fix/disable rules

LeoMcA:
Splitting from https://github.com/mdn/fred/pull/1645 so claude can have a review


LeoMcA:
No comments, so it looks like claude is happy

globau (approved):

Issue #1079, opened 2025-11-13
quicksearch in browser autocomplete


LeoMcA:
Suggestion from @pepelsbey:

> I wonder if we could also support suggestions endpoint [in the Firefox add search engine UI] (Suggestions URL under Advanced option). For example, Ecosia’s URL is https://ac.ecosia.org/autocomplete?q=%s&type=list

This should be relatively trivial to do in our cloud function, repurposing the quicksearch logic from fred. Return format is described here: https://github.com/dewitt/opensearch/blob/master/mediawiki/Specifications/OpenSearch/Extensions/Suggestions/1.1/Draft%201.wiki#user-content-JSONformatted_search_suggestion_responses

We should also add the suggestions endpoint to our `opensearch.xml`, described here: https://github.com/dewitt/opensearch/blob/master/opensearch-1-1-draft-6.md#the-url-element

While we're at it, we should check if there's anything else to add to our `opensearch.xml`: at least a self-reference for auto-updating looks like it would be useful.

s-sood:
Hi @LeoMcA I can see that there are two open pull requests. Is there anything else needed to complete this work?

LeoMcA:
@s-sood PRs are merged and this is deployed to prod!

PR #1664, opened 2026-06-19
feat: point opensearch.xml at cloud function

LeoMcA:
Part of https://github.com/mdn/fred/issues/1079

https://github.com/mdn/dex/pull/391 must land first

Wires up the fred side of the quicksearch in OpenSearch:

- removes fred's static opensearch.xml to point at the locale-dependent cloud function one
- fixes a type issue in the quicksearch code
- fixes the first (0-indexed) item of the quicksearch never being returned
- routes the paths added in the above PR to the cloud function in the dev server, necessary for local testing


globau (approved):
lgtm

PR #1689, opened 2026-07-03
fix(a11y-menu): add scroll margin to main content skip link target

knowler:
### Description

This adds some scroll margin to the “Skip to main content” target globally (i.e. `#content`), so that the target’s content does not scroll beneath the sticky header.

### Motivation

Relevant visible content is obscured when following the skip link and what’s visible is still important for keyboard users and even screen reader users.

<details><summary>Before video</summary>

Description: in Firefox a user navigates to the MDN homepage and tabs to the “Skip to main content” link, upon activation the hero content scrolls beneath the sticky header causing some of the target content to be obscured. 

https://github.com/user-attachments/assets/88652a30-c62b-4dee-b3f1-5290bd091304 

</details>

<details><summary>After video</summary>

Description: in Firefox a user navigates to the MDN homepage and tabs to the “Skip to main content” link, upon activation the page scrolls slightly so that the ad is out of view, but the target content is not obscured at all.

https://github.com/user-attachments/assets/ce9e5d45-625d-4776-bd87-fc7365002baf

</details>

### Additional details

~~As far as I can tell, this isn’t a problem on other pages.~~ This is now applied globally.

Looked, but couldn’t find any.

### DIFF
--- a/components/global/global.css
+++ b/components/global/global.css
@@ -30,3 +30,7 @@ a {
 [hidden] {
   display: none !important;
 }
+
+#content {
+  scroll-margin-block-start: var(--sticky-header-height);
+}


knowler:
Some other affected routes:

- /search
- /about
- /advertising
- /community
- /blog
  - Also individual blog posts.
- /play
- /observatory

There’s probably a better way to handle all of these rather than individual changes like this one.

components/homepage/server.css
```diff
@@ -1,6 +1,7 @@
 .homepage {
   display: grid;
   grid-template-columns: var(--layout-no-sidebar-extended);
+  scroll-margin-block-start: var(--sticky-header-height);
```
caugner, line 4: You could add this inside `#content { … }` directly to `components/global/global.css` instead, assuming it has no negative side-effects on content pages.

caugner (dismissed):

components/homepage/server.css
knowler: Resolved here: https://github.com/mdn/fred/compare/3beeff0..f4292a71277588c79567132f8c0b29f9f62e57d7

LeoMcA (approved):
Thank you!

Issue #1680, opened 2026-06-29
Review npx usages


caugner:
We should ensure that all `npx` usages are unambiguous:

1. Either they refer to a package in our dependencies.
2. Or they refer to a binary provided by a package in our dependencies, and specify that dependency via the `--package` option.

PR #1697, opened 2026-07-07
feat(cli): add single `fred` entrypoint with subcommands

caugner:
### Description

Add a single `fred` binary that dispatches `server`/`ssr` subcommands, so `@mdn/fred` can be run as `npx @mdn/fred ssr`. The existing `fred-server`/`fred-ssr` binaries are kept as deprecated shims (which print a warning and delegate) for backward compatibility, and the publish-simulation workflow now uses `npx @mdn/fred ssr`.

- Add `bin/fred.js` dispatcher (`fred <server|ssr>`).
- Convert `fred-server`/`fred-ssr` bins to deprecation shims in `bin/`.
- Point `bin` entries in `package.json`/`package-lock.json` at the new wrappers.
- Drop the now-dead shebangs from `scripts/server.js` and `build/ssr.js` (no longer `bin` entries).
- Switch the publish-simulation workflow to `npx @mdn/fred ssr`.

### Motivation

Prevent the risk of bare `npx fred-server`/`npx fred-ssr` resolving to squatted registry packages when `@mdn/fred` is not installed locally, by steering consumers to the unambiguous `npx @mdn/fred <command>` (and `npx --package=@mdn/fred fred-ssr` during the deprecation window).

### Additional details

Both target invocations are verified to work: `npx @mdn/fred ssr` produces byte-identical output to running `build/ssr.js` directly, and `npx --package=@mdn/fred fred-ssr` continues to work via the shim. The two standalone bins remain only as a temporary bridge and can be dropped in a future major.

Part of https://github.com/mdn/fred/issues/1680.

### DIFF
--- a/.github/workflows/npm-publish-simulation.yml
+++ b/.github/workflows/npm-publish-simulation.yml
@@ -101,7 +101,7 @@ jobs:
       - name: (mdn/content) Render (Linux)
         if: runner.os == 'Linux'
         working-directory: mdn/content
-        run: npx fred-ssr
+        run: npx @mdn/fred ssr
 
       - name: (mdn/fred) Checkout
         uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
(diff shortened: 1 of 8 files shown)


LeoMcA (approved):
Nice, lgtm

PR #1698, opened 2026-07-07
style: resolve all ESLint warnings

caugner:
### Description

Fix all 31 ESLint warnings (`jsdoc/reject-any-type`, `jsdoc/reject-function-type`, `jsdoc/ts-no-empty-object-type`).

- Replace `Function` with concrete callback signatures (`() => void`, `(event: Event) => void`).
- Replace genuine `any` with precise types or `unknown` (`[T?, ...unknown[]]`, `PropertyValues<this>`, …).
- Drop a redundant `{}` union member in `stateToSession` (all callers pass `any`).
- Type play-console console-emulation args as `unknown`, casting at the `%d`/`%s`/`%f` substitution sites.
- Add scoped `eslint-disable` comments only where `any` is genuinely required: TS mixin constructors, abstract `render`/`renderSimplified` (overridden by subclasses yet consumed via generic `ReturnType<…>`), and `formatObject`'s dynamic member access.

### Motivation

Ensure `npx eslint .` passes with zero warnings, improving type precision across JSDoc annotations.

### Additional details

Both `npx eslint .` (0 warnings) and `npm run tsc` (no regressions) pass.

### DIFF
--- a/components/color-theme/controller.js
+++ b/components/color-theme/controller.js
@@ -31,7 +31,7 @@ export class ThemeController {
   }
 
   /**
-   * @param {any} value
+   * @param {unknown} value
    * @returns {"light" | "dark" | undefined}
    */
   _lightOrDark(value) {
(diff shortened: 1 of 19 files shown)


LeoMcA (approved):
Nice!

PR #1705, opened 2026-07-09
fix(playground): handle malformed HTML + warn when JS cannot run

### LINKED ISSUE #1440: When `<o` characters in Playground HTML input causes entire JavaScript input to show up in HTML preview

1. Using the characters "<o" in the Playground's HTML input area, causes whatever is in the JavaScript input area to show up in the HTML preview. 

2. Behavior persists so long as the tag (intended to use `<output>` when I discovered this by accident) remains incomplete even if there are inputs that precede or succeed it.

3. Can't tell if this is the result of a mishandling of some string interpolation, some quirk with the playground or if its an inherent problem with the HTML tag itself

4. Not too sure if its isolated only to certain devices but am experiencing it on a Pixel 10 Pro Fold and a 2020 13" MacBook Pro. 

5. Also not sure if this is tantamount to using `.innerText` which might not be so worrying than of its an equivalent of `innerHTML`, for e.g., hence this issue...

6. And a [live example](https://developer.mozilla.org/en-US/play?id=QFLgcYsFoQCovKNJo8Xv4J6fuHPk1hBMMbqMCqzsYtI1t%2F0hkoDtirkPCw49ILJKVyH9N6mKZA4FnOB5) you can use as a reference if it occurs at all, 

Image
Image

otherwise please refer to the screenshots attached for more details.

Nothing. Incomplete HTML tags either show up as they are typed or as nothing at all and nor expose another element's content.

As provided:

1. [live example](https://developer.mozilla.org/en-US/play?id=QFLgcYsFoQCovKNJo8Xv4J6fuHPk1hBMMbqMCqzsYtI1t%2F0hkoDtirkPCw49ILJKVyH9N6mKZA4FnOB5)
2.  Image
3. Image

_No response_

caugner:
(⚠️ Mirrored to https://github.com/mdn/dex/pull/435. ⚠️ )

### Description

Fix the Playground runner (`vendor/yari/libs/play/index.js`) so malformed HTML no longer swallows the user's JS, and warn when it still does:

- Insert `<!-- "" '' -->` after the HTML, which closes an unclosed tag, attribute value, or comment before the runner script.
- Add a sacrificial `<script></script>` so an unterminated user script does not swallow the start flag and trigger a false warning.
- Set `window.__mdnPlayJsStarted` in a script before the runner and `window.__mdnPlayJsEnded` in the script after it.
- Add a `<head>` check on `DOMContentLoaded` that warns via `console.warn` if either flag is missing.
- Add a unit test (`test/unit/play/runner.test.js`).

### Motivation

Ensure a typo like `<o` in the HTML no longer renders the JS as text, and that users get a hint in the remaining cases (e.g. an unclosed `<textarea>`), where the JS silently does not run.

### Additional details

Flags instead of marker elements: user JS that replaces the body's children (e.g. `document.body.innerHTML = ...`, as several live samples do) removes the script elements but not the flags.

Verified with headless Chrome:

| HTML | JS runs | Warning |
| --- | --- | --- |
| `<p>ok</p>` | yes | no |
| `annn <o jwj`, `<div class="`, unclosed `<!--` | yes | no |
| JS replacing `document.body` children | yes | no |
| unclosed `<textarea>`, `<template>`, `<math>` | no | yes |

The file is mirrored in `mdn/dex` (`cloud-function/src/internal/play/index.js`), which serves the runner in deployed environments, so the fix only takes effect there once mdn/dex#435 lands.

Fixes #1440.

### DIFF
--- a/vendor/yari/libs/play/index.js
+++ b/vendor/yari/libs/play/index.js
@@ -86,6 +86,39 @@ function html(strings, ...args) {
 }
 
 /**
+ * Prevent user CSS from breaking out of the `<style>` element via a literal
+ * `</style` sequence. Inserting a backslash (`<\/style`) stops the HTML parser
+ * from ending the element while remaining equivalent inside CSS strings, where
+ * `\/` resolves to `/`.
+ * @param {string} css
+ */
+function escapeStyleText(css) {
+  return css.replace(/<\/(style)/gi, String.raw`<\/$1`);
+}
+
+/**
+ * Prevent user JS from breaking out of the `<script>` element via a literal
+ * `</script` sequence. Inserting a backslash (`<\/script`) stops the HTML
+ * parser from ending the element while remaining equivalent inside JS strings,
+ * template literals, and regexes, where `\/` resolves to `/`. The equivalence
+ * holds within literal *contents*; the pathological `x</script/` — a `<`
+ * operator immediately followed by a regex literal beginning with `script`
+ * (i.e. `x < /script/`) — would not round-trip, but such code is vanishingly
+ * rare.
+ *
+ * Known limitation: this does not neutralize a `<!--` … `<script` sequence,
+ * which pushes the parser into the script-data-double-escaped state where the
+ * runner's own closing `</script>` no longer closes the element (swallowing
+ * everything after it, including the swallow-detection warning). No backslash
+ * escape fixes this without changing behavior — `<\!--` is an invalid escape in
+ * `/u` regexes — so such input still corrupts the runner.
+ * @param {string} js
+ */
+function escapeScriptText(js) {
+  return js.replace(/<\/(script)/gi, String.raw`<\/$1`);
+}
+
+/**
  * @param {State | null} state
  * @param {string} hrefWithCode
  * @param {string} searchWithState
@@ -428,7 +461,7 @@ export function renderHtml(state = null) {
             : ""
         }
         <style id="css-output">
-          ${css}
+          ${escapeStyleText(css)}
         </style>
         <script>
           const consoleProxy = new Proxy(console, {
@@ -527,7 +560,7 @@ export function renderHtml(state = null) {
       <body>
         ${htmlCode}
         <script id="mdn-play-js" type=${defaults === "ix-wat" ? "module" : ""}>
-          ${js};
+          ${escapeScriptText(js)};
         </script>
         <script>
           try {
--- a/vendor/yari/libs/play/index.js
+++ b/vendor/yari/libs/play/index.js
@@ -526,11 +526,26 @@ export function renderHtml(state = null) {
       </head>
       <body>
         ${htmlCode}
-        <script type=${defaults === "ix-wat" ? "module" : ""}>
+        <script id="mdn-play-js" type=${defaults === "ix-wat" ? "module" : ""}>
           ${js};
         </script>
         <script>
           try {
+            // Post-rendering check: an unclosed or malformed tag in the HTML
+            // input can swallow the following <script> element (its markup is
+            // parsed as attributes of the open tag), leaving the JavaScript
+            // source visible as text content. Detect that and warn the user.
+            // Match the <script> tag with our id specifically, so a user
+            // element that happens to reuse the id cannot trigger a false
+            // warning when the real script did run.
+            const jsScript = document.querySelector("script#mdn-play-js");
+            if (!(jsScript instanceof HTMLScriptElement)) {
+              console.warn(
+                "[Playground] The JavaScript did not run because the HTML input " +
+                  'contains an unclosed or malformed tag (for example "<o"). ' +
+                  "Close the tag to run your code.",
+              );
+            }
             const uuid =
               new URLSearchParams(location.search).get("uuid") || undefined;
             window.parent.postMessage({ uuid, typ: "ready" }, "*");
--- a/vendor/yari/libs/play/index.js
+++ b/vendor/yari/libs/play/index.js
@@ -488,6 +488,19 @@ export function renderHtml(state = null) {
           window.console = consoleProxy;
           window.addEventListener("error", (e) => console.log(e.error));
         </script>
+        <script>
+          document.addEventListener("DOMContentLoaded", () => {
+            const start = document.querySelector("script#mdn-play-js");
+            const end = document.querySelector("script#mdn-play-js-end");
+            if (!(start instanceof HTMLScriptElement) || !end) {
+              console.warn(
+                "[Playground] Could not verify that the JavaScript ran. This " +
+                  "usually means the HTML input contains an unclosed or " +
+                  "malformed tag.",
+              );
+            }
+          });
+        </script>
         ${
           defaults === "ix-tabbed"
             ? html`<script>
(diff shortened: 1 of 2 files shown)


vendor/yari/libs/play/index.js
caugner, line 1: The changes would need to be replicated in Dex.

caugner: https://github.com/mdn/dex/pull/435

vendor/yari/libs/play/index.js
```diff
@@ -85,6 +85,39 @@ function html(strings, ...args) {
...
+ *
+ * Known limitation: this does not neutralize a `<!--` … `<script` sequence,
+ * which pushes the parser into the script-data-double-escaped state where the
+ * runner's own closing `</script>` no longer closes the element (swallowing
+ * everything after it, including the swallow-detection warning). No backslash
+ * escape fixes this without changing behavior — `<\!--` is an invalid escape in
+ * `/u` regexes — so such input still corrupts the runner.
+ * @param {string} js
+ */
+function escapeScriptText(js) {
```
LeoMcA: Let's not attempt to escape things - I'm not sure how often a user is seriously going to hit a case where they're putting `</script` in their own script, and the comment here justifying it reads like an essay - let's keep this implementation simple.

vendor/yari/libs/play/index.js
```diff
@@ -85,6 +85,39 @@ function html(strings, ...args) {
...
 }
 
+/**
+ * Prevent user CSS from breaking out of the `<style>` element via a literal
+ * `</style` sequence. Inserting a backslash (`<\/style`) stops the HTML parser
+ * from ending the element while remaining equivalent inside CSS strings, where
+ * `\/` resolves to `/`.
+ * @param {string} css
+ */
+function escapeStyleText(css) {
```
LeoMcA: Same wrt not escaping.

vendor/yari/libs/play/index.js
```diff
@@ -526,11 +559,26 @@ export function renderHtml(state = null) {
...
         </script>
         <script>
           try {
+            // Post-rendering check: an unclosed or malformed tag in the HTML
+            // input can swallow the following <script> element (its markup is
+            // parsed as attributes of the open tag), leaving the JavaScript
+            // source visible as text content. Detect that and warn the user.
+            // Match the <script> tag with our id specifically, so a user
+            // element that happens to reuse the id cannot trigger a false
+            // warning when the real script did run.
```
LeoMcA:
```suggestion
```

vendor/yari/libs/play/index.js
```diff
@@ -526,11 +559,26 @@ export function renderHtml(state = null) {
...
+            const jsScript = document.querySelector("script#mdn-play-js");
+            if (!(jsScript instanceof HTMLScriptElement)) {
+              console.warn(
+                "[Playground] The JavaScript did not run because the HTML input " +
+                  'contains an unclosed or malformed tag (for example "<o"). ' +
```
LeoMcA: Let's drop the example, and are we sure that it only fires in this case?

vendor/yari/libs/play/index.js
```diff
@@ -526,11 +559,26 @@ export function renderHtml(state = null) {
       </head>
       <body>
         ${htmlCode}
-        <script type=${defaults === "ix-wat" ? "module" : ""}>
-          ${js};
+        <script id="mdn-play-js" type=${defaults === "ix-wat" ? "module" : ""}>
+          ${escapeScriptText(js)};
         </script>
         <script>
```
LeoMcA: We can cover both cases: the script tag never exists, or the script tag ends early, by moving this check to the `<head>`, running after `DOMContentLoaded`, and adding a check to ensure that the node following `mdn-play-js` is the one we expect

LeoMcA (changes requested):

vendor/yari/libs/play/index.js
caugner: Dropped the example, and softened the message: you're right that it doesn't only fire in that case. The check can also fire when the JS *did* run but removed the script element itself, e.g. `document.body.innerHTML = "..."` or `document.write()`. It now says "Could not verify that the JavaScript ran. This usually means the HTML input contains an unclosed or malformed tag." rather than asserting a cause.

vendor/yari/libs/play/index.js
caugner: Done, both cases are covered now. The check moved into `<head>` and runs on `DOMContentLoaded`, and the trailing ready-message script carries `id="mdn-play-js-end"` as the end marker.

One detail: `nextElementSibling` alone is not enough for the early-close case. With `<script id="mdn-play-js">foo</script>bar;</script><script id="mdn-play-js-end">`, the leaked `bar;` is a text node and the stray `</script>` is ignored, so the end marker is still the next *element* sibling. The check therefore walks `nextSibling` and flags any non-whitespace text or element between the two.

Moving to `<head>` also fixes cases the body-side check could not see at all, because it was swallowed along with everything else: an unterminated attribute value (`<p class="ok>`) or an unclosed `<!--`.

vendor/yari/libs/play/index.js
```diff
@@ -530,10 +553,11 @@ export function renderHtml(state = null) {
       </head>
       <body>
         ${htmlCode}
-        <script type=${defaults === "ix-wat" ? "module" : ""}>
+        <script id="mdn-play-js" type=${defaults === "ix-wat" ? "module" : ""}>
           ${js};
         </script>
-        <script>
+        <script id="mdn-play-js-end">
+          // The id doubles as the end marker for the swallow detection above.
```
caugner:
```suggestion
```

vendor/yari/libs/play/index.js
```diff
@@ -488,6 +484,33 @@ export function renderHtml(state = null) {
           window.console = consoleProxy;
           window.addEventListener("error", (e) => console.log(e.error));
         </script>
+        <script>
+          // Malformed HTML can swallow the JS <script> element, and a literal
```
caugner:
```suggestion
```

vendor/yari/libs/play/index.js
```diff
@@ -488,6 +488,19 @@ export function renderHtml(state = null) {
           window.console = consoleProxy;
           window.addEventListener("error", (e) => console.log(e.error));
         </script>
+        <script>
+          document.addEventListener("DOMContentLoaded", () => {
+            const start = document.querySelector("script#mdn-play-js");
```
caugner: Nit: `querySelector("script#mdn-play-js")` only ever returns a `<script>`, so in the swallowed case (`<o id="mdn-play-js" ...>`) it returns `null` and the `instanceof HTMLScriptElement` branch never trips. The two checks express the same intent twice. Suggest picking one:

```js
const start = document.getElementById("mdn-play-js");
const end = document.getElementById("mdn-play-js-end");
if (!(start instanceof HTMLScriptElement) || !end) {
```

This keeps the `instanceof` check meaningful (it is what actually distinguishes the swallowed element) and matches the PR description.

vendor/yari/libs/play/index.js
```diff
@@ -488,6 +488,19 @@ export function renderHtml(state = null) {
...
+            const end = document.querySelector("script#mdn-play-js-end");
+            if (!(start instanceof HTMLScriptElement) || !end) {
```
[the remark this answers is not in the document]
caugner: That is genuinely a good idea.

vendor/yari/libs/play/index.js
```diff
@@ -530,10 +540,15 @@ export function renderHtml(state = null) {
       </head>
       <body>
         ${htmlCode}
-        <script type=${defaults === "ix-wat" ? "module" : ""}>
+        <!-- "" '' -->
+        <script>
```
LeoMcA, line 545: Small problem here where we could end up with false positives with an unterminated `<script>` tag: it would swallow the `started = true` assignment, but not the actual script - so we'd report that the javascript hasn't run when it has.

Should be solvable with a sacrificial script:


```suggestion
        <script></script>
        <script>
```

LeoMcA (approved):

PR #1706, opened 2026-07-09
fix(search-modal): add ARIA combobox semantics to quick search

### LINKED ISSUE #1175: [a11y] MDN search autocomplete broken for screen reader users

This is a late follow-up to https://github.com/mdn/yari/issues/4407

The current implementation for autocomplete search on MDN is borked when using a screen reader.

* no announcement is made when suggestions appear
* when using cursor keys to move between search suggestions, the search term in the combobox (rather than the value of the suggestion) is announced

Demonstration video: using Chrome + NVDA. after entering "test", results appear but NVDA remains silent. going through the results, NVDA just announces "test" for each option rather than the visible text of the result/option.

https://github.com/user-attachments/assets/90473ad0-f55c-404e-b5bf-d36327a91e12

When search suggestions/results are shown, an announcement should be made that suggestions/results are available/shown.

When navigating with the cursor key up/down to go through the options, announce the actual text or that option, not the current value of the combobox input.

As an example of a good implementation, see how the github autocomplete here does it. Check the underlying markup, adopt the same approach.

Demonstration video: Chrome + NVDA. github's combobox annoucnes that results are available/shown. using cursor keys to go through the options, the actual option text is announced

https://github.com/user-attachments/assets/d887ba30-cd4d-40cf-bd21-595e44212671

For a basic implementation, see https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/

Note that the APG pattern doesn't do the "announce the number of results", so the github version is superior here...but it gives you a first bit of direction.

_No response_

caugner:
### Description

Add ARIA combobox semantics to the quick-search autocomplete (`search-modal`):

- The input gets `role="combobox"`, `aria-autocomplete="list"`, `aria-controls`, `aria-expanded`, and `aria-activedescendant`.
- The results list becomes a `listbox` with `option` children carrying ids and `aria-selected`.
- A visually-hidden `role="status"` live region announces the result count.
- Arrow-key navigation keeps focus on the input and moves the active option via `aria-activedescendant`; the active index resets on input, and `aria-activedescendant` is only set once the search task completes, so it can't reference an option that isn't currently rendered.
- Adds `search-modal-results-status` and `search-modal-results-label` strings, translated for `de`, `fr`, `ja`, `zh-CN`.

### Motivation

The quick-search autocomplete had no combobox semantics, so with a screen reader no announcement was made when results appeared, and arrow-key navigation re-read the typed query instead of the highlighted option. This intends to follow the [APG combobox-with-listbox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-autocomplete-list/) plus GitHub's result-count announcement.

### Additional details

- Tested most of it locally with VoiceOver on macOS.
- The other locales (`es`, `ko`, `pt-BR`, `ru`, `zh-TW`) don't translate `search-modal` and fall back to en-US, so they were left unchanged.

Fixes #1175.


components/search-modal/element.js
```diff
@@ -71,6 +71,9 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
   _input({ inputType, target }) {
     if (target instanceof HTMLInputElement) {
       this._query = target.value;
+      // Reset the active option so `aria-activedescendant` and the visual
+      // highlight can't point past the end of a shrunken result set.
```
LeoMcA:
```suggestion
```

components/search-modal/element.js
```diff
@@ -249,6 +252,20 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
     const searchUrl = this._query
       ? `/${this.locale}/search?${new URLSearchParams({ q: this._query })}`
       : null;
+    // Options are the quick-search results plus the trailing "site search"
+    // entry (present whenever there is a query).
+    const optionCount = siteSearchIndex + (searchUrl ? 1 : 0);
+    const hasOptions = optionCount > 0;
```
LeoMcA: We only use `optionCount` once, here, and `hasOptions` is merely checking if `this._query` is truthy - just use `this._query` in place of `hasOptions` below.


```suggestion
```

components/search-modal/element.js
```diff
@@ -249,6 +252,20 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
...
+    // Only announce a count once the search has actually completed, so we
+    // don't say "no results" while the query is still being looked up.
```
LeoMcA:
```suggestion
```

components/search-modal/element.js
```diff
@@ -270,8 +287,16 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
             @input=${this._input}
             placeholder=${this.l10n("search-modal-search")`Search`}
             aria-label=${this.l10n("search-modal-search")`Search`}
+            role="combobox"
+            aria-autocomplete="list"
+            aria-controls="search-modal-listbox"
+            aria-expanded=${hasOptions ? "true" : "false"}
+            aria-activedescendant=${
+              hasOptions ? `search-modal-result-${this._selected}` : nothing
```
LeoMcA:
```suggestion
            aria-expanded=${this._query ? "true" : "false"}
            aria-activedescendant=${
              this._query ? `search-modal-result-${this._selected}` : nothing
```

LeoMcA (changes requested):
We can simplify, and we don't need these comments describing things trivially readable from the code:

components/search-modal/element.js
```diff
@@ -270,8 +287,16 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
...
+            }
```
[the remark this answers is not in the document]
caugner: Good catch, fixed. `aria-activedescendant` now reuses `searchComplete`, so it's only set once the options it references are rendered.

components/search-modal/element.js
```diff
@@ -271,8 +280,18 @@ export class MDNSearchModal extends L10nMixin(LitElement) {
               @input=${this._input}
               placeholder=${this.l10n("search-modal-search")`Search`}
               aria-label=${this.l10n("search-modal-search")`Search`}
+              role="combobox"
+              aria-autocomplete="list"
+              aria-controls="search-modal-listbox"
+              aria-expanded=${this._query ? "true" : "false"}
+              aria-activedescendant=${
+                searchComplete
```
LeoMcA: Hmm, actually we do have an `activedescendant` before the search is complete: the "Site search for ..." option, should this be?

```suggestion
              aria-activedescendant=${
                this._query
```

LeoMcA (approved):

components/search-modal/element.js
caugner: Yeah, I think you're right, and it makes sense to be consistent with `aria-expanded` here.

PR #1710, opened 2026-07-11
chore(navigation-menu): style and text updates to condense menu

brysonbw:
### Changes

- Remove Learn/Tools icons
- Update `Web APIs` text/verbiage to `APIs`
- Applied updated search input styling for the `--screen-menu-compact` breakpoint
- Updated the @custom-media definition for `--screen-menu-compact`

### Before

https://github.com/user-attachments/assets/ffc04c34-42ac-4a8b-afbf-58e97dc57245

### After

https://github.com/user-attachments/assets/1e2dda2d-44e4-4b00-a875-cba17879e08a

## Related issues and pull requests

- #1436

### DIFF
--- a/components/button/server.css
+++ b/components/button/server.css
@@ -142,8 +142,8 @@
   }
 
   svg {
-    width: 1.25em;
-    height: 1.25em;
+    width: var(--icon-size, 1.25em);
+    height: var(--icon-size, 1.25em);
   }
 
   /* Label */
--- a/components/button/server.css
+++ b/components/button/server.css
@@ -142,8 +142,8 @@
   }
 
   svg {
-    width: var(--icon-size, 1.25em);
-    height: var(--icon-size, 1.25em);
+    width: 1.25em;
+    height: 1.25em;
   }
 
   /* Label */
--- a/components/menu/base.css
+++ b/components/menu/base.css
@@ -7,13 +7,11 @@
   --menu-tab-text: var(--color-text-blue);
 
   &[data-section="learn"] {
-    --menu-tab-icon: url("../icon/graduation-cap.svg");
     --menu-tab-background: var(--color-background-orange);
     --menu-tab-text: var(--color-text-orange);
   }
 
   &[data-section="tools"] {
-    --menu-tab-icon: url("../icon/wrench.svg");
     --menu-tab-background: var(--color-background-purple);
     --menu-tab-text: var(--color-text-purple);
   }
@@ -67,15 +65,6 @@
 
   &[aria-expanded="true"]::after {
     scale: -1;
-  }
-
-  /* Icon */
-
-  :is([data-section="learn"], [data-section="tools"]) & {
-    &::before {
-      content: "";
-      mask-image: var(--menu-tab-icon);
-    }
   }
 }
 
--- a/components/menu/base.css
+++ b/components/menu/base.css
@@ -7,11 +7,13 @@
   --menu-tab-text: var(--color-text-blue);
 
   &[data-section="learn"] {
+    --menu-tab-icon: url("../icon/graduation-cap.svg");
     --menu-tab-background: var(--color-background-orange);
     --menu-tab-text: var(--color-text-orange);
   }
 
   &[data-section="tools"] {
+    --menu-tab-icon: url("../icon/wrench.svg");
     --menu-tab-background: var(--color-background-purple);
     --menu-tab-text: var(--color-text-purple);
   }
@@ -65,6 +67,15 @@
 
   &[aria-expanded="true"]::after {
     scale: -1;
+  }
+
+  /* Icon */
+
+  :is([data-section="learn"], [data-section="tools"]) & {
+    &::before {
+      content: "";
+      mask-image: var(--menu-tab-icon);
+    }
   }
 }
 
@@ -186,5 +197,15 @@
 
   &[data-icon="github"]::before {
     mask-image: url("../icon/social/github.svg");
+  }
+}
+
+@media (--screen-small-and-narrower) {
+  .menu__tab-button {
+    :is([data-section="learn"], [data-section="tools"]) & {
+      &::before {
+        content: none;
+      }
+    }
   }
 }
--- a/components/menu/base.css
+++ b/components/menu/base.css
@@ -71,10 +71,12 @@
 
   /* Icon */
 
-  :is([data-section="learn"], [data-section="tools"]) & {
-    &::before {
-      content: "";
-      mask-image: var(--menu-tab-icon);
+  @media not (--screen-menu-compact) {
+    :is([data-section="learn"], [data-section="tools"]) & {
+      &::before {
+        content: "";
+        mask-image: var(--menu-tab-icon);
+      }
     }
   }
 }
@@ -197,15 +199,5 @@
 
   &[data-icon="github"]::before {
     mask-image: url("../icon/social/github.svg");
-  }
-}
-
-@media (--screen-small-and-narrower) {
-  .menu__tab-button {
-    :is([data-section="learn"], [data-section="tools"]) & {
-      &::before {
-        content: none;
-      }
-    }
   }
 }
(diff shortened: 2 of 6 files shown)


components/menu/tabs.json
```diff
@@ -220,7 +220,7 @@
   },
   {
     "id": "webapis",
-    "buttonText": "Web APIs",
+    "buttonText": "APIs",
```
caugner: How about handling this similar to the JavaScript/JS menu item? (cc @mdn/content-team)

```suggestion
    "buttonText": {
      "long": "Web APIs",
      "short": "APIs"
    },
```

components/search-button/element.css
```diff
@@ -1,34 +1,24 @@
...
-  cursor: pointer;
-
-  background-color: var(--color-background-page);
-  border: 1px solid var(--color-border-primary);
-  border-radius: var(--radius-full);
+  --icon-size: 1.5em;
 
   &:hover {
-    background-color: var(--color-background-secondary);
+    opacity: 0.7;
```
caugner: I would have hoped that the hover effect of the plain button is sufficient, as we don't seem to do have this hover opacity effect elsewhere, but maybe it isn't / we do?

caugner: I would expect the search button to behave on hover like the login button (`.user-menu__login`) next to it.

pepelsbey:
> Remove Learn/Tools icons

The issue suggested _hiding_ icons at narrow viewports, not removing them.

I fully understand that the linked issue is pretty brief, but there's something that's been misunderstood in this PR: the changes proposed in the issue are adaptive (depending on the available space), not static (for all viewports).

It means, for example, that icons are there on desktop and on mobile viewports, but are hidden on narrow viewports, just before mobile. The same for Web APIs turning to APIs, etc.

brysonbw:
> > Remove Learn/Tools icons
> 
> The issue suggested _hiding_ icons at narrow viewports, not removing them.
> 
> I fully understand that the linked issue is pretty brief, but there's something that's been misunderstood in this PR: the changes proposed in the issue are adaptive (depending on the available space), not static (for all viewports).
> 
> It means, for example, that icons are there on desktop and on mobile viewports, but are hidden on narrow viewports, just before mobile. The same for Web APIs turning to APIs, etc.

Thanks for clarifying. Would using the `--screen-small-and-narrower` CSS variable in a media query target the viewport mentioned or `--screen-medium-and-narrower variable` or `--screen-menu-compact` @caugner?

components/search-button/element.css
brysonbw: I personally went with `opacity` instead of matching user menu login icon/button but that works too

pepelsbey: I would suggest not mixing colors via opacity and following existing hover patterns: see other menus and dropdowns.

components/menu/base.css
```diff
@@ -199,3 +199,13 @@
     mask-image: url("../icon/social/github.svg");
   }
 }
+
+@media (--screen-small-and-narrower) {
```
LeoMcA: Let's move this to the block in L74 so we can keep the icon logic together:

https://github.com/mdn/fred/blob/8d3853d3041b9d77e2a2eff05cf3e735530a8567/components/menu/base.css#L74-L79

That logic shouldn't apply between `--screen-menu-compact` and `--screen-menu-hamburger`. I think the most elegant solution would be to tighten up `--screen-menu-compact`'s definition to be `@custom-media --screen-menu-compact (1044px < width <= 1104px);`, and then wrap the above block with `@media not (--screen-menu-compact)`. If that doesn't work for whatever reason, we can just do `content: none;` within `@media (--screen-menu-compact)`

components/search-button/element.css
```diff
@@ -1,34 +1,16 @@
...
   display: flex;
-
-  align-items: center;
-  justify-content: space-between;
-
-  width: 5rem;
-
-  padding: 0.25rem;
-  padding-left: 0.75rem;
+  padding: 0;
```
LeoMcA: I think we do want to keep the current search button styling outside of the `--screen-menu-compact` media query: so can we move the style changes into a `@media (--screen-menu-compact)` block to remove the border, background, etc. (or, vice-versa if it makes more sense).

components/button/server.css
```diff
@@ -142,8 +142,8 @@
   }
 
   svg {
-    width: 1.25em;
-    height: 1.25em;
+    width: var(--icon-size, 1.25em);
+    height: var(--icon-size, 1.25em);
```
LeoMcA: If we keep the "fake input" styling, we can revert this change.

components/search-button/element.js
```diff
@@ -19,14 +19,16 @@ export class MDNSearchButton extends L10nMixin(LitElement) {
   }
 
   render() {
-    return html`<button
+    return html`<mdn-button
```
LeoMcA: If we keep the "fake input" styling, we can revert the changes in this file too.

LeoMcA (changes requested):
> Would using the --screen-small-and-narrower CSS variable in a media query target the viewport mentioned or --screen-medium-and-narrower variable or --screen-menu-compact

@brysonbw Almost, for the menu we use a specific set of breakpoints, independent of the general breakpoints: the main breakpoint is at `1044px`: wider than that is the "desktop menu" (`--screen-menu-full`), narrower than that is the "mobile menu" (`--screen-menu-hamburger`).

Within the "desktop menu" we have a compact mode (between `1044px` and `1104px`, though the `--screen-menu-compact` currently covers all of "less than `1104px`"). It's this compact mode we want to hide icons etc. on

components/button/server.css
brysonbw: I think making it a mdn-button would be best. Yet, all smooth

components/search-button/element.css
```diff
@@ -31,4 +31,17 @@
     border-bottom: 2px solid
       light-dark(var(--color-blue-50), var(--color-blue-80));
   }
+
+  @media (--screen-menu-compact) {
```
LeoMcA, line 35: Hmm, I think your original instincts to turn this into a `mdn-button` were right: we're missing a few things here relative to the adjacent login/out button (the background hover colour, the icon centering, etc.).

I reckon it would be safer to start with the default mdn-button styling, then build the fake input styling on top of that with a `@media not (--screen-menu-compact)` query.

LeoMcA (changes requested):
Good stuff! You were right to convert to `mdn-button` - apologies for leading your astray - let's make that change again, so styles are pretty default for the search button within the compact range, and then we build the mocked input styles on top.

Also, now that we've condensed the menu a bunch, let's update that 1044px breakpoint to be smaller - get it as small as you can without a horizontal scrollbar appearing - there should be an e2e test which catches that happening.

brysonbw:
> Good stuff! You were right to convert to `mdn-button` - apologies for leading your astray - let's make that change again, so styles are pretty default for the search button within the compact range, and then we build the mocked input styles on top.
> 
> Also, now that we've condensed the menu a bunch, let's update that 1044px breakpoint to be smaller - get it as small as you can without a horizontal scrollbar appearing - there should be an e2e test which catches that happening.

All good brother. Thanks for the feedback

components/search-button/element.css
brysonbw: Is this something we are looking to achieve @LeoMcA?

> Here in video:  `@custom-media --screen-menu-compact (960px < width <= 1104px);`

https://github.com/user-attachments/assets/12230c54-e55e-47ed-947a-a7803bfc23e3

LeoMcA: @brysonbw on the right track, but can we style the "fake input" again over the top of the `mdn-button` for the full width and mobile menus (i.e. where your video starts and ends), so resizing wide to narrow we go: "fake input", normal button, "fake input" again?

I think we can go more aggressive with that breakpoint too: the buttons and menu should nearly touch before we collapse to the compact menu, and then again before we collapse to the hamburger

PR #1713, opened 2026-07-13
chore(placement): instrument pong fetch requests and errors

caugner:
### Description

Instrument the placement (pong) fetch in `components/placement/context.js`:

- Emit `pong: pong->requested` immediately before each `/pong/get` request.
- Wrap the request in `try/catch`, emitting `pong: pong->error <name>` when it throws, and re-throw so the failure is no longer silently swallowed ahead of the `response.ok` check.

### Motivation

Make it easier to explain the mismatch between page views and the (lower) number of pong fetch requests. Today only `pong->fetched <status>` is recorded, which fires *after* a response arrives — so any drop-off before that point is invisible. Adding `pong->requested` (attempts) and `pong->error` (network/other failures) lets us line up page views → requested → fetched and see exactly where the count falls off.

### Additional details

Event funnel after this change:

- `pong->requested` — a fetch was attempted
- `pong->error <name>` — the fetch threw (e.g. `TypeError` for a network error)
- `pong->fetched <status>` — a response was received

### DIFF
--- a/components/placement/context.js
+++ b/components/placement/context.js
@@ -78,20 +78,28 @@ export function globalPlacementContext() {
  * @returns {Promise<Placements.PlacementContextData>}
  */
 async function fetchPlacementData() {
-  const response = await fetch("/pong/get", {
-    method: "POST",
-    headers: {
-      "Content-Type": "application/json",
-    },
-    body: JSON.stringify({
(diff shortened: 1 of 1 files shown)


LeoMcA (approved):
lgtm!

PR #1715, opened 2026-07-14
fix(about-tabs): add keyboard navigation to About page tabs

### LINKED ISSUE #745: About page tabs cannot be switched with keyboard
STR:

1. Go to https://developer.mozilla.org/en-US/about
2. Press <kbd>Tab</kbd> until you focus the first tab
3. Try to change the tab

Expected: to be able to change the tab

Actual: the tab cannot be changed

Note: Observatory also has tabs ([example](https://developer.mozilla.org/en-US/observatory/analyze?host=mdn.dev#scoring)), but keyboard navigation works there.

KangaZero:
### Description

Adds keyboard navigation to the About page tabs (`mdn-about-tabs`). Arrow Left/Right move between tabs (with wrap-around) and activate them. The tablist now exposes `role="tablist"` + `aria-orientation="horizontal"`, and roving `tabindex` keeps it a single Tab stop so Tab moves focus in and out of the widget.

### Motivation

Fixes #745. Previously the tabs only handled clicks, so keyboard users could focus a tab but had no way to switch between them (WCAG 2.1.1 keyboard operability). The interaction mirrors the existing `observatory-results` tablist, which the issue references as the working example — arrow keys switch and activate (automatic activation, matching its native radio inputs).

### Additional details

- Left/Right only, consistent with `aria-orientation="horizontal"`.
- Single-file change; verified locally with `prettier`, `eslint`, `lit-analyzer`, and `tsc` (all clean).

Fixes #745


KangaZero:
Addressed in `15b78f4` — unified the click and keyboard paths through `_activateTab`, so keyboard activation now performs the same deferred `scrollIntoView` when the target panel sits above the viewport (and removes the duplicated activation logic).

LeoMcA (approved):
Thank you! Looks good

PR #1743, opened 2026-07-24
fix(play): encode # in data urls

### LINKED ISSUE #1734: MDN Playground should URL-escape `#` characters when generating data URIs (as `%23`) since a raw `#` terminates a URL

STR:
1. Visit https://developer.mozilla.org/en-US/play
2. Paste this in to the HTML section: `<div style="background: #00ff00">Hello</div>`
3. Click "Share" and then "Copy data URL to clipboard"
4. Paste into URL bar in Chrome or Firefox (or probably Safari, haven't tested but likely same outcome)

EXPECTED RESULTS:
Working data URL (with a url-encoded # character)

ACTUAL RESULTS:
The data URL renders as a a blank document (because MDN fails to URL-encode the # character in the css color expresssion #00ff00 , and so the # character terminates the data URL and everything after it is treated as an anchor reference; where "anchor reference" is e.g. the `#general_support` thing at the end of `https://developer.mozilla.org/en-US/docs/MDN/Community#general_support`  for example).

The URL encoding for # is %23 , so that's what should be included in the data URI and that produces a working result.

See EXPECTED RESULTS above.

_No response_

_No response_

LeoMcA:
Fixes https://github.com/mdn/fred/issues/1734


components/playground/element.js
```diff
@@ -140,8 +140,10 @@ ${"```"}`,
       if (html) code += html;
       if (js) code += `<script>${js}</script>`;
       code += `</body>`;
-      // encode non-space whitespace and percent
-      code = code.replaceAll(/([^\S ]|%)+/g, (str) => encodeURIComponent(str));
+      // encode characters which break urls
+      code = code.replaceAll(/([^\S ]|%|#)+/g, (str) =>
```
globau: This would be cleaner as `/(?:[^\S ]|[%#])+/g`.

However, this selective char approach seems fragile (what about other potentially URL breaking chars like `& ? = % + / ;`?).  Why not encode the whole string with `encodeURIComponent`?

globau (changes requested):

globau (approved):

PR #1755, opened 2026-08-01
chore(content-section): remove no-op @ts-nocheck

KangaZero:
## Description

removes a `// @ts-nocheck` inside `Prose()` in `components/content-section/server.js`. its sitting mid function (between two statements). Typescript's `@ts-nocheck` only applies when placed on the first line of a file, so it currently suppresses nothing and treated as a plain comment.


LeoMcA (approved):
Thank you!

PR #1783, opened 2026-08-10
ci(tests): only run 1 worker in windows

LeoMcA:


LeoMcA:
Doesn't prevent failures, so I'll go with https://github.com/mdn/fred/pull/1789 instead.

PR #1825, opened 2026-08-21
fix(button): make disabled state more obvious

### LINKED ISSUE #1469: Accessibility improvement of playground & demo "Reset" button style

The "_Reset_" button for demo examples is disabled until the user alters a block of code. When I first viewed the button with its `:disabled` style, I first believed that the button was not disabled. It would be beneficial to add a hover tooltip that indicates such, a less opaque `:disabled` style, and a `cursor` value will avoid confusion as to why the button doesn't respond to user interaction.

I feel that adding `opacity: 0.5` will work well for this, along with a hover tooltip which, when the disabled button is hovered over, displays a message along the lines of "_You may reset the lines of demo code after you alter them._" A `cursor` value of `not-allowed` may provide another visual aid in this regard.

I expected to see a less opaque `:disabled` style, an explanatory tooltip, and a visual cursor aid.

See the "_Reset_" button in any "_Try it_" demo. Here is one example:
https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/animation-direction#try_it

_No response_

LeoMcA:
Fixes https://github.com/mdn/fred/issues/1469

We can't reduce contrast without making the button inaccessible, but I have adjusted the icon to show a prohibited cursor, and added a reason which appears in the title and a11y tree.

Been working on a proper tooltip component which will work on mobile as well as some Friday fun, but it's not quite ready yet.

### DIFF
--- a/components/button/element.js
+++ b/components/button/element.js
@@ -46,7 +46,7 @@ export class MDNButton extends LitElement {
    * @param {MouseEvent} event
    */
   #click(event) {
-    if (this.disabled && this.disabledReason) {
+    if (!this.href && this.disabled && this.disabledReason) {
       event.preventDefault();
       event.stopImmediatePropagation();
     }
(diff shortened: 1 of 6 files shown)


globau (changes requested):
`aria-disabled` needs to be added

LeoMcA:
The concept of a disabled link doesn't really make sense to me: a link which goes somewhere, but doesn't? It would also require more work to implement properly, so instead I've just excluded it to maintain the existing behaviour.

globau (approved):

Issue #1507, opened 2026-04-22
Playground should not be able to generate URLs that exceed the maximum length


globau:
Via https://github.com/mdn/content/issues/43892

URLs have a maximum length of 8k - it's possible using the playground to generate longer URLs.

We should, at least:
- add validation in the Playground directly that prevents generation of URLs exceeding 8000 encoded characters
- add a Content built-time test to as a backstop

LeoMcA:
Started to explore a couple of options as fallbacks when the URL does exceed 8k:

- https://github.com/mdn/fred/pull/1513
- https://github.com/mdn/fred/pull/1512

Unfortunately the PR preview doesn't build any pages which have these long examples, so testing needs to be done locally:

1. http://localhost:3000/en-US/docs/Web/Media/Guides/Audio_and_video_delivery/Video_player_styling_basics#result
2. http://localhost:3000/en-US/docs/Web/Media/Guides/Audio_and_video_delivery/Adding_captions_and_subtitles_to_HTML5_video#result
3. http://localhost:3000/en-US/docs/Web/CSS/Guides/Colors/Color_mixer
4. http://localhost:3000/en-US/docs/Web/API/CanvasRenderingContext2D/arcTo#constructing_an_arcto_path

Here's the sizes of each in bytes currently, and with the two above approaches:

| example | current | minify | postMessage |
| ------- | ------- | ------ | ----------- |
| 1       | 9807    | 9303   | 497         |
| 2       | 11034   | 10186  | 514         |
| 3       | 8013    | 7093   | 461         |
| 4       | 10162   | 4674   | 466         |

## minify approach

This uses `terser` with default options to minify the JavaScript in each example. We could also use similar libraries to do the same to the HTML and CSS - though nothing will really get around the large embedded PNGs in the first two examples.

Unfortunately, this adds quite a bit to bundle size, doubling the size of JS over the wire (146kB to 298kB compressed; 370kB to 880kB uncompressed), though this is loaded async, so only on pages which require it.

We could roll our own regex to strip whitespace, comments etc. but I'd be concerned at our ability to handle every edge case (e.g. whitespace within strings), and isn't really something I think we should maintain.

The advantage of a minimisation approach is it already works with the other things you can do from an example, namely:
- click the Play button and have it appear in the Playground
- right click within the iframe, and open that in a new tab

However, users will see the *minified* source within the Playground/new tab warning screen, which isn't the nicest. We do have a format button, but this won't handle minification which renames variables, for example.

[...]

PR #1847, opened 2026-09-03
build(sandbox): add per-component preview routes

LeoMcA:
Relates to https://github.com/mdn/fred/issues/1507 - I wanted to write some regression e2e tests before opening the PR to fix that, and this turned into a bit of a yak shaving exercise :)

Moves the "everything in one page" sandbox approach to one where each component gets its own route: this is useful for testing - e2e tests can run against these sandboxes instead of the already-overloaded Kitchensink route. Also makes the sandboxes a little nicer for us while developing components.

### DIFF
--- a/components/baseline-indicator/sandbox.js
+++ b/components/baseline-indicator/sandbox.js
@@ -194,18 +194,18 @@ const CASES = [
 
 export class BaselineIndicatorSandbox extends SandboxComponent {
   /**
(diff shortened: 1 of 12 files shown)


caugner (approved):

caugner:
Nit: Assuming this is targeted to developers, maybe use `build(sandbox):` to avoid it popping up in the release notes?

PR #1880, opened 2026-09-09
feat(compat-table): add browser visibility settings

### LINKED ISSUE #752: Allow users to customize browser selection in BCD tables

caugner:
### Description

Add browser visibility controls to compatibility tables.

- Select visible browsers in the Browser compatibility settings modal with live preview.
- Save choices locally and sync tables across tabs.
- Cap browser column widths for smaller selections.

### Motivation

Help readers configure compatibility tables for their needs.

- Show browsers hidden by default.
- Focus on browsers relevant to their project.

### Additional details

- Group browsers by platform, separating Baseline browsers, WebViews, and others. Explain unavailable columns in a legend.
- Allow any selection, including none. Save syncs tables across the page and other tabs; Cancel, close, or Escape revert the preview. Outside clicks leave the modal open.
- Saving the default selection clears browser overrides. Browsers without saved choices follow current defaults.
- Cap browser column widths and provide fallback icons.
- Extend `mdn-modal` with configurable dismissal and width, a forwarded close event, and heading typography.

#### Screenshot

image

#### Measurements

New Glean measurements:

- `bcd: browsers -> default` or `bcd: browsers -> <browser>,<browser>,...`: on table view, report `default` without saved overrides; otherwise list shown columns by BCD key in table order.
- `bcd: settings -> open`: on opening the modal.
- `bcd: settings -> show <browser>` / `bcd: settings -> hide <browser>`: on Save, once per changed browser.
- `bcd: settings -> reset`: before per-browser events when a changed selection is saved as defaults.

Fixes mdn/mdn#752.


danielhjacobs:
As an end-user I'm likely to click outside of the modal after configuring the browsers without clicking "OK". Right now that undoes my choices. It may be more user-friendly to open a confirmation prompt asking if you wish to save or discard your changes.

caugner:
Last commit before the history rewrite: https://github.com/mdn/fred/commit/b36dfc186793b69739351f24656c748bcaa775d4

This commit preserves the original commit chain for recovery.

components/compat-table-settings/element.js
```diff
@@ -0,0 +1,380 @@
...
+import "../modal/element.js";
+import styles from "./element.css?lit";
+
+/** Unlisted platforms follow in alphabetical order. */
+const PLATFORM_ORDER = ["desktop", "mobile", "server"];
+
+/**
+ * @type {readonly import("@bcd").BrowserName[]}
+ */
+const BASELINE_BROWSERS = Object.freeze([
```
LeoMcA: check: how is this used?

components/compat-table-settings/element.js
```diff
@@ -0,0 +1,380 @@
...
+  "chrome_android",
+  "edge",
+  "firefox",
+  "firefox_android",
+  "safari",
+  "safari_ios",
+]);
+
+/** @type {readonly import("@bcd").BrowserName[]} */
+const WEBVIEW_BROWSERS = Object.freeze(["webview_android", "webview_ios"]);
```
LeoMcA: check: how is this used?

components/compat-table-settings/element.js
```diff
@@ -0,0 +1,380 @@
...
+ */
+export class MDNCompatTableSettings extends L10nMixin(LitElement) {
+  static styles = styles;
+
+  static get properties() {
+    return {
+      browserInfo: { attribute: false },
+      visibility: { attribute: false },
+      hiddenBrowsers: { attribute: false },
+      _selected: { state: true },
```
LeoMcA: check: attribute false vs state true

components/compat-table-settings/element.js
```diff
@@ -0,0 +1,380 @@
...
+    </div>`;
+  }
+
+  render() {
+    return html`
+      <mdn-button variant="plain" .icon=${settingsIcon} @click=${this._open}
+        >${this.l10n("compat-settings-open")`Settings`}</mdn-button
+      >
+      <mdn-modal
+        closedby="closerequest"
```
LeoMcA, line 348: Is there a specific reason we don't allow clicking outside the modal to dismiss? It would make it consistent with our other modals:


```suggestion
        closedby="any"
```

components/compat-table-settings/element.js
```diff
@@ -0,0 +1,380 @@
...
+    /** @type {Partial<import("@bcd").Browsers>} */
+    this.browserInfo = {};
+    /**
+     * Saved visibility, synced across tabs by the parent table.
+     * @type {import("../compat-table/settings.js").BrowserVisibility}
+     */
+    this.visibility = {};
+    /**
+     * Exclusions that override user choices.
+     * @type {import("@compat").HiddenBrowsers}
```
LeoMcA: check: is this new?

components/compat-table-settings/element.js
```diff
@@ -0,0 +1,380 @@
...
+     */
+    this.hiddenBrowsers = {};
+    /** @type {Set<import("@bcd").BrowserName>} */
```
LeoMcA: nit: we're using `Browsers` and `BrowserName` and other types a lot here, might be worth elevating to a `/** @import` statement at the top for readability inline

test/unit/components/compat-table/settings.test.js
LeoMcA, line 1: This name is a bit unclear - it could be referring to testing `components/compat-table-settings`, but it's testing `components/compat-table/settings` - could we fully mirror the directory structure in this test and the other one?

components/compat-table/browsers.js
```diff
@@ -0,0 +1,95 @@
...
+ * @param {import("./settings.js").BrowserVisibility} visibility
+ * @returns {[string[], import("@bcd").BrowserName[], import("@compat").HiddenBrowsers]}
+ */
+export function gatherPlatformsAndBrowsers(
+  category,
+  data,
+  browserInfo,
+  visibility,
+) {
+  const isVisible = (/** @type {import("@bcd").BrowserName} */ browser) =>
```
LeoMcA: Could do with some top level `/* @import`s in this file too

LeoMcA (approved):
Really really nice!

I think generally we should aim to elevate things to top level `/** @import`s after a couple of uses inline, and the readability of most of the files here would benefit from that - but consider those nits.

The modal click-to-dismiss and testing file names are the only real question/concern I have.

components/compat-table-settings/element.js
caugner: See [the comment above](https://github.com/mdn/fred/pull/1880#issuecomment-5619471002). There are four ways to close the modal, and the important bit here is that the table changes in the background, so users might assume that clicking outside is equivalent to saving (which it is not, and probably shouldn't be). So disallowing clicking outside to dismiss is probably best.

test/unit/components/compat-table/settings.test.js
caugner: Agree. I was also thinking it might be nice to put the tests next to the component implementations, but then it's slightly uglier to reuse test utils.

PR #1886, opened 2026-09-10
feat(eslint): add `fred/no-external-link-whitespace` rule

caugner:
### Description

Add a custom ESLint rule `fred/no-external-link-whitespace` that flags `<a>` elements with the `external` class inside `html` templates whose content starts or ends with whitespace, with an autofix that strips it.

- Rule + `RuleTester` unit tests in `build/eslint-fred.js` and `test/unit/build/eslint-fred.test.js`.
- Enable the rule in `eslint.config.js` as `warn`, so it can land before the 14 existing violations are fixed.

### Motivation

Ensure the external-link icon stays attached to the last word of the link. Whitespace before `</a>` renders as a space, so when the link text wraps, the icon drops onto its own line. Prettier introduces this whitespace whenever it wraps the link text onto its own line, so a lint rule is needed to keep it out.

### Additional details

Prettier does not reintroduce the whitespace once removed: it formats a whitespace-free anchor as `>${text}</a\n>`, so the rule and the formatter are stable together.

Rari does not add whitespace to external links, and a scan of a fresh en-US content build found only one external link out of about 30k with trailing whitespace, so the templates in this repo are the main source.

### DIFF
--- a/build/eslint-fred.js
+++ b/build/eslint-fred.js
@@ -116,6 +116,53 @@ export default {
         };
       },
     },
+    "no-external-link-whitespace": {
+      meta: {
+        type: "problem",
+        fixable: "whitespace",
+        docs: {
+          description:
+            "disallow leading/trailing whitespace inside external links, so the icon stays attached to the last word",
+        },
+      },
+      create(context) {
+        return {
+          /**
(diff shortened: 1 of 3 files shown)


caugner:
Closing in favor of one-time fixes.

PR #1895, opened 2026-09-11
ci(dependabot): make our ignores explicit

LeoMcA:
At least one security alert wasn't automatically closed by a dependency bump because we had a stale `@rsdoctor/rspack-plugin` ignore condition hanging around.

I've gone through and removed all the comment-based ignore conditions which no longer seem necessary, so with our latest run [we had](https://github.com/mdn/fred/actions/runs/34581613076/job/103206231941#step:3:34):

`ignore-conditions":[{"dependency-name":"@types/node","version-requirement":">= 25.a, < 26","update-types":null,"source":"@dependabot ignore command","updated-at":"2025-12-11T08:38:29.000Z"},{"dependency-name":"@types/node","version-requirement":">= 26.a, < 27","update-types":null,"source":"@dependabot ignore command","updated-at":"2026-06-24T14:17:30.000Z"},{"dependency-name":"eslint-plugin-unicorn","version-requirement":"> 65.0.1","update-types":null,"source":"@dependabot ignore command","updated-at":"2026-07-03T11:45:53.000Z"},{"dependency-name":"eslint-plugin-unicorn","version-requirement":">= 69.a, < 70","update-types":null,"source":"@dependabot ignore command","updated-at":"2026-06-30T14:43:40.000Z"},{"dependency-name":"typescript","version-requirement":">= 7.a, < 8","update-types":null,"source":"@dependabot ignore command","updated-at":"2026-07-09T11:16:40.000Z"}]`

This makes these more visible by adding them to `dependabot.yml` - and I'd suggest we have a soft policy going forward of only temporarily adding comment-based ignores (e.g. to allow a grouped update to proceed if there's one misbehaving dependency) and then immediately un-ignoring them. Anything longer-running (like ensuring our node types match our node version) should be immortalised in `dependabot.yml`

Once merged I'll remove all those comment-based ignore conditions.

PR #1898, opened 2026-09-11
chore(eslint): disable jsdoc/multiline-blocks

LeoMcA:
Follow up to https://github.com/mdn/fred/pull/1895#discussion_r3987795713

Ignoring this rule is necessary for autofixing with a new unicorn version, because it was reformatting nice lines like:

```
/** @type {HTMLDialogElement} */
```

into:

```
/**
@type {HTMLDialogElement}
*/
```

And even if the reformatting wasn't broken, I don't think there's anything wrong with single line jsdoc comments - especially when we use them for type hinting so often as we do.

### DIFF
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -87,6 +87,7 @@ export default defineConfig([
         "error",
         { definedTags: ["element", "attr", "slot"] },
       ],
+      "jsdoc/multiline-blocks": "off",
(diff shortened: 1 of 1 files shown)


LeoMcA:
Oops it was actually `unicorn/single-line-block-comment-style`: https://github.com/mdn/fred/pull/1900

PR #1903, opened 2026-09-11
refactor(eslint): re-enable unicorn rules with a single error

LeoMcA:
No need to continue ignoring these because they're not auto-fixable if there's only one place to fix.

### DIFF
--- a/l10n/parser/transform.js
+++ b/l10n/parser/transform.js
@@ -7,6 +7,10 @@ import { Transformer, parse, serialize } from "@fluent/syntax";
  * @import { TextElement } from "@fluent/syntax";
  */
 
+const combiningMarks = Array.from({ length: 0x3_6f - 0x3_00 + 1 }, (_, i) =>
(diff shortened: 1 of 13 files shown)


l10n/parser/transform.js
```diff
@@ -7,16 +7,18 @@ import { Transformer, parse, serialize } from "@fluent/syntax";
...
  * Tranforms strings by:
  * - adding accents to all characters
  * - duplicating some vowels to create longer strings
  * - wrapping in square brackets to help detect truncation
  */
 class AccentTransformer extends Transformer {
-  MARKS = Array.from({ length: 0x3_6f - 0x3_00 + 1 }, (_, i) =>
-    String.fromCodePoint(0x3_00 + i),
-  );
+  MARKS = combiningMarks;
```
caugner, line 21: nit (pre-existing): Why not make this a global const?

caugner (approved):

